PR #404, opened 2024-10-30
[FIX] [STUDIO-99] Reduced new instance name max character count from 15 to 14 characters.

BboyAkers:
## Ticket: 
[STUDIO-99](https://harperdb.atlassian.net/browse/STUDIO-99)

## Overview:
- Reduces "Instance Name" character count from 15 characters to 14 characters.
- Version bump 4.6.0 -> 4.6.1

## Manual Testing:
1. Login using your credentials
2. Navigate to your "org"
3. Click "Create New HarperDB Cloud Instance" Card Button
4. Create an instance with an "Instance Name" character count to the max allowed(14) in the input

**Expected Result:**
You should be able to create an instance with 14 characters or less.

## Additional Context:
AWS expects 14 characters as the max and will throw an error if it's above it. Reducing the input to have a max of 14 characters fixes it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "harperdb-studio",
-  "version": "4.5.9",
+  "version": "4.6.1",
   "description": "A UI for HarperDB",
   "deploymentUrl": "studio.harperdb.io",
   "private": true,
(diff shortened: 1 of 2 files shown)


kriszyp (approved):
Way to jump right in with PR, good job!

PR #406, opened 2024-11-01
Allow user to update ram if the account is unpaid

Devin-Holland:


BboyAkers (approved):

BboyAkers:
Looks good! Look into the SonarCloud Code Analysis and once that's passed, merge it!!

Devin-Holland:
> Looks good! Look into the SonarCloud Code Analysis and once that's passed, merge it!!

I looked into it and I think it triggered because I added to code to a line that already was a nested ternary, not so much the code I added. 

I think the only way to fix it is to basically re-work the whole rendering part of the file.

I can do that if you'd like or we can merge.

Whatever you'd like let me know.

PR #416, opened 2024-11-15
[Enhancement] Added Dynamically Defined Keys in Rows as Table Columns 

BboyAkers:
## Ticket: 
[STUDIO-110](https://harperdb.atlassian.net/browse/STUDIO-110)

## Overview:
- Added functionality for dynamically defined keys in Rows to display as Table Columns. 

## Manual Testing:
1. Login
2. Choose an instance and navigate to the 'browse' sub-menu' tab and navigate to a database
4. Click on a row and add a new key/value pair in the JSON object save
**Expected Result:**
New column should be added to rows and clicking that column to initiate filtering should be disabled.

## Additional Context
When filtering on some columns, the data returned from the server doesn't have the dynamically added key/value pairs. The dynamically added columns disappear as well. Is that okay? @kriszyp

### DIFF
--- a/src/components/shared/DataTable.js
+++ b/src/components/shared/DataTable.js
@@ -71,6 +71,7 @@ function DataTable({
   onRowClick,
   sorted,
   loading,
+  dynamicAttributesFromDataTable,
   manual = false,
 }) {
   const { headerGroups, page, rows, prepareRow, state, setAllFilters, canPreviousPage, canNextPage, pageOptions, pageCount, gotoPage, nextPage, previousPage, setPageSize } =
@@ -111,7 +112,7 @@ function DataTable({
   return (
     <ErrorBoundary onError={(err, componentStack) => addError({ error: { message: err.message, componentStack } })} FallbackComponent={ErrorFallback}>
       <div className="react-table-scroller">
-        <DataTableHeader headerGroups={headerGroups} onSortedChange={onSortedChange} sorted={sorted} showFilter={showFilter} />
+        <DataTableHeader headerGroups={headerGroups} onSortedChange={onSortedChange} sorted={sorted} showFilter={showFilter} dynamicAttributesFromDataTable={dynamicAttributesFromDataTable} />
         {loading || localLoading ? (
           <div className="centered text-center">
             <i className="fa fa-spinner fa-spin" />
--- a/src/components/shared/DataTable.js
+++ b/src/components/shared/DataTable.js
@@ -111,7 +111,7 @@ function DataTable({
 
   return (
     <ErrorBoundary onError={(err, componentStack) => addError({ error: { message: err.message, componentStack } })} FallbackComponent={ErrorFallback}>
-      <div className="react-table-scroller">
+      <div role="table" aria-label="Table Data" className="react-table-scroller">
         <DataTableHeader headerGroups={headerGroups} onSortedChange={onSortedChange} sorted={sorted} showFilter={showFilter} dynamicAttributesFromDataTable={dynamicAttributesFromDataTable} />
         {loading || localLoading ? (
           <div className="centered text-center">
--- a/src/components/shared/DataTableHeader.js
+++ b/src/components/shared/DataTableHeader.js
@@ -1,14 +1,6 @@
 import React from 'react';
 import { Row, Col } from 'reactstrap';
 
-// // loop through each column and add ' disableSortBy: true' if column Header string matches string from dynamicAttributesFromDataTable
-// columns.forEach((column) => {
-//   dynamicAttributesFromDataTable.forEach((dynamicAttribute) => {
-//     if (column.Header === dynamicAttribute) {
-//       column.enableSorting = false
-//     }
-//   })
-// })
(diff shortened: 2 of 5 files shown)


src/functions/instance/getTableData.js
```diff
@@ -2,6 +2,20 @@ import describeTable from '../api/instance/describeTable';
 import searchByValue from '../api/instance/searchByValue';
 import searchByConditions from '../api/instance/searchByConditions';
 
+const getAttributesFromTableData = (tableData, existingAttributes) => {
+  const existing = new Set(existingAttributes);
+  const extra = new Set();
```
kriszyp, line 7: This looks good, and basically is what we want, but I would actually like to be a more adaptive with these columns. Let's make `extra` a Map of the property count, (property name is key, count of records with that property). And then let's sort by count (most first) and only add enough columns to have 8 columns:
`return Array.from(extra).sort(([,a], [,b]) => b - a).map(([key]) => key).slice(0, 8 - existingAttributes.length);`
And drop out of this function at the beginning if there are already eight columns ` if (existingAttributes.length >= 8) return;`

src/components/shared/DataTableHeader.js
```diff
@@ -10,7 +10,11 @@ const DataTableHeader = ({ headerGroups, onSortedChange, sorted, showFilter }) =
           {headerGroup.headers.map((column) => (
             <Col
               key={column.id}
-              onClick={() => onSortedChange([{ id: column.id, desc: sorted[0]?.id === column.id ? !sorted[0]?.desc : false }])}
+              onClick={() => {
+                if (!dynamicAttributesFromDataTable.includes(column.id)) {
```
kriszyp, line 14: There is no visual indication of what can be sorted and what can't, right?

kriszyp (commented):
Nice work with this, glad to see this going in. Maybe we can do a follow-up PR that provides some visual indication of sortable colums and that also enforces that non-indexed columns can't be sorted (the `indexed` flag should be returned from the `describe_table` call).

Also, not sure I understand the target branch. Is `4.8.0` a release branch? Why not `stage`?

kriszyp:
> The dynamically added columns disappear as well. Is that okay? @kriszyp

Yes, if the returned data doesn't have data for dynamic columns, then having them disappear is great.

src/components/shared/DataTableHeader.js
BboyAkers: No, I'll create some sort of "non-intractable" look for the column header!

BboyAkers:
> Nice work with this, glad to see this going in. Maybe we can do a follow-up PR that provides some visual indication of sortable colums and that also enforces that non-indexed columns can't be sorted (the `indexed` flag should be returned from the `describe_table` call).
> 
> Also, not sure I understand the target branch. Is `4.8.0` a release branch? Why not `stage`?

Looking at bundling several PRs together to create a 4.8.0 release. I'd be more than happy to just point this PR to stagging and increase the patch version by 1!

kriszyp:
Bundling PRs sounds good. If we merge several PRs into stage, and the merge the whole set from stage -> main, then that would effectively bundle several PRs into a single release, wouldn't it?

src/components/shared/DataTable.js
```diff
@@ -110,8 +111,8 @@ function DataTable({
 
   return (
     <ErrorBoundary onError={(err, componentStack) => addError({ error: { message: err.message, componentStack } })} FallbackComponent={ErrorFallback}>
-      <div className="react-table-scroller">
-        <DataTableHeader headerGroups={headerGroups} onSortedChange={onSortedChange} sorted={sorted} showFilter={showFilter} />
+      <div role="table" aria-label="Table Data" className="react-table-scroller">
+        <DataTableHeader headerGroups={headerGroups} onSortedChange={onSortedChange} sorted={sorted} showFilter={showFilter} dynamicAttributesFromDataTable={dynamicAttributesFromDataTable} />
```
cjoshmartin, line 115: @BboyAkers a little off topic but i think you would benefit from adding a linter and a git hook.
github is wrapping these commits in a weird way and would be easier to read your changes if it was all formatted the same

* https://github.com/prettier/eslint-config-prettier#installation
* https://prettier.io/docs/en/install.html#git-hooks

BboyAkers: Thanks for the feedback!! I'll add one this week!! @cjoshmartin  😃

src/functions/instance/getTableData.js
```diff
@@ -2,6 +2,20 @@ import describeTable from '../api/instance/describeTable';
...
+  for (const dataRow of tableData) {
+    for (const key of Object.keys(dataRow)) {
```
cjoshmartin, line 9: if  `extra` was a normal array, maybe you can do something like this: 
```javascript
const keys = new Set(Object.keys(dataRow));
extra.push(...Array.from(keys.intersection(existing)));
```

here are some set comparison functions: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set#set_composition

PR #420, opened 2024-11-18
[v4.8.0] Create Dynamic Columns from Dynamic Data & Form Field Fixes

BboyAkers:
Overview:
- Added functionality for dynamically defined keys in Rows to display as Table Columns. #416 
- Added Max Length to necessary Input Fields on Sign Up Page. Prevents people/bots from inputting unnecessarily long strings in our form fields #417
- Version bump: 4.7.1 -> 4.8.0


src/components/auth/SignUp.js
```diff
@@ -44,6 +44,7 @@ function SignUp() {
               <Input
                 id="firstname"
                 name="fname"
+                maxLength={20}
```
kriszyp, line 47: Why are these in this PR? We rejected that PR. Can you revert these changes?

src/functions/instance/getTableData.js
```diff
@@ -2,6 +2,26 @@ import describeTable from '../api/instance/describeTable';
...
+    for (const key of Object.keys(dataRow)) {
+      if (!existing.has(key)) {
+        const count = extra.get(key) || 0;
+        extra.set(key, count + 1);
+      }
+    }
+  }
+  const result = [];
+  for (const [key, count] of extra) {
+    if (count >= 8) {
```
kriszyp, line 18: This doesn't do the same thing I had suggested with my code. What I want:
* Dynamic attributes should only be added if they result in a maximum of eight columns
* Include dynamic attributes with the maximum count of records, up to the eight column maximum.

There is nothing about this that has anything to do with the record count being more or less than 8.

kriszyp (changes requested):

src/functions/instance/getTableData.js
BboyAkers: I interpreted it as creating a character map and only displaying key/value pairs of dynamically added columns if there 8+ instances of that key defined throughout the loaded table rows

BboyAkers:
@kriszyp ready for another round of reviews 🙂

kriszyp (approved):
Nice!

src/functions/instance/getTableData.js
```diff
@@ -76,6 +92,9 @@ export default async ({ schema, table, filtered, pageSize, onlyCached, sorted, p
     }
   }
 
+  dynamicAttributesFromDataTable = getAttributesFromTableData(newData, allAttributes)
+  allAttributes.push(...dynamicAttributesFromDataTable);
+
   // sort columns, but keep primary key / hash attribute first, and created and updated last.
   // NOTE: __created__ and __updated__ might not exist in the schema, only include if they exist.
   const orderedColumns = allAttributes.filter((a) => ![hashAttribute, '__createdtime__', '__updatedtime__'].includes(a)).sort();
```
kriszyp, line 100: In testing this, one thing I realized is that while we are intentionally sorting the columns so that extra columns appear on the right and least frequently used columns are on the furthest right, this is all undone by the sort call on this line. I kind of think we should remove this sort(), as I think there is more value in ordering columns by frequency/declaration order rather than just alphabetic.

DavidCockerill (approved):

src/functions/instance/getTableData.js
BboyAkers: Done!

kriszyp (approved):

PR #426, opened 2024-11-22
Caching Describe Table Call, Reducing Expensive Call To Backend

BboyAkers:
## Ticket

[STUDIO-122](https://harperdb.atlassian.net/browse/STUDIO-122)

## Overview
- Caching `describe_table` call to reduce the frequency of that expensive call being made to the backend
- Describe Table cache is cleared every minute as well.

## Manual Testing
Testing:
1. Navigate to a table
2. Interact with it via going to the next “page” of table data results

**Expected Result:**
The describe_table api call should not be triggered when navigating through pages of table data results.

## Additional Notes:
The describe_table cache is reset every 60 seconds.


src/functions/instance/state/describeTableCache.js
```diff
@@ -0,0 +1,23 @@
+const descriptionCache = new Map(JSON.parse(sessionStorage.getItem('descriptionCache')) || []);
```
Ethan-Arrowood, line 1: JSON.parse will fail if `sessionStorage.getItem('descriptionCache')` is invalid or `undefined`. If its `null` it will return `null`. I think you should try to make this safer. 

Maybe a function that wraps it in a try/catch: `() =>  { try { JSON.parse... } catch { return [] } }`

src/functions/instance/state/describeTableCache.js
```diff
@@ -0,0 +1,23 @@
...
+
+const getTableKey = (schema, table) => `${schema}/${table}`;
+
+const getTableDescriptionFromCache = (schema, table) => {
+	const tableKey = getTableKey(schema, table);
+	if (!descriptionCache.has(tableKey)) return null;
+	return descriptionCache.get(tableKey);
```
Ethan-Arrowood, line 8: Simplify, `.get()` will return `undefined` if the thing isn't in the Map so you can use `?? null` to eliminate the `.has` check.
```suggestion
	return descriptionCache.get(tableKey) ?? null;
```

Ethan-Arrowood (approved):
lgtm with some minor improvements

src/functions/instance/state/describeTableCache.js
kriszyp: Or just do a .get and let it return `undefined` (`undefined` is a good value to describe an undefined value).

src/functions/instance/state/describeTableCache.js
```diff
@@ -0,0 +1,23 @@
...
+};
+
+const setTableDescriptionInCache = (schema, table, description) => {
+	const tableKey = getTableKey(schema, table);
+	descriptionCache.set(tableKey, description);
+	sessionStorage.setItem('descriptionCache', JSON.stringify(Array.from(descriptionCache)));
```
kriszyp, line 14: I really don't think we even need or want to use session storage at all. The problem is that with this there is no timer for expiring the content, and it could be unbounded in how old the cached data is, until the 60 second timer fires a minute after the page loads. But on initial page load, we could end up showing very old cached table structure.

src/functions/instance/state/describeTableCache.js
BboyAkers: How about this?
```javascript
const descriptionCache = () => new Map(JSON.parse(sessionStorage.descriptionCache ?? '{}'));
```

src/functions/instance/state/describeTableCache.js
BboyAkers: Gotcha! I'll remove the use of sessionStorage 🙂

PR #439, opened 2024-12-02
Adding Pre-commit Hooks & Improving Prettier/ESLint rules

BboyAkers:
## Overview
- Added 'husky' for pre-commit hooks
  - `"prettier --write --ignore-unknown"` & `"eslint --fix"` run as pre-commit hooks
- Updated/Configured Prettier and ESLint rules
- Updated min node version to 20.18.0

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,12 +1,14 @@
 {
-  "printWidth": 180,
-  "singleQuote": false,
-  "endOfLine": "auto",
-  "tabWidth": 2,
-  "useTabs": false,
-  "semi": true,
-  "bracketSameLine": false,
-  "bracketSpacing": true,
-  "jsxSingleQuote": false,
-  "arrowParens": "always"
+	"printWidth": 120,
+	"quoteProps": "consistent",
+	"singleQuote": true,
+	"trailingComma": "es5",
+	"useTabs": true,
+	"endOfLine": "auto",
+	"tabWidth": 2,
+	"semi": true,
+	"bracketSameLine": false,
+	"bracketSpacing": true,
+	"jsxSingleQuote": false,
+	"arrowParens": "always"
 }
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,4 +1,9 @@
 {
   "printWidth": 180,
-  "singleQuote": true
+  "singleQuote": true,
+  "endOfLine": "auto",
+  "tabWidth": 2,
+  "useTabs": false,
+  "jsxBracketSameLine": false,
+  "jsxSingleQuote": false
 }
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -50,7 +51,7 @@
     "whatwg-fetch": "^3.6.2"
   },
   "engines": {
-    "node": ">=14.0.0"
+    "node": ">=20.18.0"
```
kriszyp, line 54: Does the build really not work on Node versions before v20?

kriszyp (changes requested):

package.json
BboyAkers: It does, I changed this because of 2 things:
- Node 14,16, & 18 are deprecated.
- As we start to update and leverage our libraries more having a min version of 20.18.0 prevents us from having possible upgrade issues in the future. I'm eventually wanting to push us to React 19 mid next year.

TBH it's primarily helping with dependency management and preventing future headaches 😅.

.prettierrc
```diff
@@ -1,4 +1,12 @@
 {
   "printWidth": 180,
-  "singleQuote": true
+  "singleQuote": false,
+  "endOfLine": "auto",
+  "tabWidth": 2,
+  "useTabs": false,
```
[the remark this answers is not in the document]
BboyAkers: I'll revisit them, thanks for catching this!

BboyAkers:
@kriszyp Updated PR 🙂

package.json
kriszyp: If Node's has already deprecated these versions, doesn't that make this unnecessary?

.prettierrc
```diff
@@ -1,4 +1,14 @@
 {
-  "printWidth": 180,
-  "singleQuote": true
+	"printWidth": 120,
+	"quoteProps": "consistent",
+	"singleQuote": true,
+	"trailingComma": "es5",
+	"useTabs": true,
```
kriszyp, line 6: I presume you are going to have a lot of whitespace updates in your commits going forward with this. Do you want that?

kriszyp (approved):

.prettierrc
BboyAkers: Hmmm 🤔 I'll look into:
- Creating a separate PR running prettier on the entire app
and/or
- Investigating if it's possible to run prettier and not effect the git history

kriszyp: Usually the strategy is to add (mass) prettier/lint commits in .git-blame-ignore-revs: https://github.com/HarperDB/harperdb/blob/main/.git-blame-ignore-revs
I just didn't know if you really wanted to do that right now.

PR #453, opened 2025-01-15
[Feature] Added sorting filter for org lists

BboyAkers:
## Ticket
[STUDIO-157](https://harperdb.atlassian.net/browse/STUDIO-157)

## Overview
Added button to sort list of orgs form A-Z and Z-A.

## Manual Testing:
1. Navigate to Sign-In Page and login
2. Once you are logged in and are at the `/organizations` page press the "Sort by Name:..." button next to the filter

**Expected Result**
User should be able to have list of orgs sorted by name from A-Z and Z-A. by pressing the button.


src/components/organizations/filter/SortOrgs.js
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import { Button } from 'reactstrap';
+import { useStoreState } from 'pullstate';
+
+import appState from '../../../functions/state/appState';
+
+const sortOrgsList = (orgs, isAscending) => {
```
cb1kenobi, line 7: **suggestion (non-blocking):** ‏Usually I advise against defining functions this way as an error thrown from this function will say the function is anonymous.

```suggestion
function sortOrgsList(orgs, isAscending) {
```

Then you can remove the `;` at the end.

I can't imagine the lint rules would require this way of defining the function, but if it does then I would consider changing or relaxing the rule.

cb1kenobi (approved):
LGTM

PR #456, opened 2025-01-23
[Feature] Logs Page

BboyAkers:
## Ticket
[STUDIO-160](https://harperdb.atlassian.net/browse/STUDIO-160)

## Overview
- Added a `/Logs` sub-route for instances to view server logs
- Added filtering capabilities for logs


kriszyp (commented):
Great job, this is a huge improvement over the previous log viewer.
I will share some of my own (mostly subjective) preferences for log viewing. First, (maybe a little more objective), reading the body of a log message in log info dialog in dark mode is pretty hard:
image

Generally, the act of log viewing generally involves looking through a large amount of log information and quickly scanning for key bits of information. Consequently, good/mature log viewing tools generally feature high information density, balanced with text legibility. However, in the studio log viewer, much of the screen real estate is "wasted":
image
The first five columns take up most of the horizontal space even though they are small bits of text. The log message is by far the most important part of the log entry, but it receives a minority of the space. The columns should be adjusted so that the Message column uses about 60%-70% of the horizontal space.
Vertically, the space usage is sub-optimal as well, with the actual content often compromising a small amount of the available space, with font size that is probably a little small, and too much line spacing. I would think you typically want the text height to at least be 50% of the line height. Here is an example of typical log viewing from a terminal that I would consider highly efficient:
image
Also, why does the scrollbar have to be so tiny for dragging (scrolling is often the first thing I do when I am looking at logs)?
image

Anyway, some of this is probably beyond the scope of what you were aiming to tackle with this PR, and some might be somewhat preferential. You should probably fix the Log Info dialog in dark mode for this PR, but beyond that, I am fine with merging this PR and doing further log updates at a different time, if you prefer.

Again, this is a great step forward, nice work!

BboyAkers:
Fixed:
- Log Row column widths/sizing
- Scrollbar width
- Increased height of each Log Row to show more of the message.
- Fixed Log Modal Message text color in dark mode

@kriszyp @mihaiharperdb

kriszyp:
It looks like the rows are fixed size now, taking too much space for small space and overflowing for large messages:
image

BboyAkers:
Update:
- Log Limit selection issue: Fixed
- Log Level "Info": Should be fixed as well! If not, DM me @mihaiharperdb and we can hop on a call so we can dive into it and figure it out. :slightly_smiling_face:
- Table Row overflow issue: Fixed
- `datetime-local` Input Calendar UI for FireFox: Firefox doesn't support HH:MM:SS in the default calendar UI. It defaults to a text input for that section of the input according to MDN. I can look into importing a Calendar UI dependency in a follow-up PR. I does however work as expected with the full HH:MM:SS UI for:

kriszyp (approved):

PR #459, opened 2025-02-12
v4.8.15

BboyAkers:
## PR
- #458 
- Version bump `4.8.14` -> `4.8.15`

## Ticket
[STUDIO-164](https://harperdb.atlassian.net/browse/STUDIO-164)

## Overview
When clicking on folders in our application editor, folders will now collapse.

## Manual Testing:
1. Log in and  navigate to `/Applications`
2. Once the page loads click on a folder in the file browser sidebar.

**Expected Result**
User should be able to collapse and expand the contents of a folders' children when clicking the folder icon.

## Additional Context
Looked into having it all collapse initially but that's a bit more tricky and will require more time to look into. This SHOULD be sufficient and allow our end users to be a bit more happier when navigating through the file browser sidebar.


kriszyp (commented):
The directory collapsing looks good. Can we apply this collapsing functionality to the application folder as well? That is, you should be able to click on the application "cube" and collapse the whole application directory. And starting with the application directory collapsed is probably the preferred state.

BboyAkers:
> The directory collapsing looks good. Can we apply this collapsing functionality to the application folder as well? That is, you should be able to click on the application "cube" and collapse the whole application directory. And starting with the application directory collapsed is probably the preferred state.

Done!!!! 🙂

Ethan-Arrowood (approved):
lgtm. nice work! we should get this over to RL asap. Make them even happier.

kriszyp:
Looking great! A few issues:
* I think the target for collapsing/expanding should be the whole folder name, not just the icon. When everything is collapsed and you click on the app and nothing shows or expands, I think that is confusing.
* When you save any changes, everything collapses. I don't think we want to change the collapse/expand state when things are saved
* I can't expand when I am connecting to some of our remote servers. Is it possible that maybe this can't be expanded for apps that are deployed by github reference (and are read-only)?

mihaiharperdb:
@BboyAkers I agree with @kriszyp on:

- the target for collapsing/expanding should be the whole folder name, not just the icon. When everything is collapsed and you click on the app name (text) and nothing shows or expands, I think that is confusing.

- when you save any changes or add/delete a folder/file, all applications expand, although they were collapsed and I was working on only one expanded application


@kriszyp I deployed locally apps by github reference and I can expand/collapse.

PR #462, opened 2025-04-22
v4.8.17

BboyAkers:
Overview:
- Auth cleanup
- Fixed Local Sign-in bug [CORE-2694](https://harperdb.atlassian.net/browse/CORE-2694)

### DIFF
--- a/src/components/instances/new/MetaLocal.js
+++ b/src/components/instances/new/MetaLocal.js
@@ -13,6 +13,7 @@ import registrationInfo from '../../../functions/api/instance/registrationInfo';
 import isAlphaUnderscoreHyphen from '../../../functions/util/isAlphaUnderscoreHyphen';
 import isAlphaNumericHyphen from '../../../functions/util/isAlphaNumericHyphen';
 import userInfo from '../../../functions/api/instance/userInfo';
+import login from '../../../functions/api/instance/login';
 
 function MetaLocal() {
 	const navigate = useNavigate();
@@ -50,7 +51,23 @@ function MetaLocal() {
 				setFormState({ error: 'usernames must have only letters, underscores, and hyphens' });
 			} else if (instance_name.length && user.length && pass.length && host.length && port.length) {
 				try {
-					const currentUser = await userInfo({ auth: { user, pass }, url, is_local: true, customer_id });
+					let login_not_supported;
+					let currentUser;
+					if (pass) {
+						const login_result = await login({ auth: { user, pass }, url });
+						if (login_result.error) {
+							setFormState({ error: 'Login failed. Using instance credentials?' });
+						} else {
+							currentUser = await userInfo({ url });
+							if (currentUser.error) {
+								login_not_supported = true;
+							}
+							if (login_not_supported)
+								currentUser = await userInfo({ auth: { user, pass }, url, is_local: true, customer_id });
+						}
+					} else {
+						currentUser = await userInfo({ url });
+					}
 
 					if (currentUser.error && currentUser.message === 'Login failed') {
 						setFormState({ error: 'The provided credentials cannot log into that instance.' });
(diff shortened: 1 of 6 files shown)


src/components/instances/new/MetaLocal.js
```diff
@@ -50,7 +51,23 @@ function MetaLocal() {
...
+						const login_result = await login({ auth: { user, pass }, url });
+						if (login_result.error) {
+							setFormState({ error: 'Login failed. Using instance credentials?' });
+						} else {
+							currentUser = await userInfo({ url });
+							if (currentUser.error) {
+								login_not_supported = true;
+							}
+							if (login_not_supported)
+								currentUser = await userInfo({ auth: { user, pass }, url, is_local: true, customer_id });
```
cb1kenobi, line 66: Not sure what's going on here, but this seems like unnecessary complexity.

```suggestion
							if (currentUser.error) {
								currentUser = await userInfo({ auth: { user, pass }, url, is_local: true, customer_id });
							}
```

src/components/instances/new/MetaLocal.js
```diff
@@ -50,7 +51,23 @@ function MetaLocal() {
 				setFormState({ error: 'usernames must have only letters, underscores, and hyphens' });
 			} else if (instance_name.length && user.length && pass.length && host.length && port.length) {
 				try {
-					const currentUser = await userInfo({ auth: { user, pass }, url, is_local: true, customer_id });
+					let login_not_supported;
+					let currentUser;
+					if (pass) {
+						const login_result = await login({ auth: { user, pass }, url });
```
cb1kenobi, line 57: Please be consistent with your casing. It seems much of the code is camelcase, so all new code should be camelcase.

src/components/instances/new/MetaLocal.js
```diff
@@ -50,7 +51,23 @@ function MetaLocal() {
 				setFormState({ error: 'usernames must have only letters, underscores, and hyphens' });
 			} else if (instance_name.length && user.length && pass.length && host.length && port.length) {
 				try {
-					const currentUser = await userInfo({ auth: { user, pass }, url, is_local: true, customer_id });
+					let loginNotSupported;
```
kriszyp, line 54: I'm confused by what we are trying to do here. It looks like this was just copy and pasted from CardBackLogin, but that is where the actual login occurs, whereas I though this code was just checking to see if the user was already authenticated. And this logic largely existed to determine if login was supported and appropriately save the credentials, but `loginNotSupported` is even't used here (and shouldn't be); and it doesn't make sense to determine that here, I don't think. And if there is an existing `pass` then we should already have determined that basic auth is the way (not login).

kriszyp (changes requested):

src/components/instances/new/MetaLocal.js
BboyAkers: Transparently, I was trying to do a quick patch and didn't dive to far into the code 😅  to focus back on Fabric UI at the time. I'll properly fix it!

src/components/instances/new/MetaLocal.js
```diff
@@ -50,7 +51,20 @@ function MetaLocal() {
 				setFormState({ error: 'usernames must have only letters, underscores, and hyphens' });
 			} else if (instance_name.length && user.length && pass.length && host.length && port.length) {
 				try {
-					const currentUser = await userInfo({ auth: { user, pass }, url, is_local: true, customer_id });
+					let currentUser;
+					if (pass) {
```
kriszyp, line 55: I still don't understand the purpose of this whole code block/change, why login if the password is already persisted? What is wrong with the previous one-liner to get userInfo?

BboyAkers: I reverted it

kriszyp (approved):
Thank you!

PR #464, opened 2025-05-16
v4.8.19

BboyAkers:
Overview:
- Updated from 4.8.18 -> 4.8.19
- "Only cached" is defaulted to 'on'

### DIFF
--- a/src/components/instance/browse/BrowseDatatableHeader.js
+++ b/src/components/instance/browse/BrowseDatatableHeader.js
@@ -64,7 +64,7 @@ function BrowseDatatableHeader({
 					onClick={toggleOnlyCached}
 				>
 					<span className="me-2">only cached</span>
-					<i className={`fa fa-lg fa-toggle-${!onlyCached ? 'on' : 'off'}`} />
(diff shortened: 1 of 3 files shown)


src/components/instance/browse/BrowseDatatableHeader.js
```diff
@@ -64,7 +64,7 @@ function BrowseDatatableHeader({
 					onClick={toggleOnlyCached}
 				>
 					<span className="me-2">only cached</span>
-					<i className={`fa fa-lg fa-toggle-${onlyCached ? 'on' : 'off'}`} />
+					<i className={`fa fa-lg fa-toggle-${!onlyCached ? 'on' : 'off'}`} />
```
kriszyp, line 67: Does this change the default or just change how the button displays the value (seems like it reversing how it shows the value and shows in incorrectly reversed)?
I thought you needed to change the default here: https://github.com/HarperDB/hdbms/blob/master/src/components/instance/query/Datatable.js#L28

kriszyp:
Thank you!

PR #467, opened 2025-06-04
Updates on /Instance

BboyAkers:
Overview:
Instance pages that were worked on in the PR:
`/browse` page
 - Adding a database
 - Adding a table
 - Deleting a database
 - Deleting a table

  `/applications/new`
- Create new application
- Import new application(with github only atm)

`/applications/edit`
 - Basic editor (WIP)

`/logs`
  - View & filter logs

ect.


src/features/cluster/modals/NewInstanceModal.tsx
```diff
@@ -27,7 +27,7 @@ import {
 	SelectValue,
 } from '@/components/ui/select';
 import { NewInstanceInfo, useCreateNewInstanceMutation } from '@/features/cluster/hooks/useCreateNewInstance';
-import { renderInstanceTypeOption } from '@/shared/functions/InstanceType';
+import { InstanceTypes, renderInstanceTypeOption } from '@/shared/functions/InstanceType';
```
cb1kenobi, line 30: Import types using the `type` keyword.

```
import { type InstanceTypes, renderInstanceTypeOption } from '@/shared/functions/InstanceType';
```

BboyAkers: Ahhhh interesting!!🤔 🙂 Does that mean I shouldn't export it as a type?
https://github.com/HarperDB/hdbms/blob/harper/fabric/src/shared/functions/InstanceType.ts#L18

cb1kenobi:
Unclear why the empty file `src/features/instance/applications/editor/components/ApplicationsSidebar/FileTreeExplorer.tsx` was committed. Does it have a purpose? Is there a ticket to implement a file tree explorer?

src/features/cluster/modals/NewInstanceModal.tsx
cb1kenobi: No, you should export it as a type and subsequently import it as a type.

I would go one step further and just export the type declaration in a single line. There's no need to separate them.

```typescript
export type InstanceTypes = 'typ-1' | 'typ-2' | 'typ-3';
```

src/features/instance/applications/new/ImportProjectForm.tsx
```diff
@@ -0,0 +1,127 @@
...
+const ImportProjectSchema = z.object({
+	newApplicationName: z
+		.string()
+		.min(1, { message: 'Project name is required' })
+		.max(75, { message: 'Project name must be less than 75 characters' })
+		.regex(/^[a-zA-Z0-9-_]+$/, { message: 'Can only contain letters, numbers, dashes and underscores' }),
+	applicationUrl: z.string(),
+});
+
+const route = getRouteApi('');
```
cb1kenobi, line 27: Is there an advantage to passing an empty string as opposed to the more clear `'/'` route id?

src/features/instance/applications/new/functions/getGithubRepo.ts
```diff
@@ -0,0 +1,119 @@
...
+
+async function getGithubRepo(url: URL) {
+	if (url && url.pathname && url.hostname.includes('github.com')) {
+		const parts = url.pathname.split('/').filter(Boolean);
+
+		try {
+			const response = await fetch(`https://api.github.com/repos/${parts[0]}/${parts[1]}`);
+			const data = await response.json();
+
+			return response.status < 400 ? data.name : null;
```
cb1kenobi, line 13: It's easier to check the `.ok` prop: https://developer.mozilla.org/en-US/docs/Web/API/Response/ok

```javascript
return response.ok ? data.name : null;
```

BboyAkers:
> Unclear why the empty file `src/features/instance/applications/editor/components/ApplicationsSidebar/FileTreeExplorer.tsx` was committed. Does it have a purpose? Is there a ticket to implement a file tree explorer?

Currently a WIP in a different branch. Did a bit of a pivot to merge this in the middle of me working on a feature 😅

src/features/instance/applications/new/functions/getGithubRepo.ts
```diff
@@ -0,0 +1,119 @@
...
+	const meta: PackageMeta = {
+		type: null,
+		user: null,
+		repo: null,
+		url: null,
+		package: null,
+		tag: null,
+	};
+
+	if (pkg.url.match('://')) {
```
cb1kenobi, line 80: `match()` should only be passed a regexp. You should either use `includes()` or use the regexp `/:\/\//` (which looks a little gross).

src/features/instance/applications/new/functions/getGithubRepo.ts
```diff
@@ -0,0 +1,119 @@
...
+		// it's a url
+		meta.url = pkg.url;
+		meta.type = 'url';
+	} else if (pkg.url.match('semver:')) {
+		// it's a github repo
+		const [user, repo, semverTag] = pkg.url.split(/[/#]/);
```
cb1kenobi, line 86: This regex is bad. You need to escape the forward slash in the middle: `/[\/#]/`.

src/features/instance/applications/new/functions/isValidTarballUrl.ts
```diff
@@ -0,0 +1,11 @@
+import isValidUrl from '@/shared/functions/isValidUrl';
+
+function isValidTarballUrl(url: string) {
+	// npm restrictions on the tarball url install here: https://docs.npmjs.com/cli/v9/commands/npm-install
+	// updated to indexOf to allow signed URLs (with tokens appended)
+	return isValidUrl(url) && (url.indexOf('.tar') !== -1 || url.indexOf('.tar.gz') !== -1 || url.indexOf('.tgz') !== -1);
```
cb1kenobi, line 6: I recommend using `includes()` for readability.

```typescript
return isValidUrl(url) && (url.includes('.tar') || url.includes('.tar.gz') || url.includes('.tgz'));
```

You could even do it all in one regex:

```typescript
return isValidUrl(url) && /\.(tar(\.gz)?|tgz)$/.test(url);
```

src/features/instance/applications/new/functions/parsePackageType.ts
```diff
@@ -0,0 +1,63 @@
...
+	const meta: PackageMetaType = {
+		type: null,
+		user: null,
+		repo: null,
+		url: null,
+		package: null,
+		tag: null,
+	};
+
+	if (pkg.url.match('://')) {
```
cb1kenobi, line 24: Same thing here as mentioned in `getGithubRepo`.

src/features/instance/applications/new/functions/getGithubRepo.ts
```diff
@@ -0,0 +1,119 @@
+function isValidProjectName(name: string) {
```
cb1kenobi, line 1: Just a nitpick, but the filename should be `GitHub`, not `Github`.

src/features/instance/applications/new/functions/getGithubRepo.ts
```diff
@@ -0,0 +1,119 @@
...
+	return /^[a-zA-Z0-9-_]+$/.test(name);
+}
+
+async function getGithubRepo(url: URL) {
```
cb1kenobi, line 5: I would make all references to "GitHub" with the correct casing.

src/features/instance/applications/new/functions/parsePackageType.ts
```diff
@@ -0,0 +1,63 @@
...
+		// it's a url
+		meta.url = pkg.url;
+		meta.type = 'url';
+	} else if (pkg.url.match('semver:')) {
+		// it's a github repo
+		const [user, repo, semverTag] = pkg.url.split(/[/#]/);
```
cb1kenobi, line 30: Same thing here as mentioned in `getGithubRepo`.

src/shared/functions/isValidUrl.ts
```diff
@@ -0,0 +1,9 @@
+const isValidUrl = (url: string) => {
```
cb1kenobi, line 1: I would discourage defining arrow functions that are exported. The function will be anonymous in stacktraces.

I would also export the function in the same line as the declaration. Export semantics overhead is low.

```typescript
export default function isValidUrl(url: string) {
    <snip>
}
````

File: vite.config.ts [hunk 12]
```diff
@@ -1,14 +1,18 @@
-import path from "path"
-import { defineConfig } from 'vite'
-import tailwindcss from '@tailwindcss/vite'
-import react from '@vitejs/plugin-react'
+import path from 'path';
```
cb1kenobi, line 1: Can this be `import path from 'node:path';`?

vite.config.ts
```diff
@@ -1,14 +1,18 @@
...
-  resolve: {
-    alias: {
-      "@": path.resolve(__dirname, "./src"),
-    },
-  },
-})
+	plugins: [react(), tailwindcss()],
+	resolve: {
+		alias: {
+			'@': path.resolve(__dirname, './src'),
```
cb1kenobi, line 11: I'm not sure how this is working today. This repo has `"type": "module"` defined in the `package.json`, it should treat all .js/.ts files as ESM in which case `__dirname` does not exist: https://nodejs.org/docs/latest/api/esm.html#no-__filename-or-__dirname.

The docs suggest using `import.meta.dirname`, but that wasn't generally available until Node.js 24, so you will need to do the `const __dirname = dirname(fileURLToPath(import.meta.url));` hack.

What am I missing?

src/features/instance/operations/queries/getRegistrationInfo.ts
```diff
@@ -0,0 +1,26 @@
...
+	license_expiration_date: string;
+	ram_allocation: number;
+	registered: boolean;
+	version: string;
+};
+
+function getRegistrationInfoQueryOptions(instanceId: string) {
+	return queryOptions({
+		queryKey: [instanceId, 'registration_info'] as const,
+		queryFn: async () => {
```
cb1kenobi, line 15: Since this function doesn't reference `this`, I don't see the point in doing an anonymous function.

```typescript
		async queryFn() {
			const { data } = await instanceClient.post('/', {
				operation: 'registration_info',
			});
			return data as RegistrationInfoResponse;
		},
```

src/components/RadioButtonGroup/index.tsx
```diff
@@ -84,11 +84,11 @@ function RadioButtonGroup<
 		<RadioGroup value={localValue} onValueChange={handleValueChange} className="flex flex-wrap gap-2" name={name}>
 			{options.map((option) => (
 				<div key={option.value} className="flex items-center space-x-2">
-					<RadioGroupItem value={option.value} id={`${name}-${option.value}`} className="peer sr-only" />
+					<RadioGroupItem value={option.value} id={`${name}-${option.value}`} className="sr-only peer" />
 					<Label
 						htmlFor={`${name}-${option.value}`}
 						className={cn(
-							'flex h-10 cursor-pointer items-center justify-center rounded-md border border-input px-4 py-2 text-sm font-medium ring-offset-background transition-colors hover:bg-accent hover:text-accent-foreground focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring focus-visible:ring-offset-2',
+							'flex h-10 cursor-pointer items-center bg-black/30 justify-center rounded-md border border-input px-4 py-2 text-sm font-medium ring-offset-background transition-colors hover:bg-accent hover:text-white focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring focus-visible:ring-offset-2',
```
heskew, line 91: tiny nit: why so many class names? 😭

src/features/auth/hooks/useSignOut.ts
```diff
@@ -5,10 +5,8 @@ import { useMutation } from '@tanstack/react-query';
 
 export const onSignOutSubmit = async () => {
 	const { status } = await apiClient.post('/Logout');
-	if (status === 200) {
+	if (status === 204) {
```
heskew, line 8: what happens if the status is some other 2xx or 3xx? does it matter, meaning we don't need to check?

src/features/cluster/modals/InstanceLoginInModal.tsx
```diff
@@ -64,7 +59,7 @@ function InstanceLogInModal({
 			toast.success('Logged in to instance');
 			form.reset();
 		} catch (error) {
-			toast.error('Error logging in to instance');
+			toast.error(`Error: ${error}`);
```
heskew, line 62: will something more useful be logged later?

src/features/cluster/modals/NewInstanceModal.tsx
```diff
@@ -71,27 +71,19 @@ function NewInstanceModal({ clusterId }: { clusterId: string }) {
...
+	// const parseUrlInput = (url: string) => {
+	// 	if (isValidUrl(url)) {
+	// 		const parsedUrl = new URL(url);
+	// 		const { protocol, hostname, port } = parsedUrl;
+	// 		return { protocol, hostname, port };
+	// 	}
+	// };
 
 	// console.log('test url:', parseUrlInput('http://127.0.0.2:9925'));
 	const { mutate: submitNewInstanceData } = useCreateNewInstanceMutation();
```
heskew, line 89: more related fixing the storage type below, could the zod validation to coercion so formData is more correct?

src/features/cluster/modals/NewInstanceModal.tsx
```diff
@@ -165,7 +157,7 @@ function NewInstanceModal({ clusterId }: { clusterId: string }) {
 												<SelectGroup>
 													{instanceTypes?.map((type) => (
 														<SelectItem key={type.id} value={type.id}>
-															{renderInstanceTypeOption(type.id)}
+															{renderInstanceTypeOption(type.id as InstanceTypes)}
```
heskew, line 160: could this type assertion be avoided somehow, especially since this appears to be dynamic data, since it doesn't really do anything for the app? could try to ensure the data is appropriately validated and typed? food for thought later. :)

src/features/cluster/modals/NewInstanceModal.tsx
heskew: also, on creating updatedFormData below, is it expected that the formData could override clusterId and operations properties?

src/components/RadioButtonGroup/index.tsx
cb1kenobi: Tailwind.css. This is normal. This is the way.

src/features/auth/hooks/useSignOut.ts
cb1kenobi: Like I mentioned elsewhere, I would prefer:

```typescript
const response = await apiClient.post('/Logout');
if (response.ok) {
```

vite.config.ts
BboyAkers: Reading this 🙂 
https://v2.vitejs.dev/config/#config-file-resolving

src/features/instance/operations/queries/getRegistrationInfo.ts
BboyAkers: If i'm not mistaken this is a recommended ways for an async function to work inside `queryFn()` I'll double check!

src/components/RadioButtonGroup/index.tsx
BboyAkers: 😅

heskew: > This is the way.

image

src/features/auth/hooks/useSignOut.ts
heskew: 💯

BboyAkers: There's no `.ok` on this call ☹️

src/features/cluster/modals/NewInstanceModal.tsx
BboyAkers: I'll have to ask @DavidCockerill  more about that 🤔 . I'm unsure how we want to handle creating new Instances inside clusters.

BboyAkers: This work will be coupled with me updating and working on the "new cluster" work

src/features/cluster/modals/NewInstanceModal.tsx
BboyAkers: I'll if I can, I was having a ton of trouble with this 😓. Would love to pair on this!

src/features/instance/applications/new/ImportProjectForm.tsx
BboyAkers: According to the tanstack docs this would be the best way for me to get the route params and stuff 🙂

src/features/auth/hooks/useSignOut.ts
heskew: that's...not ok ;)

there's nothing saying this 'okay'? :( what is this api client?

cb1kenobi: How can this be?

src/features/instance/applications/new/ImportProjectForm.tsx
cb1kenobi: Link?

src/features/cluster/modals/NewInstanceModal.tsx
DavidCockerill: To create new instances in an existing cluster you'll be able to use update cluster or add instance endpoints

PR #476, opened 2025-06-20
Fabric config

dawsontoth:


src/features/auth/SignUp.tsx
```diff
@@ -51,9 +51,9 @@ export function SignUp() {
 
 	const submitForm = async (formData: z.infer<typeof SignInSchema>) => {
 		submitSignUpData(formData, {
-			onSuccess: (message) => {
+			onSuccess: () => {
 				toast.success('Success', {
-					description: `${message}`,
+					description: 'Your account has been created! Please sign in with your new credentials.',
```
dawsontoth, line 56: In practice, this API doesn't respond with a `message`, it responds with a user object (which has no message field on it). So I switched to a hard coded string. 🤔 We don't translate our UI yet at all, right?

src/features/instance/browse/BrowseDataTableView.tsx
```diff
@@ -42,45 +28,102 @@ export function BrowseDataTableView() {
...
+		pagination.pageSize,
+		pagination.pageIndex,
+	]);
+	useEffect(() => {
+		void refetchSearchByValueOptions();
+	}, [
+		refetchSearchByValueOptions,
+		pagination.pageSize,
+		pagination.pageIndex,
+	]);
```
dawsontoth: I pulled all of the useEffects to be in one spot, and I broke a few of the larger ones apart a bit. You'll also see a pattern of refetchBySearchHash and refetchBySortTableParams. What that lets us do is trigger a refetch, but wait until the params are updated. It's very possible that we don't need to do this, and we could instead use the query key in the same way, and rely on react-query to only refetch when we're making a new request.

src/features/instance/operations/queries/getSearchByValue.ts
```diff
@@ -44,9 +45,20 @@ function getSearchByValueOptions({
...
+		queryKey: [
+			'search_by_value',
+			hash_attribute,
+			instanceId,
+			pagination.pageIndex || 0,
+			pagination.pageSize || 0,
+			schemaName,
+			sortTableDataParams.attribute || 'default',
+			sortTableDataParams.descending || false,
+			tableName,
```
dawsontoth, line 57: Adding these here allows us to fetch whenever these change, instead of having to trigger the refetch manually.

BboyAkers (approved):

PR #478, opened 2025-06-30
[STUDIO-167] Cloud Auto Sign In

dawsontoth:
I made a few advancements for us here:
1. I added an example of a lazy loaded route. I chose Profile because it isn't implemented, and isn't important, so if this breaks stuff once deployed, we'll notice and be OK. 😄 
2. I shifted the approach for tracking the logged in user to use context and hooks. Now when checking the routes, we can redirect the user as appropriate, and maintain where they were in the UI with a `?redirect=/wherever` param.
3. The auth pages have a nav bar now, just like Studio. I also added icons, though I didn't get their hover state quite right. Progress over perfection!

https://harperdb.atlassian.net/browse/STUDIO-167


src/App.tsx
```diff
@@ -29,18 +33,20 @@ export function App() {
...
 		context: {
 			queryClient,
+			authentication: authenticationContextValue,
 		},
 	});
 
 	return (
 		<>
-			<QueryClientProvider client={queryClient}>
+			<AuthenticationContext value={authenticationContextValue}>
```
dawsontoth, line 42: I pushed the QueryClientProvider out to the `main.tsx` so that we can utilize the api client in the `useRootAuthenticationContextValue()`.

src/components/Navbar/index.tsx
```diff
@@ -36,75 +39,84 @@ function MobileNav({ signOut }: { signOut: () => void }) {
...
-					Sign Out
+					<LogOutIcon className="mr-4" /> Sign Out
 				</Link>
 			</div>
 		</div>
 	);
 }
 
-const DesktopNav = ({ signOut }: { signOut: () => void }) => {
+function DesktopNav({ signOut }: { signOut: () => void }) {
```
dawsontoth, line 78: Bike shed color choice: thoughts on standardizing on the stylistic choice here? `const [name] = () => {}` vs `function [name] {}`? I prefer `function` because it is less strict on the ordering, which lets you organize the code for readability a bit more flexibly. Plus, `function` is immediately recognizable when you're reading through the code, vs all the different stuff that can be stored in a `const`. Those are my preferences around it, take them for what you will!

src/features/auth/ForgotPassword.tsx
```diff
@@ -77,7 +77,7 @@ export function ForgotPassword() {
 				<Link className="text-sm" to="/">
 					Sign in to your account
 				</Link>
-				<Link className="text-sm" to="/signup">
+				<Link className="text-sm" to="/sign-up">
```
dawsontoth, line 80: Studio has a `-` in the name, so I made the slight parity adjustment here. I didn't touch the `resetpassword` route since it looks like central-manager has a reference to that.

src/hooks/use-root-authentication-context-value.ts
```diff
@@ -0,0 +1,36 @@
...
+		queryFn: getCurrentUser,
+		retry: false,
+		enabled: !isLocal,
+	});
+	useEffect(() => setUser(data || null), [data, setUser]);
+	return useMemo(() => {
+		return {
+			isLoading,
+			isLocal,
+			setUser,
```
dawsontoth, line 22: setUser was pretty much the entire onus for creating this context, instead of doing stuff through react-query entirely. It lets us immediately set or unset the user when signing in or out.

src/router/cloudRouter.ts
```diff
@@ -50,16 +51,22 @@ const signInRoute = createRoute({
 	getParentRoute: () => authLayout,
 	path: '/',
 	component: SignIn,
+	beforeLoad: ({ context, location }) => {
+		if (context.authentication.user) {
+			const search: Record<string, string> = location?.search;
+			throw redirect({ to: search?.redirect?.startsWith('/') ? search.redirect : '/orgs' });
+		}
```
dawsontoth, line 58: I only added this check to the sign-in page. I figure someone might legitimately end up at the sign-up page or forgot-password. But if you're on the sign-in page, and you're already signed in, we should redirect you.

src/router/cloudRouter.ts
```diff
@@ -81,13 +88,23 @@ const dashboardLayout = createRoute({
 	getParentRoute: () => rootRoute,
 	id: '_dashboardLayout',
 	component: Dashboard,
+	beforeLoad: ({ context, location }) => {
+		if (!context.authentication.isLoading && !context.authentication.user) {
+			throw redirect({
+				to: '/',
+				search: {
+					redirect: location.href,
```
dawsontoth, line 96: We should make sure this works with more complex routes, I got a bit blocked by central manager changes for creating clusters 😮‍💨

src/components/Navbar/index.tsx
BboyAkers: sounds good to me!!

BboyAkers (approved):

PR #479, opened 2025-06-30
[STUDIO-182] Local Sign In and Routing

dawsontoth:
Screenshot 2025-06-30 at 5 10 10 PM

This cleans up the routes a bit between cloud and local modes, and utilizes the same mechanisms I introduced in https://github.com/HarperDB/hdbms/pull/478 to check the sign in status and redirect as appropriate.

There's still more work to be done on the local mode, but it does appear to be mostly working.

https://harperdb.atlassian.net/browse/STUDIO-182
https://harperdb.atlassian.net/browse/STUDIO-183


src/features/instance/queries/getInstanceInfoQuery.ts
```diff
@@ -37,8 +39,16 @@ type Instance = {
...
 const getInstanceInfo = async (instanceId: string) => {
+	if (isLocalStudio) {
+		return null;
+	}
 	// TODO: Work through any disagreements between the SchemaInstanceType and our local types here.
 	const { data } = await apiClient.get(`/HDBInstance/${instanceId}` as '/HDBInstance/{id}');
+	if (data) {
+		// Set the base URL for the instance client to the first FQDN of the instance
+		// This allows all subsequent API calls to use the correct base URL for the instance
+		instanceClient.defaults.baseURL = data.instanceFqdn;
```
dawsontoth: When we're doing the "local" version, in a customer's like... deployed instance... how will HDBMS be running? Is it served on the same port as the operations API? And in that case, we're OK not setting the baseURL. If the port is different, we'll find out!

BboyAkers (approved):

PR #484, opened 2025-07-02
Harper/case consistency

dawsontoth:
1. folders are always camelCase
2. any file exporting multiple things is also camelCase.ts
3. a file exporting a single class or component is PascalCase.tsx
4. numbers in paths are separated by `-`, though we don't have any of those here yet
5. index.ts[x] is always lowercase
6. avoid single-index-folders for components, until said component grows to have other sub-components (this will make opening files simpler for us devs)

Here's a few useful shell scripts:
```
find . -type f -name "*-*" | grep -v "node_modules" | grep -v ".git"
```


BboyAkers (approved):

kriszyp:
According https://docs.google.com/document/d/1UhP9fqUa-DqIMCPlf6qMWJxWa0YIqerxNnqXfNOdw14/edit?tab=t.0#heading=h.icw6vb57e95z our node.js coding standard dictates that modules should be camelCased.ts.
I guess we could interpret this a few different ways:
* That's node.js, naming conventions should be different in the browser
* Studio team is lobbying for a change in naming convention that should be adopted across the company (it is certainly a proposal worth entertaining)
* Wait, do we really want inconsistent naming between studio and core?

dawsontoth:
@kriszyp I apologize, this was something Austin and I had talked about, and there's inconsistency in the codebase currently. I've had compilation and production issues result from the case sensitivity discrepancies that can come between OS and Git. There's an article over here -- https://reidburke.com/updates/2025/01/the-case-for-kebab-case/ -- and the stylistic choice in this PR applies that. Typical Angular apps carry that naming convention as well. This does run counter to well adopted coding stylings from the likes of [AirBNB](https://github.com/airbnb/javascript/tree/master/react#naming), but I've already seen it happen in the last week (a file named `something.tsx` on disk, and a path reference of `Something.tsx` in an import).

dawsontoth:
All that to say, we can pivot the other direction, it's not that much work. Over time, we can get automated enforcement in place as well.

kriszyp (approved):
No need to apologize, this is great to pursue consistency here! And I agree kebab-case has merits. We have been using camelCase on core for many years without any real issues, so I'd empirically say Kent's dire warnings are... a little exaggerated :). If you feel like kebab-case would be better for studio, happy to discuss this. But matching core's camelCase is obviously nice for consistency. Thank you for working on this!

PR #485, opened 2025-07-02
Fix browse record selection

dawsontoth:
The instance browse page's "select a record" feature wasn't working, it wasn't loading the data. I simplified the logic a bit, so now it will load up properly.

While in there, I also extracted out sub-routes for the browse pages, preparing for the same treatment to the config users (and roles, soon, too).

### DIFF
--- a/src/features/instance/operations/queries/getSearchByHash.ts
+++ b/src/features/instance/operations/queries/getSearchByHash.ts
@@ -1,37 +1,15 @@
 import { instanceClient } from '@/config/instanceClient';
 import { queryOptions } from '@tanstack/react-query';
 
-// type SearchConditions = {
-// 	search_attribute: string;
-// 	search_type: string;
-// 	search_value: string;
-// };
-
-// type SearchByValueRequest = {
-// 	conditions?: [SearchConditions];
-// 	schema: string;
-// 	table: string;
-// 	search_attribute: string;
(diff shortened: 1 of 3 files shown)


src/features/instance/operations/queries/getSearchByHash.ts
```diff
@@ -1,37 +1,15 @@
...
+function getSearchByHashOptions(
+	isEditModalOpen: boolean,
+	instanceId: string,
+	schemaName: string,
+	tableName: string,
+	hashAttribute?: unknown,
+) {
 	return queryOptions({
-		queryKey: [instanceId, 'search_by_hash'] as const,
+		queryKey: ['search_by_hash', instanceId, schemaName, tableName, hashAttribute] as const,
```
kriszyp: Minor nit, but can you change `search_by_hash` to `search_by_id`? We changed the name of this operation a couple years ago (hash was an inaccurate name). It is also more correct to use `primaryKey` than `hashAttribute`, although that's just an internal variable here.

kriszyp (approved):

src/features/instance/operations/queries/getSearchByHash.ts
dawsontoth: @kriszyp Done!

BboyAkers (approved):

PR #487, opened 2025-07-03
Clusters Updates

BboyAkers:


dawsontoth:
@BboyAkers I rebased this, and then I implemented these bullet points:
- We'll group the clusters into "Running" and "Inactive".
- The cluster page will refetch every 10 seconds when you're looking at the page.
- If a cluster isn't self-managed, display a ‘Login’ or ‘View’ button. ‘View’ will take us directly to the instance browse page for that cluster with a simplified URL route (no `instance/{id}`).
- If it’s a self-managed cluster hide the ‘Login’ button and display only the ‘Instances’ button, which will take you to the list of instances for the cluster.
- Non-self-managed clusters have a new "Instances" context menu option to take you to the overall list.
- "Clone Ready" is now in the yellow/warning bucket.
- All the instance pages are now routed in a secondary way, when running in the cloud, a la `http://localhost:5173/#/orgs/{orgId}/clusters/{clusterId}/config`, which knows what to do to use the cluster's fqdn.

src/features/cluster/queries/getClusterInfoQuery.ts
```diff
@@ -3,18 +3,17 @@ import { queryKeys } from '@/react-query/constants';
...
+	const { data } = await apiClient.get(`/Cluster/${clusterId}` as '/Cluster/{id}');
 	return data as unknown as Cluster;
 }
 
-function getClusterInfoQueryOptions(clusterId?: string) {
+export function getClusterInfoQueryOptions(clusterId?: string, refetch?: boolean) {
 	return queryOptions({
 		queryKey: [queryKeys.cluster, clusterId],
-		queryFn: getClusterInfo,
+		queryFn: () => getClusterInfo(clusterId!),
```
dawsontoth, line 14: I keep going back and forth on this pattern... on the one hand, I really like not creating more contexts. By using the passed `context.queryKey`, memory management becomes simpler for the runtime. But as soon as we go to utilize `getClusterInfo` from somewhere else in the codebase, it's much harder to call...

src/features/cluster/routes.ts
```diff
@@ -3,25 +3,41 @@ import { clustersLayoutRoute } from '@/features/clusters/routes';
...
-		context.queryClient.ensureQueryData(getClusterInfoQueryOptions(params.clusterId));
+		return context.queryClient.ensureQueryData(getClusterInfoQueryOptions(params.clusterId));
 	},
 });
 
 const clusterIndexRoute = createRoute({
 	getParentRoute: () => clusterLayoutRoute,
 	path: '/',
 	component: ClusterIndex,
+	// TODO: We're going to want to resolve auth by instanceId and clusterId too for faster checks...
```
dawsontoth, line 21: 🤔 I'm still pondering this one.

dawsontoth (approved):

src/features/cluster/queries/getClusterInfoQuery.ts
BboyAkers: Hmmmmm 🤔, while memory management becomes simpler for the runtime, does the overhead of creating more context outweigh the simplicity of being able to utilize this elsewhere in the codebase? I personally say no but it's definitely something for us to be aware of as we improve the standardize patterns in the codebase overtime

dawsontoth (approved):
Proxy approval 😆

PR #489, opened 2025-07-03
[FIX] Fixed port not being used in url and toast success popup on error during instance login

BboyAkers:


dawsontoth (approved):

src/features/instance/operations/mutations/readInstanceLogin.ts
```diff
@@ -14,10 +15,11 @@ type InstanceLoginResponse = {
 
 const onInstanceLoginSubmit = async ({
 	instanceUrl,
+	port = 9925, // Default port for HarperDB instances
 	username,
 	password,
 }: InstanceLoginCredentials): Promise<InstanceLoginResponse> => {
-	const { data } = await instanceClient.post(`${instanceUrl}`, {
+	const { data } = await instanceClient.post(`${instanceUrl}:${port}`, {
```
dawsontoth: Actually, this bit isn't working for me locally when connecting to a self-managed instance, where the instanceUrl has the port embedded in it. It's resulting in the port being specified twice.
```suggestion
	const url = instanceUrl.includes(':') ? instanceUrl : `${instanceUrl}:${port}`;
	const { data } = await instanceClient.post(url, {
```

PR #490, opened 2025-07-07
New hook for tracking global sign in and out

dawsontoth:
This PR shifts our auth strategy into a nice `useAuth` hook for global and server specific auth tracking. Here's how it works, in short:

## 1. You `useAuth` in one of two ways:
```
export function useAuth(): AuthenticatedConnection;
export function useAuth(entity: Instance | Cluster | null): AuthenticatedConnection;
```
... which will give you back an authenticated connection:
```
export interface AuthenticatedConnection {
	isLoading: boolean;
	user: User | LocalUser | null;
}
```

## 2. That hook can be used all over the place performantly,
because under the hood it uses a singleton `authStore`. It `useEffect`s to leave the normal React lifecycle, using a standard event emitter style to watch for changes to auth, followed by `useState` to come back into the React lifecycle, and `useEffect`'s cleanup function.
```
export function useAuth(entity?: Instance | Cluster | null): AuthenticatedConnection {
	const key = useMemo(() =>
		calculateKeyFromEntity(entity), [entity]);
	const [connection, setConnection] = useState<AuthenticatedConnection>({ user: null, isLoading: true });
	useEffect(() =>
		authStore.listenToKey(key, connection => {
			setConnection(connection);
		}), [key]);
	return connection;
}
```

## 3. We turn that global, cluster or instance auth check into a specific server key (or URL)
Once we have that, we can see if this browser has logged into that server before. If it has, then we can make an API request to validate the cookie is still good! We track this in local storage, which lets us stay efficient with our API requests.

```
	private flagKeyAsSignedIn(key: AuthenticatedConnectionKey) {
		if (!this.potentiallyAuthenticatedKeys.includes(key)) {
			this.potentiallyAuthenticatedKeys.push(key);
			localStorage.setItem(this.localStorageKey, this.potentiallyAuthenticatedKeys.join(','));
		}
	}
```

## 4. Whenever we sign in or out, we can update the authStore:
```
authStore.setUser('global', null);
```
That will cascade to all active and relevant `useAuth` hooks.
```
	private async updateListeners(key: AuthenticatedConnectionKey): Promise<void> {
		await sleep(1);
		if (this.listeners[key]) {
			for (const listener of this.listeners[key]) {
				this.updateListener(key, listener);
			}
		}
	}
```

## 5. What's next?
Well, there are a few things we can do in the future:
1. This new code will be able to handle logging into a cluster, too. We need to introduce the right UI elements to make that happen still, but that will be pretty quick!
2. We can let the user log out of clusters and instances using this new code, too.
3. When we bookmark some authenticated page, like a particular instance logs page, and we aren't signed in, we can redirect the user to that instance's login form and redirect the user back to the logs page after logging in.
4. When logging out, we can log out of any clusters or instances they've logged into, trashing all the cookies so we don't leave holes in our security behind us.

😲 😮 😎 🍪 

This tackles both https://harperdb.atlassian.net/browse/STUDIO-237 and https://harperdb.atlassian.net/browse/STUDIO-230


src/features/auth/hooks/useInstanceLogoutMutation.ts
```diff
@@ -0,0 +1,29 @@
+import { useMutation } from '@tanstack/react-query';
+import { instanceClient } from '@/config/instanceClient';
+import { signOutOnSuccess } from '@/features/auth/hooks/signOutOnSuccess';
```
BboyAkers: Move this out of hooks/?

BboyAkers (approved):

kriszyp (approved):

PR #492, opened 2025-07-08
Sort orgs and clusters alphabetically

dawsontoth:
This will make it easier for me to find stuff when I’m clicking around, and it’s prettier... I also commented out the filtering and sorting UI elements for the time being. If we get to them as stretch goals, great! The code is there for us.

### DIFF
--- a/src/features/clusters/ClustersList.tsx
+++ b/src/features/clusters/ClustersList.tsx
@@ -1,13 +1,11 @@
 import { getRouteApi } from '@tanstack/react-router';
 import { ClusterCard } from '@/features/organization/components/ClusterCard';
-import { ChevronDown, Plus } from 'lucide-react';
+import { Plus } from 'lucide-react';
 import { Button } from '@/components/ui/button';
-import { Input } from '@/components/ui/input';
 import { getOrganizationQueryOptions } from '@/features/organization/queries/getOrganizationQuery';
 import { NewClusterModal } from '@/features/clusters/modals/NewClusterModal';
 import { useSuspenseQuery } from '@tanstack/react-query';
-import { notYetImplemented } from '@/lib/notYetImplemented';
-import { useState } from 'react';
+import { useMemo, useState } from 'react';
 
 const route = getRouteApi('');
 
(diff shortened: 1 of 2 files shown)


src/features/clusters/ClustersList.tsx
```diff
@@ -1,33 +1,32 @@
...
-import { useState } from 'react';
+import { useMemo, useState } from 'react';
 
 const route = getRouteApi('');
 
 export function ClustersList() {
 	const { organizationId } = route.useParams();
 	const { data: orgInfo, isSuccess } = useSuspenseQuery(getOrganizationQueryOptions(organizationId));
 	const [isNewClusterModalOpen, setIsNewClusterModalOpen] = useState(false);
+	const sortedClusters = useMemo(() => orgInfo?.clusters?.slice().sort((a, b) => a.name > b.name ? 1 : -1) || [], [orgInfo]);
```
dawsontoth, line 16: Could I have used our sorting on the server side for this? Well, yeah. But I didn't think about it on clusters since I did orgs first, and we extract that from user.roles...

src/features/clusters/ClustersList.tsx
```diff
@@ -1,33 +1,32 @@
...
-									<ChevronDown className="inline-block" />
-								</span>
-							</Button>
+							{/*<Input placeholder="Filter clusters by name" className="inline-block w-3/5 md:w-64" onChange={notYetImplemented} />*/}
+							{/*<Button className="inline-block w-2/5 md:w-auto md:ml-4" onClick={notYetImplemented}>*/}
+							{/*	Sort by A-Z*/}
+							{/*	<span>*/}
+							{/*		<ChevronDown className="inline-block" />*/}
+							{/*	</span>*/}
+							{/*</Button>*/}
```
BboyAkers, line 29: We can remove this code!

BboyAkers (approved):

src/features/clusters/ClustersList.tsx
dawsontoth: I figured we'll probably use it in the near future, if we get time for it!

dawsontoth: Want me to remove it, though?

PR #495, opened 2025-07-09
Parse errors for display better

dawsontoth:
https://harperdb.atlassian.net/browse/STUDIO-184

### DIFF
--- a/src/react-query/queryClient.ts
+++ b/src/react-query/queryClient.ts
@@ -1,9 +1,22 @@
 import { MutationCache, QueryCache, QueryClient } from '@tanstack/react-query';
 import { toast } from 'sonner';
+import { AxiosError } from 'axios';
 
-function errorHandler(errorMsg: Error) {
+function errorHandler(rawErr: unknown) {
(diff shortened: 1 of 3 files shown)


src/react-query/queryClient.ts
```diff
@@ -1,9 +1,22 @@
...
+import { AxiosError } from 'axios';
 
-function errorHandler(errorMsg: Error) {
+function errorHandler(rawErr: unknown) {
+	let errorMsg = 'We had some trouble!';
+	const axiosErr = rawErr as AxiosError<{ error?: string; message?: string; }>;
+	const otherErr = rawErr as { message?: string; };
+	if (axiosErr?.response?.data) {
+		if (axiosErr.response.data.error) {
+			errorMsg = axiosErr.response.data.error;
```
dawsontoth, line 13: This was the main culprit of the `[object Object]` errors showing up. We'd get `data`, which was an object, and it had something like: `{ "error": "failed because..." }`

BboyAkers (approved):

PR #501, opened 2025-07-15
Config Roles

BboyAkers:


dawsontoth (commented):
If you copy-paste from the latest Config/Users, you'll get most of what you need for the modals, too!

src/features/instance/config/roles/constants/tableDefinition.ts
```diff
@@ -0,0 +1,41 @@
...
+			translateSecondsToAgo(
+				(Date.now() - props.row.original.__createdtime__) / 1000,
+				props.row.original.__createdtime__
+			),
+	}),
+	{
+		header: 'Updated',
+		accessorKey: '__updatedtime__',
+		enableSorting: false,
+		cell: (props) =>
```
dawsontoth, line 29: I _think_ the `columnHelper.display` helps with type inference when you're doing stuff with `cell (props) =>` 🤷 I don't quite remember though.

src/features/instance/config/roles/index.tsx
```diff
@@ -1,5 +1,158 @@
...
+						onSuccess: () => {
+							toast.success('Role updated successfully!');
+						},
+						onError: (error) => {
+							toast.error(`Failed to update role: ${error.message}`);
+						},
+					}
+				);
+				// Call the API to update the role with parsedPermissions.
+				// This part is not implemented here, but you would typically call an API service.
```
dawsontoth: Hi AI 😆

src/features/instance/config/roles/index.tsx
```diff
@@ -1,5 +1,158 @@
...
+			}
+			void refetch();
+			onSelectRole(undefined);
```
dawsontoth: You'll probably want these in the unSuccess callback up above, instead of down here!

dawsontoth (commented):
Looking real good, Mr. Austin, sir! 🥳

src/features/instance/config/roles/index.tsx
BboyAkers: 😂 😂  yeahhhh totally went crazy with some autocomplete last night

src/features/instance/config/roles/index.tsx
BboyAkers: so we don't want this triggered for both?

update:
nvm totally understand what you meant 😅

src/features/instance/config/roles/constants/tableDefinition.ts
BboyAkers: added!

dawsontoth (approved):
Looks great, in theory! Can't test it 😆

PR #505, opened 2025-07-15
[STUDIO-248] Allow customizing the credentials during initial setup

dawsontoth:
I started off trying to prevent the password manager from setting the username, but @Devin-Holland had the great idea to actually let the user change their username 😄 so let's do that!

I also found and fixed a bug preventing the authStore from storing that initial user properly without having to refresh the page (which also showed up when logging out from the cluster or instance).

https://harperdb.atlassian.net/browse/STUDIO-248

### DIFF
--- a/src/features/cluster/ClusterSetPassword.tsx
+++ b/src/features/cluster/ClusterSetPassword.tsx
@@ -1,3 +1,4 @@
+import { defaultClusterUsername } from '@/config/constants';
 import { getRouteApi, Navigate, useNavigate } from '@tanstack/react-router';
 import { useForm } from 'react-hook-form';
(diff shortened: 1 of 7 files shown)


Devin-Holland (approved):

src/features/cluster/ClusterSetPassword.tsx
```diff
@@ -112,7 +117,6 @@ export function ClusterSetPassword() {
 										<FormLabel>Username</FormLabel>
 										<FormControl>
 											<Input
-												disabled={true}
 												autoComplete="username"
 												type="text"
 												className="bg-purple-400 border-purple-400 dark:bg-black dark:border-black"
```
BboyAkers: remove the classes
```suggestion
```

PR #509, opened 2025-07-17
[STUDIO-174] Cluster Form: Plans and Regions Cascading Dropdowns

dawsontoth:
This is mostly focused on the cascade of plan selection to region selection. There is more work that will need to be done, and temporary API mocking for regions and plans that we'll need to remove.

You can create a cluster with the new form using these selections:
1. Enter a valid name, and it should calculate an abbreviated name for you.
2. Select "Free" as the deployment type, and leave the performance at "Basic".
3. Change the region to "US-NE", and leave the latency at "Low".
4. Click Create!

Here is what I've changed, high level:
1. You can Ctrl+Opt+N to open the create modal from the clusters page.
2. We show the price in the top right of the form.
3. Plan is picked at the top level, which cascades down into a performance, and then regions and latencies. There's more work to be done for self-managed, and likely for org and cluster dedicated.
4. I used the NewClusterSchema types in more spots to be more DRY.
5. I split the usage limits out into a new component, which animates as an accordion. We'll have more work on what shows up in that thing.
6. I mocked the region and plan lookups for this form, so that we could have all the new field values from the big central-manager PR to test things out.
7. We'll calculate a placeholder for the abbreviated name based on the system name. The field is optional now too, which lets us default it for the user if they're happy with the calculated value.
8. I added a few new libs that will be useful for us, and some more unit tests.
9. Sometimes the API gives a direct error string as its response, not wrapped in JSON. I handled that case in our query error handler too.

https://harperdb.atlassian.net/browse/STUDIO-174


src/features/clusters/modals/NewClusterModal/PriceDisplay.tsx
```diff
@@ -0,0 +1,6 @@
+export function PriceDisplay({ price }: { readonly price?: string }) {
+	const parts = (price || '$0.00').split('$')[1].split('.');
```
BboyAkers: Should we consider using this instead? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat

BboyAkers (approved):

src/features/clusters/modals/NewClusterModal/PriceDisplay.tsx
dawsontoth: @BboyAkers Great idea, done!

PR #517, opened 2025-07-22
Expand the types for the instance schema and table responses

dawsontoth:
I was hoping I could skirt by the CORE ticket 2836 by switching us to use the `describe_all` API, but it shows the same bug. When doing that, I cleaned up and unified the types we use between the two APIs. Now there's just one set of types for these two APIs, and they should be comprehensive.

`buildInstanceDataStructure` also proved to not be necessary. It was sorting the names, but we do that anyway inside src/features/instance/browse/index.tsx


BboyAkers (approved):
🔥

PR #521, opened 2025-07-23
Org Roles

BboyAkers:


src/components/Navbar.tsx
```diff
@@ -77,32 +93,65 @@ function MobileNav({ signOut }: { signOut: () => void }) {
...
+						{!isLocalStudio && (
+							<>
+								<NavigationMenuItem>
+									<NavigationMenuLink asChild>
+										<Link to="/orgs" className="flex-row items-center" activeProps={activeLinkProps}>
+											<BuildingIcon /> Organizations
+										</Link>
+									</NavigationMenuLink>
+								</NavigationMenuItem>
+								{organizationId ? (
```
dawsontoth, line 137: Looks good, but how will mobile nav get at these?

src/features/organization/config/mutations/addOrganizationRole.ts
```diff
@@ -0,0 +1,52 @@
...
+}
+
+export interface AddOrgRoleFormData {
+	roleName: string;
+	organizationId: string;
+	updateOrganization: boolean;
+	deleteOrganization?: boolean;
+	roles: RolesPermissions;
+	clusters: ClustersPermissions;
+}
```
dawsontoth: Take a look at `SchemaRole`

src/features/organization/config/mutations/updateOrganizationRole.ts
```diff
@@ -0,0 +1,36 @@
...
+	deleteOrganization: z.boolean(),
+});
+
+export async function onUpdateOrganizationRole({
+	roleId,
+	updatedRoleInfo,
+}: {
+	roleId: string;
+	updatedRoleInfo: Partial<OrganizationRole>;
+}): Promise<OrganizationRole> {
```
dawsontoth: I believe this will return `SchemaRole`, but I'm not sure.

src/features/organization/config/routes.ts
```diff
@@ -0,0 +1,48 @@
+// import { createRoute } from '@tanstack/react-router';
```
dawsontoth: I think putting these routes within the org routes is fine, breaking it down to sub-routes like this is likely overkill, so this file could be deleted IMO...

src/components/Navbar.tsx
BboyAkers: Great catch!

src/features/organization/mutations/deleteOrganizationRole.ts
```diff
@@ -0,0 +1,17 @@
+import { apiClient } from '@/config/apiClient';
+import { useMutation } from '@tanstack/react-query';
+
+export async function onDeleteOrganizationRole({ roleId }: { roleId: string }) {
+	const { data } = await apiClient.delete(`/Role/${roleId}` as '/Role/{id}', {
+		//@ts-expect-error would love to figure this out
+		id: roleId,
```
dawsontoth, line 7: Are you sure we need to send the id in the body in addition to sending it in the path?

src/features/organization/mutations/addOrganizationRole.ts
```diff
@@ -0,0 +1,29 @@
...
+		.regex(/^[a-zA-Z_]+$/, {
+			message: 'Role must contain only letters and underscores.',
+		})
+		.max(30, { message: 'Role name must be less than 30 characters.' }),
+	updateOrganization: z.boolean(),
+	deleteOrganization: z.boolean(),
+});
+
+export async function onAddOrganizationRoleSubmit(formData: SchemaRole) {
+	const { data } = await apiClient.post('/Role' as '/Role/', formData);
```
dawsontoth, line 21: Does posting to /Role/ work? Then you won't need the cast.

src/features/organization/queries/getOrganizationRoleInfo.ts
```diff
@@ -0,0 +1,56 @@
+import { queryOptions } from '@tanstack/react-query';
+import { apiClient } from '@/config/apiClient';
+import { SchemaRole } from '@/lib/api.gen';
+import { queryKeys } from '@/react-query/constants';
+
+export interface GetOrganizationRoleInfoResponse extends SchemaRole {
```
dawsontoth, line 6: Can you put this type in the patch file instead?

src/features/organization/roles/modals/EditOrganizationRoleModal.tsx
```diff
@@ -0,0 +1,261 @@
...
+								<DialogFooter className="col-span-2">
+									<div className="flex justify-between w-full">
+										<Button
+											variant="destructiveOutline"
+											className="rounded-full"
+											onClick={() => setIsConfirmingRoleDeletion(true)}
+										>
+											Delete Role
+										</Button>
+										<Button variant="submit" className="rounded-full" disabled={!isValidJSON}>
```
dawsontoth: Can we disable this while the mutation is pending too?

dawsontoth (approved):

src/features/organization/mutations/addOrganizationRole.ts
BboyAkers: It works!

src/features/organization/mutations/deleteOrganizationRole.ts
BboyAkers: The api doc shows I need to

src/features/organization/queries/getOrganizationRoleInfo.ts
BboyAkers: For sure!

BboyAkers: ended up leaving it here since I significantly reduced the amount of types under this

PR #529, opened 2025-07-30
Add Studio Version

BboyAkers:


dawsontoth (approved):

dawsontoth:
I'm curious if we set this in the deployed environments, too!

BboyAkers:
> I'm curious if we set this in the deployed environments, too!

Hmmm that'd be very nice! 🤔

PR #539, opened 2025-08-08
[STUDIO-281] Fixes the error notification "primaryKeyfromAnotherTable" is not a defined attribute

BboyAkers:
## Ticket:

[STUDIO-281](http://harperdb.atlassian.net/browse/STUDIO-281)

## Overview:

Fixes the error notification "primaryKeyfromAnotherTable" is not a defined attribute.

## Addition Info:
Would love to have a better alternative than this solution @dawsontoth. Please give suggestions!

[STUDIO-281]: https://harperdb.atlassian.net/browse/STUDIO-281?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/features/instance/browse/components/BrowseDataTableView.tsx
+++ b/src/features/instance/browse/components/BrowseDataTableView.tsx
@@ -42,10 +42,11 @@ export function BrowseDataTableView() {
 
 	const { dataTableColumns, hash_attribute } = formatBrowseDataTableHeader(describeTableData);
 	const [isAddModalOpen, setIsAddModalOpen] = useState(false);
-	const [sortTableDataParams, setSortTableDataParams] = useState({
+	let sortTableDataParams = {
 		attribute: hash_attribute,
 		descending: false,
-	});
+	};
+
 	const sortingState = useMemo(
 		() => [
 			{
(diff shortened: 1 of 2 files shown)


src/features/instance/browse/components/BrowseDataTableView.tsx
```diff
@@ -142,10 +143,10 @@ export function BrowseDataTableView() {
 		setIsEditModalOpen(!isEditModalOpen);
 	};
 	const onColumnClick = (accessorKey: string, isAscending: boolean) => {
-		setSortTableDataParams({
+		sortTableDataParams = {
 			attribute: accessorKey,
 			descending: !isAscending,
-		});
+		};
```
dawsontoth, line 149: By getting rid of the hook call, there's uncertainty on if React will re-render or not based on the mutation of the object. Remember that `let sortTableDataParams = {` won't live for very long, so by changing it in this callback, you're changing an ephemeral copy of it. Because it doesn't trigger a re-render, you're avoiding the error, but only by not triggering the desired behavior (I believe).

dawsontoth: My guess is that we need to proactively reset the sorted attribute when changing tables. Storing the sort state in the URL might be a good solution to that. 🤔

PR #544, opened 2025-08-11
Shift browse props

BboyAkers:


dawsontoth:
In https://github.com/HarperDB/hdbms/pull/543 I refactored the names of `databases` and `tables`, which will cause conflicts here, sorry!

dawsontoth (approved):

PR #562, opened 2025-08-19
Instance Restarting in /applications

BboyAkers:


src/features/instance/applications/editor/components/TextEditorView/index.tsx
```diff
@@ -41,25 +80,35 @@ export function TextEditorView() {
...
+						disabled={
+							!selectedFolderFile.filePath ||
+							(selectedFolderFile.filePath && updateFileContent == selectedFolderFile.content) ||
+							isSavingFile
+						}
+					>
+						<Save />
+						<span className="ms-1">Save</span>
+					</Button>
+					<Button
```
kriszyp: Can we include a tooltip giving some more information about the impact of restarts? I think it is probably important to inform users that the full restart on the config page has about 100x the impact of this restart. Maybe a tooltip with verbiage:

> Performs a rolling restart of all the threads running this service to ensure all application changes are applied. This should not result in any downtime, but may incur some slower performance during the restart process.

In contrast, it would be nice to add a tooltip to the config page restart stating:

> This will perform a full restart of the entire Harper service and will incur downtime while the restart is taking place. Please use the restart on the application tab for restarting without downtime.

src/features/instance/applications/editor/components/TextEditorView/index.tsx
```diff
@@ -28,9 +31,45 @@ export function TextEditorView() {
...
 	const [language, setLanguage] = useState('javascript');
 	const [updateFileContent, setUpdateFileContent] = useState<string>(selectedFolderFile.content || '');
+	const { clusterId, instanceId } = useParams({ strict: false });
+	const targetId = instanceId ?? clusterId;
+	const targetNoun = instanceId ? 'Instance' : 'Cluster';
+	const { mutate: restartInstance, isPending: isRestartInstancePending } = useUpdateRestartInstance();
 
 	const crumbPath = selectedFolderFile.filePath.split('/').slice(1).join('/').replace(/\//g, ' > ');
 
+	const restartingInstance = () => {
```
kriszyp, line 42: I suppose `restartInstance` is a bit of a misnomer since it is being used to restart a cluster and _not_ an instance when in cluster mode, right?

src/features/instance/applications/new/ImportProjectForm.tsx
```diff
@@ -39,12 +44,45 @@ export function ImportProjectForm() {
 		},
 	});
 
+	const restartingInstance = () => {
+		const toastId = toast.loading('Restarting', {
+			description: `Restarting ${targetNoun.toLowerCase()}. This may take up to 60 seconds.`,
```
kriszyp: Our replicated restart actually returns a job id, so the UI can monitor the progress of the restart and let the user know when it is finished (rather than just generically saying it will be a minute).
Not something we need for v1 though.

dawsontoth (approved):

PR #566, opened 2025-08-21
Don’t rely on global base url set in axios for instances

dawsontoth:
We were relying on the Axios baseURL getting set when the instance would load. This mostly worked, but was problematic with race conditions, switching between instances and clusters, and preloading. It would give the appearance of trying to load information across those boundaries.

Instead, we can be a bit more verbose (ok, a lot more verbose), and require that the client be passed to us. I'm still pondering how to clean this up, and the types being very strong will make future refactors quite easy (albeit broad reaching).

As a part of this, I noticed and added strong types to all `useParams` variants in the codebase. Without the strong types, these were of the type `any`, which was making finding broken signatures very difficult for me. 😄 

Some of the instance operations weren't located in the proper instance folder for queries or mutations. And a few that _were_ in there, weren't actually instance specific, but rather, cluster specific! Straightened all that out, and I iterated through all of the operations to make them nice and consistent (only export what is used, use interfaces, use named functions).

https://harperdb.atlassian.net/browse/STUDIO-282
https://harperdb.atlassian.net/browse/STUDIO-324


BboyAkers:
Really solid!!! 🚀

BboyAkers (approved):

BboyAkers:
This PR has a ton of awesome changes 🔥 😊 that may cause some merge conflicts. Thinking of breaking a branch and using this as the source of truth for my current task until we merge the other PRs.

Also, thinking of the PR order:
 1. Devin's Payment PR.
 2. Your Stripe PR
 3. This PR
 4. Everything else

How does this sound?

PR #567, opened 2025-08-21
Available Deployment and Region Types Filters

BboyAkers:


src/features/clusters/queries/getRegionLocationsQuery.ts
```diff
@@ -1,15 +1,15 @@
...
-	const { data } = await apiClient.get(`/Region/`);
+async function getRegionLocations(orgId: string) {
+	const { data } = await apiClient.get(`/Region/?availableHosts=true&organizationId=${orgId}`);
 	return data;
 }
 
-function getRegionLocationsOptions() {
+function getRegionLocationsOptions(orgId: string) {
 	return {
 		queryKey: [queryKeys.cluster, 'regionLocations'],
```
dawsontoth: Try to remember to always add parameters to the queryKey.

dawsontoth (commented):
Most of this is whitespace changes that eslint doesn't dictate, I think your vscode is being overly aggressive in formatting stuff...

src/features/clusters/queries/getRegionLocationsQuery.ts
BboyAkers: Gotcha!

dawsontoth (approved):
OK I think I got this working properly... we've got quite a few effects that change the selections...

If you pick a new deployment type that changes the available regions, it will look through your selected regions. If it finds one that isn't available anymore, we'll wipe it out, and force the user to select it again.

PR #575, opened 2025-08-25
Handle create button in empty apps

dawsontoth:
Buttons work better when they actually do something 😆 

https://harperdb.atlassian.net/browse/STUDIO-354


src/features/instance/applications/index.tsx
```diff
@@ -7,19 +7,13 @@ export function ApplicationsIndex() {
...
 					<Button className="w-full py-10 text-lg" variant="positiveOutline">
 						<Edit />
 						Edit Applications
 					</Button>
 				</Link>
-				<Link
-					to={`new`}
-					className="w-full max-w-80"
-				>
+				<Link to="new" className="w-full max-w-80">
```
dawsontoth, line 16: These changes weren't necessary, but I thought it would be nice to simplify them ever so slightly, and have them for reference when considering the change to the EmptyApplicationsView. 🥳

BboyAkers (approved):

PR #576, opened 2025-08-25
Tweaks to Browse

dawsontoth:
I made a few improvements to the browse page:

1. The "[primaryKey](https://harperdb.atlassian.net/browse/STUDIO-288)" when creating a table is now optional, and it will default to "id" (which is the placeholder).
2. The [current and total pages](https://harperdb.atlassian.net/browse/STUDIO-310) will now show up in the footer.
3. I de-prioritized the database and table deletion buttons, moving them away from the most important sight lines.
4. I used all of the available viewport, fixing the header and footer in place and scrolling the table when we have more records than we can show on the screen (except at smaller viewports).
5. The database creation is now in a modal, which will pop open automatically once when you have 0 databases.
6. I added access keys for creating databases, tables, and records.
7. When creating or editing a record, if you don't make changes, you can hit ESC to close the modal.

1-please-createa-a-database
2-create-a-database
3-create-a-table
4-browsing-data

### DIFF
--- a/src/components/ConfirmDeletionModal.tsx
+++ b/src/components/ConfirmDeletionModal.tsx
@@ -6,8 +6,10 @@ import {
 	DialogFooter,
 	DialogHeader,
 	DialogTitle,
+	DialogTrigger,
 } from '@/components/ui/dialog';
 import { ArrowLeft, Trash, TriangleAlert } from 'lucide-react';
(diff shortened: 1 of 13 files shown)


kriszyp (approved):
Awesome!

BboyAkers (approved):

PR #580, opened 2025-08-28
Logs Page Refactor

BboyAkers:
# Tickets:
[STUDIO-177](https://harperdb.atlassian.net/browse/STUDIO-177)
[STUDIO-309](https://harperdb.atlassian.net/browse/STUDIO-309)
[STUDIO-318](https://harperdb.atlassian.net/browse/STUDIO-318)
[STUDIO-319](https://harperdb.atlassian.net/browse/STUDIO-319)

# Overview:
Fixes multiple issues with the application logs page:
- Clearing all filter inputs
- Time selection in the start and end input fields
- Refresh logs
- Auto-refresh logs

[STUDIO-177]: https://harperdb.atlassian.net/browse/STUDIO-177?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[STUDIO-309]: https://harperdb.atlassian.net/browse/STUDIO-309?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[STUDIO-318]: https://harperdb.atlassian.net/browse/STUDIO-318?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[STUDIO-319]: https://harperdb.atlassian.net/browse/STUDIO-319?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/features/instance/log/components/LogsFiltersForm.tsx
+++ b/src/features/instance/log/components/LogsFiltersForm.tsx
@@ -1,157 +1,134 @@
-import { Button } from "@/components/ui/button";
-import { Form } from "@/components/ui/form/Form";
-import { FormControl } from "@/components/ui/form/FormControl";
-import { FormField } from "@/components/ui/form/FormField";
-import { FormItem } from "@/components/ui/form/FormItem";
-import { FormLabel } from "@/components/ui/form/FormLabel";
-import { FormMessage } from "@/components/ui/form/FormMessage";
-import { Input } from "@/components/ui/input";
-import { Select, SelectTrigger, SelectValue, SelectContent, SelectGroup, SelectItem } from "@/components/ui/select";
+import { Button } from '@/components/ui/button';
+import { Form } from '@/components/ui/form/Form';
+import { FormControl } from '@/components/ui/form/FormControl';
+import { FormField } from '@/components/ui/form/FormField';
+import { FormItem } from '@/components/ui/form/FormItem';
+import { FormLabel } from '@/components/ui/form/FormLabel';
+import { FormMessage } from '@/components/ui/form/FormMessage';
+import { Input } from '@/components/ui/input';
+import { Select, SelectTrigger, SelectValue, SelectContent, SelectGroup, SelectItem } from '@/components/ui/select';
+import { UseFormReturn } from 'react-hook-form';
+import { LogFiltersFormSchema } from '@/features/instance/operations/queries/getReadLog';
+import z from 'zod';
 
-export function LogsFiltersForm({form, resetFilters, submitFilters}) {
+export function LogsFiltersForm({
+	form,
+	resetFilters,
+	submitFilters,
+}: {
+	form: UseFormReturn;
+	resetFilters: () => void;
+	submitFilters: (data: z.infer<typeof LogFiltersFormSchema>) => void;
+}) {
 	return (
 		<div>
 			<Form {...form}>
-					<form onSubmit={form.handleSubmit(submitFilters)} className="flex-col space-y-5">	
-						<FormField
-							control={form.control}
-							name="limit"
-							render={({ field }) => (
-								<FormItem>
-									<FormLabel>Log Limit</FormLabel>
-									<Select
-										onValueChange={(value) => {
-											// field.onChange(parseInt(value));
-											field.onChange(value);
-										}}
-										value={field.value}
-									>
-										<SelectTrigger className="w-full">
-											<SelectValue placeholder="Select log limit" />
-										</SelectTrigger>
-										<SelectContent>
-											<SelectGroup>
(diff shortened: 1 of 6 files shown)


BboyAkers:
Still needing to add the refresh logs and auto-refresh logs buttons.

src/features/instance/log/index.tsx
```diff
@@ -86,28 +78,34 @@ const isValidDateRange = (startDate?: Date, endDate?: Date) => {
...
 export function Logs() {
-	const [logFilters, setLogFilters] = useState<z.infer<typeof LogFiltersSchema>>({
-		limit: 1000,
-		order: 'desc',
+	const [logFilters, setLogFilters] = useState<z.infer<typeof LogFiltersFormSchema>>({
+		limit: '1000',
+		level: 'undefined',
+		from: '',
+		until: '',
+		order: 'asc',
```
dawsontoth: Idea: extract this and the default values on lines 104-108 to a constant, so they're only specified in one place. Ditto on reset filters.

src/features/instance/log/components/LogsFiltersForm.tsx
```diff
@@ -0,0 +1,134 @@
...
+	return (
+		<div>
+			<Form {...form}>
+				<form onSubmit={form.handleSubmit(submitFilters)} className="flex-col space-y-5">
+					<FormField
+						control={form.control}
+						name="limit"
+						render={({ field }) => (
+							<FormItem>
+								<FormLabel>Log Limit</FormLabel>
```
dawsontoth: Another idea: the start and end date have a `:` after the label, should we use that everywhere?
```
								<FormLabel>Log Limit:</FormLabel>
```

src/features/instance/log/index.tsx
```diff
@@ -86,28 +78,34 @@ const isValidDateRange = (startDate?: Date, endDate?: Date) => {
 };
 
 export function Logs() {
-	const [logFilters, setLogFilters] = useState<z.infer<typeof LogFiltersSchema>>({
-		limit: 1000,
-		order: 'desc',
+	const [logFilters, setLogFilters] = useState<z.infer<typeof LogFiltersFormSchema>>({
+		limit: '1000',
```
dawsontoth: What do you think about defaulting this to 100? 1000 feels a bit large, my browser starts to lag with it a bit. A mobile browser would get impacted even more.

dawsontoth (approved):

src/features/instance/log/index.tsx
```diff
@@ -135,164 +133,19 @@ export function Logs() {
...
-			from: undefined,
-			until: undefined,
-			order: 'desc',
+			limit: '1000',
+			level: 'undefined',
+			from: '',
+			until: '',
+			order: 'asc',
 		});
 	};
```
dawsontoth, line 150: When you implement the refresh button, the `useRefreshClick` hook we have will be helpful for mimicking the behavior we have elsewhere!

src/features/instance/log/index.tsx
kriszyp: I agree, 100 is a better default, 1000 puts too much strain on the server (at least as the default).

src/features/instance/log/index.tsx
```diff
@@ -135,164 +133,19 @@ export function Logs() {
 	const resetFilters = async () => {
 		form.reset();
 		setLogFilters({
-			limit: 1000,
-			level: undefined,
-			from: undefined,
-			until: undefined,
-			order: 'desc',
+			limit: '1000',
+			level: 'undefined',
```
kriszyp: we are really converting this to a string called 'undefined'?

src/features/instance/log/index.tsx
BboyAkers: I was thinking the exact same thing tbh.

src/features/instance/log/index.tsx
BboyAkers: The inputs for values are strings so this simplifies me from having to convert it to type `undefined`. When passed into the payload if it's `'undefined'` it won't send with it. I could convert it to type `undefined` but it inherently becomes unnecessary

PR #584, opened 2025-08-29
Studio 264 2

kriszyp:
This the rest of https://github.com/HarperDB/hdbms/pull/578, as I think Dawson did a force-push that removed this commit (and I have merged/resolved the conflict with harper/fabric).

https://harperdb.atlassian.net/browse/STUDIO-264

### DIFF
--- a/src/features/clusters/upsert/ClusterDetails.tsx
+++ b/src/features/clusters/upsert/ClusterDetails.tsx
@@ -61,7 +61,7 @@ export function ClusterDetails({
 	const isSelfManaged = selectedDeployment === 'Self-Hosted';
 
 	return (<>
-		<div className="grid grid-cols-3 gap-6 text-white md:grid-cols-6 overflow-auto max-h-[calc(100vh-theme(spacing.52))]">
+		<div className="grid grid-cols-3 gap-6 text-white md:grid-cols-6">
 			<FormField
 				control={form.control}
 				name="systemName"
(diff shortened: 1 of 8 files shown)


src/features/clusters/upsert/ClusterDetails.tsx
```diff
@@ -242,9 +242,9 @@ export function ClusterDetails({
 				/>)
 			}
 
-			{selectedPlan?.planLimits && selectedPlan.resourcesPerInstance && (
-				<ResourcesPerInstance planLimits={selectedPlan.planLimits} resourcesPerInstance={selectedPlan.resourcesPerInstance} />
-			)}
+			{selectedPlan?.planLimits?.totalReadCount! > 0 ? (
```
dawsontoth: I like it! Alright if I take this PR over? I'm going to move this logic and text into the component because we use the component in a couple places now. And I can fix the lint errors.

kriszyp: Sounds great, thank you!

dawsontoth (approved):
OK I made a few minor adjustments.
1. The largest thing I did was use the expiration months in the other places of the codebase that were hard coded to say 3 months. Now we'll respect whatever the selected plan dictates to us.
2. I added a `isPositive` function that narrows the type too, letting me remove the type assertions. It also has the benefit of communicating that we want to make sure the number is defined, but also that it isn't negative (because `-1` communicates unlimited, right?) so a future dev won't say "hey, we could replace this with a truthy check!" and break stuff.
3. I tweaked the wrapping a little bit so that it will `<br />` where Kris wants, but also not do that when we're on the smallest viewports. That would result in fragmented lines.
4. I added a `<ContactUs />` component for that anchor tag we had in a few spots, mostly so that our line wraps can be cleaner 😆 
If you hide whitespace in this PR, the diff will be a lot easier to see.

BboyAkers (approved):

PR #602, opened 2025-09-04
feat: Set package.json down to 0.0.0

dawsontoth:
BREAKING CHANGE: Let’s bump up to 2.0.0, ok?

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
 	"name": "fabric_ui",
 	"private": true,
-	"version": "5.0.0",
+	"version": "0.0.0",
 	"type": "module",
 	"scripts": {
 		"dev": "vite",


BboyAkers:
The package.json version coincides with the Harper Studio version. Current studio version is `v4.8.20`. 5.0.0 does seem like a valid package version

dawsontoth:
@BboyAkers Take a look at what https://github.com/HarperDB/hdbms/pull/591 is doing, and go look at the version shown in https://stage.studio.harperfabric.com/#/ to see more.

dawsontoth:
If we want, we can bump semantic release up to v5.0.0 manually, but we want to wipe out the package.json version regardless. I can do the manual bump via pushing a tag.

BboyAkers:
Gotcha! Let's do 2.0.0. We should check with @kriszyp to make sure this isn't an issue.

BboyAkers (approved):

PR #656, opened 2025-09-08
[STUDIO-13] Revisit Validation Rules

dawsontoth:
In the first commit, I went through all of our Zod validations, and I made the messages be consistent in a few ways:
1. We'll use `nonempty` instead of `min(1`.
2. We'll consistently say something like `{name} cannot be longer than {length} characters` instead of `{name} must be less than {length} characters` (since the latter was technically inaccurate at the maximum length).
3. Error messages will end with a `.`.
4. I extracted some common, shared schemas, and referenced them from others to reduce duplication.
5. I removed spots where we had both `.string({ error })` and `.min(8, { error })` or `.nonempty({ error })` to simply have `.string()` followed by the more precise error. The spots with `.email` have both error spots, since `email` has a regex with it.

Then in the second commit, I looked carefully at the source code for both core and central-manager. Then I adjusted our validation of a few things to match. Notably changes were:
1. to the maximum password lengths (I removed them),
2. and the database and table name lengths and allowed characters (no backticks or forward slashes).
3. Usernames for instances can contain just about anything, but we'll lowercase and trim them now.
4. When signing in, we don't enforce any min or max on usernames or passwords. They have to specify them, but we don't do any other checks. This is to prevent us from accidentally blocking a valid username or password.
5. When signing up or adding a user, I did enforce a higher minimum password length of 8 still, despite the servers not checking that. If someone wants to bypass that and set ` ` as their password, they can do that via the API.

https://harperdb.atlassian.net/browse/STUDIO-13


BboyAkers (approved):
Approving, but I'm curious to know if we have a max username and max password length in Harper core for our instance login? 🤔

dawsontoth:
@BboyAkers not that I was able to find. We only check if it's being provided.

dawsontoth:
And I successfully made a username that was my email, for example.

PR #670, opened 2025-09-10
[STUDIO-402] Fix Preload Error by Preventing on Mobile

dawsontoth:
This was an odd one. If you do this on certain phones, tanstack will preload incorrectly (specifically the `$orgId/$orgId` when looking at the org page). The second $orgId is used as the cluster id, which we try to lookup as the cluster, which 404s.

For now, let's detect touch based browsers, and disable preloading for them.

https://harperdb.atlassian.net/browse/STUDIO-402

### DIFF
--- a/src/lib/browserIsTouchBased.ts
+++ b/src/lib/browserIsTouchBased.ts
@@ -1,0 +1,3 @@
+export function browserIsTouchBased(): boolean {
+	return 'ontouchstart' in window || navigator['maxTouchPoints'] > 0;
+}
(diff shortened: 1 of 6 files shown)


src/lib/browserIsTouchBased.ts
```diff
@@ -0,0 +1,3 @@
+export function browserIsTouchBased(): boolean {
+	return 'ontouchstart' in window || navigator['maxTouchPoints'] > 0;
```
BboyAkers, line 2: 🔥

BboyAkers (approved):

PR #672, opened 2025-09-10
feat: Limit what we deploy to only the built contents

dawsontoth:
This does a few things fancy for us:
1. We only need the build `web` folder! Nothing else.
2. We can isolate it, plus our config.yaml, in to a new `deploy` folder at build time.
3. We don't need the notFound handler in the config. That was when we were doing browser based routing instead of hash routing.
4. Then we don't need to remove our `node_modules` anymore, resulting in cache _hits_ in our workflows! That'll speed stuff up too.
5. Harper needs a package.json, but we don't need any of the node_modules once built, so we can create an bare `{}` package.json instead in the `deploy` folder.
6. Once we `cd deploy`, we can do our normal harper deploy!

### DIFF
--- a/.github/workflows/deploy-dev.yaml
+++ b/.github/workflows/deploy-dev.yaml
@@ -40,5 +40,9 @@ jobs:
         run: npm install -g harperdb@4.6.0
       - name: Deploy to dev
         run: |
-          rm -rf node_modules # Clean up node_modules, harper will install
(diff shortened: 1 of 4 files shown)


Ethan-Arrowood (approved):
nice

BboyAkers (approved):

PR #674, opened 2025-09-11
Adjust resource limit for regions and adjust deployment type descriptions

kriszyp:
* Move resource limit reporting to region level and adjust for distribution size
* Add some more information to deployment description

I don't know if this is the right way to pass the selected region.

### DIFF
--- a/src/components/ui/select.tsx
+++ b/src/components/ui/select.tsx
@@ -89,12 +89,12 @@ function SelectItem({ className, children, ...props }: React.ComponentProps<type
 			)}
 			{...props}
 		>
+			<SelectPrimitive.ItemText>{children}</SelectPrimitive.ItemText>
 			<span className="absolute right-2 flex size-3.5 items-center justify-center">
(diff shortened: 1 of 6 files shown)


dawsontoth:
@kriszyp I rebased this, removed the ResourcesPerInstance from the billing view (we'd have to show it for each region) to restore the build, and then I adjusted the dropdowns to be split into two lines so that we can responsively size better. See screenshots below for examples. Look ok?

Screenshot 2025-09-11 at 11 08 41 AM
Self-managed was slightly harder to split (without the clear `Title (Description)` pattern), but worked IMO:
Screenshot 2025-09-11 at 11 12 44 AM

dawsontoth (approved):

dawsontoth:
I also prefixed your commit with `feat: ` to make commitlint happy 😄

BboyAkers (approved):

kriszyp:
Awesome, I love it, thank you again for cleaning up my messes!

PR #675, opened 2025-09-11
feat: Restore Husky

dawsontoth:
This should work now that we've sorted out our deployments running `npm install` when we don't want them to.

I verified this is working by pushing this to dev. https://github.com/HarperDB/hdbms/actions/runs/17649700845/job/50157074931 The run failed because we haven't configured dev yet, but it got all the way to the actual deployment, which is what we want to see! (`TypeError: Invalid URL` is a symptom of dev not being configured yet.)

### DIFF
--- a/.github/workflows/deploy-dev.yaml
+++ b/.github/workflows/deploy-dev.yaml
@@ -24,6 +24,8 @@ jobs:
         uses: actions/setup-node@a0853c24544627f65ddf259abe73b1d18a591444 # v5.0.0
         with:
           cache: 'pnpm'
+      - name: Disable Husky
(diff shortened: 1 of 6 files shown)


BboyAkers (approved):

kriszyp:
I think I saw husky complain that a commit didn't have a "subject". What is a commit subject vs a commit message?

dawsontoth:
@kriszyp https://www.conventionalcommits.org/en/v1.0.0/ the subject would be when you have a first line that follows the general form of `(fix|feat|chore|docs|a few others): I did something great"

PR #677, opened 2025-09-11
feat: Make the status page a little prettier

dawsontoth:
I got a little fancy with the recursion, let me know if that doesn't seem sustainable, and we can rethink it.

The end result of this is that we'll render a flat list of key-value pairs, with a padding depth added to them, and reasonable value formatting.

On the left is a colocated cluster, and on the right is local mode.

image

https://harperdb.atlassian.net/browse/STUDIO-404


dawsontoth:
@BboyAkers I'm guessing our next incarnation of this page will be much more refined and designed, with like... widgets of each component, badges highlighting the important stuff... not this programmatic table of data. But this was an easy-ish solution for the time being.

BboyAkers (approved):

PR #698, opened 2025-09-16
fix: Consistently trim and lowercase usernames and emails

dawsontoth:
I extracted a few shared zod constants that we can use, and grow in the future.

The effective changes are as follows:
1. When specifying a username or email, we'll always trim it, and lowercase it, regardless of signing-up or signing-in, for both cloud and instances.
2. When signing-up, we'll also apply the length restrictions.
3. Sign in, reset password and forgot password won't apply the length restrictions (since they aren't creating a user).

Plus, my auto-formatter wrapped the new privacy statement to meet the column wrap lint.

https://harperdb.atlassian.net/browse/STUDIO-420


dawsontoth:
A counter argument could be made that we should do this only for cloud usernames, where we have complete control over them and the API enforces the same thing. But let the user do what they want with their instance credentials.

BboyAkers (approved):

PR #699, opened 2025-09-16
feat: Order logs desc by default

dawsontoth:
Now they'll show up in descending order, so the most recent logs will be at the top.

https://harperdb.atlassian.net/browse/STUDIO-407

### DIFF
--- a/src/features/instance/log/components/LogsFiltersForm.tsx
+++ b/src/features/instance/log/components/LogsFiltersForm.tsx
@@ -16,7 +16,7 @@ export function LogsFiltersForm({
 	resetFilters,
 	submitFilters,
 }: {
-	form: UseFormReturn;
+	form: UseFormReturn<z.infer<typeof LogFiltersFormSchema>>;
 	resetFilters: () => void;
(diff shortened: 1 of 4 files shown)


kriszyp:
Cool. I had talked to @BboyAkers about this: the `order` property should _never_ be "asc", it should always be "desc". However, showing the logs with newest at the top or the bottom should be configurable. Honestly, I think 50% of log viewers show newest at top and 50% show newest at the bottom. This configuration should just be how you display the returned array (reversed or as returned), it should not be part of the filters and should not affect the `order` property. And "Log Order" should probably be removed as a drop from the filters section. It is not a filter, and shouldn't affect the requested data. Display ordering should probably be an icon outside of the filters, maybe at the top.

BboyAkers (approved):

dawsontoth:
@kriszyp I removed the sort option, making it always `desc`. I created a ticket https://harperdb.atlassian.net/browse/STUDIO-434 for us to tackle in the near future logs-from-the-bottom-of-the-screen.

PR #712, opened 2025-09-17
feat: Hide terminated clusters

dawsontoth:
At some point in the future, we can give users a way to see these terminated clusters if we come up with useful functionality for them (like viewing past invoices or something). Until then… let’s just hide them completely.

https://harperdb.atlassian.net/browse/STUDIO-340


Devin-Holland (approved):

Devin-Holland:
Thanks!

kriszyp (approved):
Nice

PR #713, opened 2025-09-17
feat: UI Tweaks

dawsontoth:
## Set Password
If you've got a new cluster in `Set Password` mode, and you try to circumvent it by going to the sign-in screen for one of its instances, you'll get properly redirected to the `Set Password` page.

## Status Nesting
We won't repeat the array name on each element anymore. Instead, we'll put it up at the top once, then count them off one by one. Note that arrays of size 1 will only show the array name, without any number.
image

## Breadcrumb Tweaks
I removed the ids from breadcrumbs, added instance names, and made the cluster breadcrumb link to the instances page _only_ when you've navigated to an instance within the cluster.
image

## Sidebar Folders
We'll track the depth of our folders for applications, and auto-collapse everything deeper than 1.
image

https://harperdb.atlassian.net/browse/STUDIO-435

### DIFF
--- a/src/components/Breadcrumbs.tsx
+++ b/src/components/Breadcrumbs.tsx
@@ -23,22 +23,25 @@ export function Breadcrumbs() {
 				continue;
 			}
 
-			const path = `/${routeHistory.slice(0, index + 1).join('/')}`;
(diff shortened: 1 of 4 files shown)


kriszyp (approved):
Amazing work. This is awesome.

PR #719, opened 2025-09-17
feat: Make packages read-only and re-deployable

dawsontoth:
1. I broke out the components from the File Tree Explorer so that things are a bit easier to find.
2. Then I made all the icons have the same width so that they'll visually line up.

Writable folders and files have their same actions available:
editable-file
editable-folder

But now packages have two new icons: a package 😆 and a lock!
collapsed

If you hover over the box (or an expandable folder or application icon) you'll notice we have a new pointer, which should differentiate from clicking on the name of the thing! Click to open the package.
expanded

Now when looking at the files within the package, they're all read-only. You'll have two options available: delete or redeploy the whole component. If you have a file within it selected, and pick "delete" for example, it will issue the proper request.
re-deploy-or-delete

https://harperdb.atlassian.net/browse/STUDIO-386
https://harperdb.atlassian.net/browse/STUDIO-425


src/features/instance/applications/new/ImportProjectForm.tsx
```diff
@@ -53,23 +51,23 @@ export function ImportProjectForm({
 	const submitForm = async (formData: DeployComponentFormData) => {
 		deployNewApplication({ ...formData, ...instanceParams }, {
 			onSuccess: () => {
-				toast.success(`Application ${formData.newApplicationName} created successfully`);
-				triggerRestart();
+				toast.success(`Application ${formData.applicationName} imported successfully`);
+				onRestartedSuccessfully();
```
dawsontoth, line 55: Deploying the new app does a rolling restart already, so all we need to do is trigger the cache invalidation and toast!

src/features/instance/operations/mutations/createComponent.ts
```diff
@@ -2,18 +2,18 @@ import { InstanceClientConfig } from '@/config/instanceClientConfig';
...
 async function onCreateComponentSubmit({
-	newApplicationName,
+	applicationName,
 	instanceClient,
 	replicated,
 }: CreateComponentFormData & InstanceClientConfig) {
 	const { data } = await instanceClient.post('/', {
 		operation: 'add_component',
-		project: newApplicationName,
+		project: applicationName,
```
dawsontoth, line 16: I renamed this because we can use it to re-deploy an app too.

src/features/instance/operations/queries/getComponents.ts
```diff
@@ -30,7 +31,7 @@ type GetComponentsResponseWithMetaData = GetComponentsResponse & {
 	key?: string;
 	path?: string;
 	project?: string;
-	readOnly?: boolean;
+	pkg?: string;
```
dawsontoth, line 34: By tracking the package name, instead of coercing it to a bool, we can redeploy regardless of what file is selected within the package. I used "pkg" so we could destructure, and avoid overlap with the "package" name itself.

kriszyp (approved):
At some point, when we have better mechanisms to deploy without a restart, we may want to revisit if we only restart after a deploy if it is necessary or the user wants it, but I think this is the best behavior for now.
Anyway, this is fantastic, great work!

dawsontoth:
@BboyAkers I was able to stabilize the keys for the file tree, so now when we refresh that list, things don't collapse down. But then I added session storage as well so if you navigate away and back, you'll get to see the tree in the same state! And it will handle it across clusters and instances.

dawsontoth:
Sonar is just never happy, is it?

PR #724, opened 2025-09-18
Improved Free Validations

dawsontoth:
We'll do a few things better now:
1. The resources usage block will wrap properly now. When we moved it, we didn't update the classes.
2. If you select a free performance tier, we won't show the "add region" button.
3. If you add multiple regions on a paid tier, then select the free tier, it will warn you on the 2nd+ region.
4. If your org has already created a free cluster, we'll auto-pick a paid tier when opening the new cluster form.
5. If you try to pick the free tier in that org, we'll show a validation error on the performance tier dropdown.

https://harperdb.atlassian.net/browse/STUDIO-441

### DIFF
--- a/src/features/clusters/upsert/index.tsx
+++ b/src/features/clusters/upsert/index.tsx
@@ -2,7 +2,6 @@ import { ContactUs } from '@/components/ContactUs';
 import { ErrorComponent } from '@/components/ErrorComponent';
 import { Loading } from '@/components/Loading';
 import { SubNavSimpleLayout } from '@/components/SubNavSimpleLayout';
-import { getClusterInfoQueryOptions } from '@/features/cluster/queries/getClusterInfoQuery';
 import { getPlanTypesOptions } from '@/features/cluster/queries/getPlanTypesQuery';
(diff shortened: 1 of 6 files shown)


src/features/clusters/upsert/index.tsx
```diff
@@ -37,6 +35,23 @@ export function UpsertCluster() {
 	const { data: regionLocationsAll } = useQuery(getRegionLocationsOptions({ organizationId }));
 	const regionLocations = onlyShowAvailableHosts ? regionLocationsAvailableHosts : regionLocationsAll;
 
+	const alreadyUsingFree = useMemo(() => {
```
kriszyp, line 38: Would this be better computed on the backend, and returned in the cluster endpoint response? (I assume it has similar logic for validating the creation of a free instance).

kriszyp (approved):
Made a note about how we might want to check free instances in the future, but I think this is fine for now.
This looks great!

src/features/clusters/upsert/index.tsx
dawsontoth: Yeah, I think so! And it could be a computed property? I do worry about how much stuff we are hanging off of our top level entities. But we can request smaller parts of them with Harper, right?

PR #726, opened 2025-09-18
fix: Update language to indicate we are renewing

kriszyp:


dawsontoth (approved):
Much clear. Great words.

PR #729, opened 2025-09-19
Embed Swagger UI

dawsontoth:
I added a new `APIs` nav item to instances (and made that code a bit DRYer), then lazy loaded [Swagger UI](https://swagger.io/tools/swagger-ui/) once we check the docs are accessible. We'll show decent errors when something stops us from getting through.

- [x] The [core PR](https://github.com/HarperDB/harperdb/pull/2872) needs to merge.
- [x] A beta needs to get published and used by our instances with the changes.
- [x] If the user doesn't have CORS enabled for HTTP, or hasn't added the current origin to it, we need to warn them and give them an easy "plz enable" button.

image

I also added a custom snippet for Node.js Fetch, which, being a template, is a little hard to read, but I think you'll get the idea of it.

image

This pairs with some changes in core: https://github.com/HarperDB/harperdb/pull/2872

https://harperdb.atlassian.net/browse/STUDIO-344


src/features/instance/apis/plugins.ts
```diff
@@ -0,0 +1,52 @@
...
+\t\theaders: {
+\t\t\t${request.get('headers').map((val: string, key: string) => `"${key}": "${val}"`).valueSeq().join(',\n\t\t\t')}
+\t\t},` : ''}${reqBody ? `
+\t\tbody: JSON.stringify(${stringBody}),` : ''}
+\t});
+\tconst data = await response.json();
+\tconsole.log(data);
+}
+
+main().catch(console.error);
```
kriszyp: Can we drop the catch and the `main` function and just have the `fetch` call? (immediately catching and I believe consoling an error is neither recommended nor typically all that helpful in most contexts).

src/features/instance/apis/requestSnippets.ts
```diff
@@ -0,0 +1,8 @@
+export const requestSnippets = {
+	generators: {
+		node_fetch: {
+			title: 'Node.js Fetch',
```
kriszyp: Just `fetch`. There isn't any node.js specific about this, I don't think.

src/features/instance/InstanceNavBar.tsx
```diff
@@ -32,6 +32,11 @@ export function InstanceNavBar() {
 			shortName: 'Apps',
 			icon: <PackageIcon className="inline-block" />,
 		},
+		canManage && {
```
kriszyp: I think maybe we should put this after "Databases".

src/features/instance/apis/requestSnippets.ts
```diff
@@ -0,0 +1,8 @@
+export const requestSnippets = {
```
kriszyp, line 1: Does this still show the `curl` example(s) as well? Do we have control over the order? Can `fetch` come before `curl`?

kriszyp (commented):
This is such a powerful feature, great work.

src/features/instance/apis/APIDocs.tsx
```diff
@@ -0,0 +1,105 @@
...
+			showReturnToHome={false}
+		>
+			<Button disabled={isSettingConfiguration} variant="warning" className="rounded-full" onClick={enableCORS}>
+				<Plus />
+				{isSettingConfiguration ? 'Enabling CORS...' : `Enable CORS for ${window.location.origin}`}
+			</Button>
+		</ErrorComponent>)}
+		<SwaggerUI
+			spec={spec}
+			persistAuthorization={true}
```
kriszyp, line 98: Generally, authorization takes place with the pre-existing cookie, right?

dawsontoth: Not across domains, or across ports... right?

src/features/instance/apis/plugins.ts
dawsontoth: Done!

src/features/instance/apis/requestSnippets.ts
dawsontoth: Done.

src/features/instance/apis/requestSnippets.ts
dawsontoth: I think if we were to hard-code each of the examples, we could make fetch show up first. I'll save this in the back of my mind for when we rewrite this page (soon)!

BboyAkers (approved):

PR #743, opened 2025-09-22
Updated Cluster Card and Cluster Card Dropdown

BboyAkers:
Ticket:
[STUDIO-438](https://harperdb.atlassian.net/browse/STUDIO-438)

Overview:
- Replaced cluster id for fqdn url on cluster card with copy icon
- Added copy api url in cluster card dropdown for non-self-managed clusters
- Display "Self-Hosted" if the cluster is self-hosted in replacement of the fqdn url
- Removed copy fqdn url on cluster card dropdown if cluster is self-hosted

Screenshot 2025-09-22 at 2 17 23 PM

Screenshot 2025-09-22 at 2 19 38 PM

Screenshot 2025-09-22 at 2 19 49 PM

[STUDIO-438]: https://harperdb.atlassian.net/browse/STUDIO-438?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/features/clusters/components/ClusterCard.tsx
+++ b/src/features/clusters/components/ClusterCard.tsx
@@ -116,7 +116,8 @@ export function ClusterCard({
 					{cluster.fqdn ? (
 						<>
 							<span className="truncate max-w-48">{cluster.fqdn}</span>
-							<CopyIcon onClick={onCopyFQDNClick} size={16} />
+							<CopyIcon onClick={onCopyFQDNClick} size={16} className="cursor-pointer" />
+							<span className="grow"></span>
 						</>
 					) : (
 						<span>Self-Hosted</span>
--- a/src/features/clusters/components/ClusterCard.tsx
+++ b/src/features/clusters/components/ClusterCard.tsx
@@ -37,6 +37,7 @@ export function ClusterCard({
 }) {
 	const { view, update, remove } = useOrganizationClusterPermissions(cluster.organizationId, cluster.id);
 	const auth = useInstanceAuth(cluster.id);
+	const isSelfManaged = useMemo(() => !!cluster?.plans?.[0]?.planId?.startsWith('self-hosted'), [cluster]);
 
 	const isActive = useMemo(() => cluster.status && activeClusterStatuses.includes(cluster.status), [cluster.status]);
 	const isTerminated = useMemo(
@@ -71,6 +72,11 @@ export function ClusterCard({
 		toast.info('FQDN url copied to clipboard');
 	}, [cluster.fqdn]);
 
+	const onCopyAPIClick = useCallback(() => {
+		navigator.clipboard.writeText(`https://${cluster.fqdn}`);
+		toast.info('API url copied to clipboard');
+	}, [cluster.fqdn]);
+
 	const menuItems = [
 		isActive && update && (
 			<Link to={`${cluster.id}/edit`} disabled={signingOut}>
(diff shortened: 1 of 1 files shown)


src/features/clusters/components/ClusterCard.tsx
```diff
@@ -82,11 +88,16 @@ export function ClusterCard({
 				<DropdownMenuItem>Instances</DropdownMenuItem>
 			</Link>
 		),
-		isActive && view && (
+		isActive && view && !isSelfManaged && (
```
dawsontoth: A self-hosted actually can have a FQDN, so let's tweak this to instead only show up if we have a `cluster.fqdn`! Make sense?

Image

src/features/clusters/components/ClusterCard.tsx
```diff
@@ -82,11 +88,16 @@ export function ClusterCard({
...
 			<DropdownMenuItem onClick={onCopyFQDNClick} disabled={signingOut}>
 				Copy FQDN Url
 			</DropdownMenuItem>
 		),
+		isActive && view && !isSelfManaged && (
```
dawsontoth: Ditto here -- show this if we've got a `cluster.fqdn`!

src/features/clusters/components/ClusterCard.tsx
```diff
@@ -103,7 +114,14 @@ export function ClusterCard({
 		<Card className="relative h-full justify-between">
 			<CardHeader>
 				<CardDescription className="flex items-center justify-between">
-					<span className="truncate">{cluster.id}</span>
+					{!isSelfManaged ? (
```
dawsontoth: This could instead be `{cluster.fqdn ? (`, and then the `isSelfManaged won't be used anywhere and can be removed, I think.

src/features/clusters/components/ClusterCard.tsx
```diff
@@ -71,6 +72,11 @@ export function ClusterCard({
 		toast.info('FQDN url copied to clipboard');
```
dawsontoth:
```suggestion
		toast.info('FQDN copied to clipboard');
```

src/features/clusters/components/ClusterCard.tsx
BboyAkers: Ahhh gotcha!

src/features/clusters/components/ClusterCard.tsx
```diff
@@ -103,7 +113,14 @@ export function ClusterCard({
 		<Card className="relative h-full justify-between">
 			<CardHeader>
 				<CardDescription className="flex items-center justify-between">
-					<span className="truncate">{cluster.id}</span>
+					{cluster.fqdn ? (
+						<>
+							<span className="truncate max-w-48">{cluster.fqdn}</span>
+							<CopyIcon onClick={onCopyFQDNClick} size={16} />
```
dawsontoth: Can we have a pointer cursor on the icon? And then add a `grow` after it too so that the icon stays snapped to its best buddy! Particularly noticeable when on mobile sizes.
```suggestion
							<CopyIcon onClick={onCopyFQDNClick} size={16} className="cursor-pointer" />
							<span className="grow"></span>
```

dawsontoth (approved):

dawsontoth:
You'll want to re-write history a tiny bit to make the commitlint happy, too.

dawsontoth:
@BboyAkers can you re-write history again? Notice that it's happening when committing suggestions via GitHub. You can choose the commit message when accepting suggestions, just remember to add a relevant prefix.

PR #748, opened 2025-09-24
fix: Add runner.html for when Studio is off

dawsontoth:
When running Harper, if you don't enable Studio, core will try to render the runner.html file. We didn't have one. But now we do. 🥳 

I mostly copy-pasted this from old Studio, but I tweaked it for our rebranding.

### DIFF
--- a/public/running.html
+++ b/public/running.html
@@ -1,0 +1,90 @@
+<!doctype html>
+<html lang="en">
+<head>
+	<title>Harper is Running!</title>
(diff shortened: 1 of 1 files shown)


cap10morgan:
Nice! Should we have marketing take a look at this?

dawsontoth:
@cap10morgan yeah, makes sense!

dawsontoth:
< conversation shifted to slack marketing channel >

BboyAkers (approved):

PR #753, opened 2025-09-25
fix: Handle abbreviated name under self-managed properly

dawsontoth:
There were inconsistencies based on what you’d type, and when you’d switch, that could prevent Zod from passing validation. Or we’d accidentally pass an empty string to the server, causing a validation error there. This will avoid those odd cases, and make sure we pass along the abbreviated name when appropriate.

### DIFF
--- a/src/features/clusters/upsert/index.tsx
+++ b/src/features/clusters/upsert/index.tsx
@@ -112,7 +112,7 @@ export function UpsertCluster() {
 		return {
 			autoRenew: cluster?.plans?.[0]?.autoRenew ?? true,
 			systemName: cluster?.name ?? '',
-			abbreviatedName: cluster?.abbreviatedName || '',
+			abbreviatedName: cluster?.abbreviatedName ?? '',
 			deploymentDescription: selectedPlan?.deploymentDescription ?? defaults?.deploymentDescription ?? '',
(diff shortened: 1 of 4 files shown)


src/features/clusters/upsert/index.tsx
```diff
@@ -112,7 +112,7 @@ export function UpsertCluster() {
 		return {
 			autoRenew: cluster?.plans?.[0]?.autoRenew ?? true,
 			systemName: cluster?.name ?? '',
-			abbreviatedName: cluster?.abbreviatedName ?? '',
+			abbreviatedName: cluster?.abbreviatedName || '',
```
cap10morgan: Might benefit from a comment as to why this needs `||` when the other similar lines around it all use `??`?

cap10morgan (approved):
One small suggestion but looks good overall

src/features/clusters/upsert/index.tsx
dawsontoth: Actually, with the solution I came up with around this, we can go back to the nullish coalescing operator!

cap10morgan: > Actually, with the solution I came up with around this, we can go back to the nullish coalescing operator!

Even better! 👍🏻

BboyAkers (approved):

PR #782, opened 2025-10-02
fix: Tweak copy dropdown item names

dawsontoth:
I think the simple change could be:
"Copy FQDN Url" -> "Copy host name" (its not a URL, and I would think host name is friendlier)
"Copy API Url" -> "Copy API URL" (consistent casing)


kriszyp (approved):
Thank you!

PR #783, opened 2025-10-02
Added redeploy package reference modal

BboyAkers:
Overview:
- Added redeploy package reference modal

Screenshot 2025-10-02 at 7 11 03 PM

### DIFF
--- a/src/features/instance/applications/modals/RedeployApplicationModal.tsx
+++ b/src/features/instance/applications/modals/RedeployApplicationModal.tsx
@@ -15,36 +15,41 @@ export function RedeployApplicationModal({
 	isModalOpen = false,
 	setIsModalOpen,
 	redeployPackage,
-  packageUrl,
+	isRedeployPackagePending,
+	packageUrl,
 }: {
 	readonly isModalOpen: boolean;
 	readonly setIsModalOpen: (value: boolean) => void;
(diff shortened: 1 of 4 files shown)


src/features/instance/applications/modals/RedeployApplicationModal.tsx
```diff
@@ -0,0 +1,109 @@
...
+								control={control}
+								name="applicationUrl"
+								render={({ field }) => (
+									<FormItem>
+										<FormLabel className="pb-1">Package Reference</FormLabel>
+										<FormControl>
+											<Input
+												type="text"
+												defaultValue={packageUrl}
+												placeholder="https://github.com/HarperDB/nextjs-example"
```
dawsontoth:
```suggestion
												placeholder="https://github.com/HarperFast/status-check"
```

dawsontoth (approved):

src/features/instance/applications/modals/RedeployApplicationModal.tsx
kriszyp: Could we use this as the default or placeholder for the import dialog too? nextjs-example is very slow to install, and is actually a template, not a "finished" app. This dialog will almost always be filled in, so the place holder is actually fairly rarely seen, right?

PR #786, opened 2025-10-03
Feature: basic analytics visualizations

cap10morgan:
Screenshot 2025-10-03 at 8 49 18 AM

Implements CORE-2767

I still need to add a few more metrics (once I confirm I can access them via `get_analytics`, but the basic structure is ready. This was meant as a research spike, so feel free to take it, leave it, or anything in between!

### DIFF
--- a/src/features/instance/operations/queries/getAnalytics.ts
+++ b/src/features/instance/operations/queries/getAnalytics.ts
@@ -1,0 +1,46 @@
+import { InstanceClientConfig } from '@/config/instanceClientConfig.ts';
+import { queryOptions } from '@tanstack/react-query';
+
+export interface MetricConfig {
+	id: string;
+	name: string;
+	unit: string;
+	path?: string;
+}
+
+interface GetAnalyticsParams extends InstanceClientConfig {
+	metricConfig: MetricConfig;
+	startTime: number;
+	endTime: number;
+}
+
+interface GetAnalyticsRequest {
+	operation: 'get_analytics';
+	metric: string;
+	start_time: number;
+	end_time: number;
+	conditions?: {
+		attribute: string;
+		value: string|number|boolean;
+		comparator?: string;
+	}[];
+}
+
+export function getAnalyticsQueryOptions({ metricConfig, startTime, endTime, instanceClient }: GetAnalyticsParams) {
+	return queryOptions({
+		queryKey: ['get_analytics', metricConfig.name, metricConfig.path, startTime, endTime] as const,
+		queryFn: async () => {
+			const req: GetAnalyticsRequest = {
+				operation: 'get_analytics',
+				metric: metricConfig.name,
+				start_time: startTime,
+				end_time: endTime,
+			}
+			if (metricConfig.path) {
+				req.conditions = [{ attribute: 'path', value: metricConfig.path }];
+			}
+			const { data } = await instanceClient.post('/', req);
+			return data;
+		}
+	});
+}
--- a/src/features/instance/operations/queries/getAnalytics.ts
+++ b/src/features/instance/operations/queries/getAnalytics.ts
@@ -1,4 +1,4 @@
-import { InstanceClientConfig } from '@/config/instanceClientConfig.ts';
+import type { InstanceClientIdConfig, InstanceTypeConfig } from '@/config/instanceClientConfig.ts';
 import { queryOptions } from '@tanstack/react-query';
 
 export interface MetricConfig {
@@ -8,10 +8,11 @@ export interface MetricConfig {
 	path?: string;
 }
 
-interface GetAnalyticsParams extends InstanceClientConfig {
+interface GetAnalyticsParams {
 	metricConfig: MetricConfig;
 	startTime: number;
 	endTime: number;
+	instanceParams: InstanceClientIdConfig & InstanceTypeConfig;
 }
 
 interface GetAnalyticsRequest {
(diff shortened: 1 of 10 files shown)


src/features/instance/operations/queries/getAnalytics.ts
```diff
@@ -0,0 +1,46 @@
...
+				operation: 'get_analytics',
+				metric: metricConfig.name,
+				start_time: startTime,
+				end_time: endTime,
+			}
+			if (metricConfig.path) {
+				req.conditions = [{ attribute: 'path', value: metricConfig.path }];
+			}
+			const { data } = await instanceClient.post('/', req);
+			return data;
```
dawsontoth, line 50: Can we type the response?

cap10morgan:
It's hard to see in the screenshot, but it also adds a new line for each node in a cluster. The new lines follow the Harper color palette (which may need some tweaking contrast-wise against this background color).

src/features/instance/operations/queries/getAnalytics.ts
cap10morgan: > Can we type the response?

Not easily. There's a _lot_ of variation, and with custom metrics it's pretty open-ended in the general case. I could put a very loose type on it that reflects that, though.

src/features/instance/status/index.tsx
```diff
@@ -3,5 +3,5 @@ import { CloudStatus } from '@/features/instance/status/CloudStatus';
 import { LocalStatus } from '@/features/instance/status/LocalStatus';
 
 export function StatusIndex() {
-	return isLocalStudio ? (<LocalStatus />) : (<CloudStatus />);
+	return isLocalStudio ? <LocalStatus /> : <CloudStatus />;
 }
```
dawsontoth, line 20: We'll probably want to shift this to use a lazy loaded component so that the charting library only comes in when you visit the status page, not all the time. The profile page has an example of this.
```
const profileRoute = createRoute({
	getParentRoute: () => dashboardLayout,
	path: 'profile',
	component: lazyRouteComponent(async () => import('@/features/profile/index'), 'ProfileIndex'),
});

export const profileRoutes = [profileRoute];
```
You could also shift the common components you added to Local and Cloud Status into this index (essentially reverting them to what they were before, with all your changes up here instead).

dawsontoth:
Thoughts on ways we could handle the empty data state better? I'm guessing overlaying a bit of text to explain why it's empty would help. "No reads for this time period!"
Screenshot 2025-10-03 at 11 32 41 AM

src/features/instance/status/components/Monitoring.tsx
```diff
@@ -0,0 +1,108 @@
...
+					{/*  onChange={(e) => setTimeWindowInputValue(Number(e.target.value))}*/}
+					{/*  onBlur={(e) => setTimeWindow(Number(e.target.value) * 60_000)}*/}
+					{/*  value={timeWindowInputValue}*/}
+					{/*/>*/}
+					{/*minutes*/}
+					<Label className="ml-8 mr-2">Update every</Label>
+					<Input
+						className="mr-2 w-16 inline-block"
+					  id="updateIntervalInput"
+					  type="text"
```
dawsontoth: I recommend type="number" here. For fun, though, see what happens when you type `20a` and get NaN, particularly watch your network requests.

dawsontoth: Alternatively, a dropdown would force the user to pick from some sane values instead. Less can go wrong then.

src/features/instance/operations/queries/getAnalytics.ts
```diff
@@ -0,0 +1,52 @@
...
+			const req: GetAnalyticsRequest = {
+				operation: 'get_analytics',
+				metric: metricConfig.name,
+				start_time: startTime,
+				end_time: endTime,
+			}
+			if (metricConfig.path) {
+				req.conditions = [{ attribute: 'path', value: metricConfig.path }];
+			}
+			const { data } = await instanceClient.post<GetAnalyticsResponse>('/', req);
```
dawsontoth: Array?

cap10morgan: > Array?

Haha yup!

dawsontoth:
The units on the CPU graph are a bit perplexing to me. Is that showing user time? What does it cap at? If we can convert that to show a percentage, and force that graph to clamp to 0-100% (or whatever our max is) that would help with readability.

cap10morgan:
> The units on the CPU graph are a bit perplexing to me. Is that showing user time? What does it cap at? If we can convert that to show a percentage, and force that graph to clamp to 0-100% (or whatever our max is) that would help with readability.

Yeah I'm not super clear on it either. This is where it gets calculated / set: https://github.com/HarperFast/harperdb/blob/007ccfe314624832be5ff322c212ac52ea5ff7ed/resources/analytics/profile.ts#L37-L98 I think ultimately we want it to be clear how it correlates to what we measure for billing purposes. Maybe @kriszyp has some more insight?

cap10morgan:
> Thoughts on ways we could handle the empty data state better? I'm guessing overlaying a bit of text to explain why it's empty would help. "No reads for this time period!" Screenshot 2025-10-03 at 11 32 41 AM

Something like this?

Screenshot 2025-10-03 at 10 39 46 AM

cap10morgan:
OK @dawsontoth I think that's everything you pointed out. Thanks for the thorough review!

src/features/instance/status/components/Monitoring.tsx
cap10morgan: > Alternatively, a dropdown would force the user to pick from some sane values instead. Less can go wrong then.

Yeah, that's better. Done in https://github.com/HarperFast/hdbms/pull/786/commits/53e7f82c0f67e0858ef7e3d577a65b7485fe093e

dawsontoth (approved):
I like it!

BboyAkers (approved):

dawsontoth:
@cap10morgan this can merge whenever you're ready for it to merge

dawsontoth (approved):

PR #806, opened 2025-10-09
chore: Add .editorconfig and `VITE_HOST` example

dawsontoth:
This gives us a few things:
1. Tabs in all our JSON files (I did a quick reformat on them) that our editors will respect.
2. Consistent final newlines
3. We can `VITE_HOST=localhost` or whatever we want like `VITE_HOST=harper.localhost` (with changes in your `/etc/hosts` to map this too).
4. I split the release config out to its own file to make the package.json a little shorter.


BboyAkers (approved):
🔥 🔥 🔥 🔥

PR #817, opened 2025-10-10
feat: Add additional metrics & units scaling

cap10morgan:
Completes STUDIO-96

This completes the initial builtin Fabric analytics visualizations work.

### DIFF
--- a/src/features/instance/status/components/monitoring/MetricVisualization.tsx
+++ b/src/features/instance/status/components/monitoring/MetricVisualization.tsx
@@ -3,6 +3,7 @@ import { getAnalyticsQueryOptions, type MetricConfig, type Metric, type MetricDa
 import type { InstanceClientIdConfig, InstanceTypeConfig } from '@/config/instanceClientConfig.ts';
 import { useMemo, useState } from 'react';
 import { Legend, Line, LineChart, ResponsiveContainer, Tooltip, XAxis, YAxis } from 'recharts';
+import { harperPalette } from '@/lib/colorPalette.ts';
 
 type MetricValue = string | number | boolean;
 type NullableMetricValue = MetricValue | null;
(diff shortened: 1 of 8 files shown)


src/features/instance/status/components/monitoring/MetricVisualization.tsx
```diff
@@ -28,11 +26,72 @@ const harperPalette = {
...
+			availableUnits = timeUnits;
+			break;
+		default:
+			return baseUnits;
+	}
+	let i = 0;
+	while (value > 1024 && i < availableUnits.length - 1) {
+		value /= 1024;
+		i++;
+	}
```
dawsontoth: Can you add unit tests for these? And split them into separate files? This appears to be, in some cases, dividing time by 1024 to shift between the units of secs, mins, hrs?...

cap10morgan: Uhh, yup! I meant to circle back to this and make it work with time too haha

dawsontoth: Please take a look at src/lib/humanFileSize.ts and src/lib/translateSecondsToAgo.ts. Depending, we may instead want to use src/lib/humanNumber.ts -- which will yield bigger numbers, that may be desired. Not sure if 1024 or 1000 is the designated base.

cap10morgan:
@dawsontoth Thanks for catching that and the duplicated logic pointer. I need something slightly different in the analytics visualization code, so I reworked `humanFileSize` to use my code and reorganized things a bit. And added and updated tests! And maybe fixed a small bug (where byte values less than 1 KiB weren't being formatted)? I also changed all of the file size units to IEC since it was already using 1024 instead of 1000.

src/lib/units.ts
```diff
@@ -0,0 +1,56 @@
+export type Units = 'bytes' | 'secs';
+
+const dataUnits = ['B', 'KiB', 'MiB', 'GiB', 'TiB', 'PiB', 'EiB', 'ZiB', 'YiB'];
+const timeUnits = ['secs', 'mins', 'hrs'];
+const unitsMultipliers = {
+	bytes: 1024,
```
kriszyp: Our pricing is all based on decimal (K/M/G) conversions, not binary (Ki/Mi/Gi). I would think we would want to follow that with our conversions? Or does this align us more Grafana and other tools?

kriszyp (commented):
Our analytics is based on decimal/SI/1000, rather than binary/EIC/1024 conversions, maybe we want to use that?

src/lib/units.ts
cap10morgan: Nope, using the decimal form would be fine. I'll switch it to that.

kriszyp (approved):

dawsontoth (approved):

PR #820, opened 2025-10-13
Beta in header

dawsontoth:
You can test this in your .env.local really easily:

```
VITE_STUDIO_VERSION="v2.45.2"
```

image

### DIFF
--- a/.github/workflows/deploy-dev-with-restart.yaml
+++ b/.github/workflows/deploy-dev-with-restart.yaml
@@ -31,7 +31,10 @@ jobs:
         run: pnpm lint
       - name: Determine version tag for build
         run: |
-          VERSION_TAG="SHA-${GITHUB_SHA:0:7}"
+          VERSION_TAG=$(git tag --points-at HEAD | grep -E '^v?[0-9]+\.[0-9]+\.[0-9]+' | sort -V | tail -n1)
+          if [ -z "$VERSION_TAG" ]; then
+            VERSION_TAG="dev_${GITHUB_SHA:0:7}"
+          fi
           echo "VITE_STUDIO_VERSION=$VERSION_TAG" >> $GITHUB_ENV
       - name: Build dev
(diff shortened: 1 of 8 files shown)


.github/workflows/deploy-dev-with-restart.yaml
```diff
@@ -31,7 +31,10 @@ jobs:
         run: pnpm lint
       - name: Determine version tag for build
         run: |
-          VERSION_TAG="SHA-${GITHUB_SHA:0:7}"
+          VERSION_TAG=$(git tag --points-at HEAD | grep -E '^v?[0-9]+\.[0-9]+\.[0-9]+' | sort -V | tail -n1)
+          if [ -z "$VERSION_TAG" ]; then
+            VERSION_TAG="dev_${GITHUB_SHA:0:7}"
+          fi
```
dawsontoth, line 37: I decided that if we have a nice version tag, we should use it in dev too. It means the same thing as a commit SHA, but it's easier to read.

.github/workflows/deploy-prod-with-restart.yaml
```diff
@@ -33,7 +33,7 @@ jobs:
         run: |
           VERSION_TAG=$(git tag --points-at HEAD | grep -E '^v?[0-9]+\.[0-9]+\.[0-9]+' | sort -V | tail -n1)
           if [ -z "$VERSION_TAG" ]; then
-            VERSION_TAG="v0.0.0+${GITHUB_SHA:0:7}"
+            VERSION_TAG="prod_${GITHUB_SHA:0:7}"
```
dawsontoth, line 36: DataDog accepts these as versions too. So if for some reason we don't have a v1.2.3, we should use this `{image}_{sha_0-7}` format instead.

kriszyp (approved):

Devin-Holland (approved):

BboyAkers (approved):

PR #827, opened 2025-10-16
fix: Don’t double stringify the body you nut

dawsontoth:
I tried too hard to make fetch happen.


cap10morgan (approved):
It's not going to happen!

dawsontoth:
@cap10morgan 😆

PR #835, opened 2025-10-21
feat: Replicate set component file

dawsontoth:
Now when you add or update files, it will replicate across the cluster properly.

https://harperdb.atlassian.net/browse/STUDIO-523


kriszyp:
Thank you!

PR #865, opened 2025-10-28
fix: ensure SDK ready before startView and add login_success for retention tracking

baileydunning:
Made a few small tweaks to fix missing retention data in Datadog. Wrapped our `startView()` calls in `datadogRum.onReady()` so the SDK doesn’t drop the first view before it’s initialized, and added a `login_success` action after sign-in (right after `setUser()`) so Datadog has a clear cohort anchor for retention. Engagement metrics like session duration were already working. This just makes sure retention and post-login tracking show up correctly in Product Analytics.

### DIFF
--- a/src/features/auth/SignIn.tsx
+++ b/src/features/auth/SignIn.tsx
@@ -20,7 +20,7 @@ import { Link, useNavigate, useRouter, useSearch } from '@tanstack/react-router'
 import { useEffect } from 'react';
 import { useForm } from 'react-hook-form';
 import { z } from 'zod';
-import { datadogRum } from '@datadog/browser-rum';
+import { loginSuccessDatadogAction } from '@/integrations/datadog/datadog';
 
 const SignInSchema = z.object({
 	email: zodRequireEmail,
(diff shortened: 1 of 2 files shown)


src/features/auth/SignIn.tsx
```diff
@@ -56,6 +57,14 @@ export function SignIn() {
 				authStore.setUserForEntity(OverallAppSignIn, data);
 				const defaultCloudRoute = getDefaultSignedInCloudRouteForUser(data);
 
+				datadogRum.setUser({
+					id: data.id,
+					email: data.email,
+					name: [data.firstname, data.lastname].filter(Boolean).join(' ') || undefined,
+				});
+				
+				datadogRum.addAction('login_success', { userId: data.id });
```
dawsontoth: Everything RUM related needs to be inside the Datadog integration, and guarded by if it's `enabled` or not. Local Studio won't have RUM, so this will crash the user when they try to login, as-is.

dawsontoth (changes requested):
The fixes look good, aside from the unguarded calls in the login. Thanks for contributing the fix!

src/features/auth/SignIn.tsx
baileydunning: thank you for calling this out!

dawsontoth:
sending to dev

dawsontoth (approved):
Verified in dev. Locally verified in local-studio mode (that is, no changes or breaks in that mode).

Image

PR #873, opened 2025-11-03
fix: Contact us through Discord

dawsontoth:
https://harperdb.atlassian.net/browse/STUDIO-573

### DIFF
--- a/src/components/ContactUs.test.tsx
+++ b/src/components/ContactUs.test.tsx
@@ -1,0 +1,38 @@
+/**
+ * @vitest-environment jsdom
+ */
+import { TestProvider } from '@/lib/test/TestProvider';
+import { cleanup, render, screen } from '@testing-library/react';
+import { act } from 'react';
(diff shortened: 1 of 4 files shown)


kriszyp:
Contact us through discord about billing information? That doesn't seem like the right communication medium for billing questions, does it? (or at least not the only medium to suggest?)

dawsontoth:
@kriszyp yeah that's true... unless it's getting in touch with us personally on discord-as-a-platform? Where should we send people for billing Qs?

kriszyp:
Do we not have just a normal "Contact Us" page with the standard things like an email address, link to support, link to discord?

Ethan-Arrowood:
I think support@harperdb.io may be sufficient? We have a number of emails for folks to use such as opensource@harperdb.io depending on the circumstances

dawsontoth:
I put in that email address in a few spots, I think that'll work nicely. Good call, @Ethan-Arrowood!

BboyAkers (approved):

PR #877, opened 2025-11-05
Fix analytics aggregation

cap10morgan:
Fixes STUDIO-542

Prior to this change I wasn’t aggregating analytics across all tables in the results. So these charts would only show metrics for the last table in the get_analytics response.

Now they sum across all returned metrics for the same node and coalesced time window.


dawsontoth (approved):

dawsontoth:
I'll merge in a bit, once we get what's in stage approved and released.

PR #901, opened 2025-11-11
Monitoring chart stability follow-up

cap10morgan:
More fixes for STUDIO-550


cap10morgan:
This doesn't have the correct set of commits. Will redo shortly.

PR #902, opened 2025-11-11
fix: Reserve space for monitoring chart w/ no data

cap10morgan:
...above the sys info stuff. Follow up to STUDIO-550


dawsontoth (approved):
Thanks! I like it. Part of this is de-emphasizing the text-based status, too.

PR #920, opened 2025-11-14
[STUDIO-536] Download Applications

dawsontoth:
https://github.com/user-attachments/assets/a077f152-e9ab-4761-ac56-bbf2e9c72c20

https://harperdb.atlassian.net/browse/STUDIO-536

### DIFF
--- a/src/features/instance/applications/components/ContentActions.tsx
+++ b/src/features/instance/applications/components/ContentActions.tsx
@@ -9,6 +9,7 @@ import { useInstanceBrowseManagePermission } from '@/hooks/usePermissions';
 import { useEmitToListeners } from '@/lib/events/listener';
 import { useSetWatchedValue } from '@/lib/events/watcher';
 import {
+	DownloadIcon,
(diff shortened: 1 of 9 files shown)


kriszyp (commented):
This is awesome. Could we make the option be "Include node_modules" (default to off, which is skip_node_modules=true)? And is "Skip Symlinks" a meaningful option to include? Do users know what this means (we don't document it, I don't think)?

dawsontoth:
@kriszyp we do skip by default! And I don't know what the other option does either :)

kriszyp:
Right, the default is correct, I was more just suggesting presenting this as a positive option ("Include") instead of a negative option ("Skip").

dawsontoth:
Because we serialize to base64, and then decode that in memory on the client side... I'm not sure what performance will be like with a bunch of node modules too. It'd be nice if we could have the server stream the file out directly (with some parameter toggling that behavior) :thinking:

PR #929, opened 2025-11-18
fix: Refactor user tracking actions in Datadog integration

baileydunning:
- Switched first/return tracking to fire on login, not on random user state changes.
- Added first_time_login and return_login events using per-user localStorage keys.
- Made login_time_gap a true numeric measure by sending the value via @event.value.
- Made all login-related actions fire after setUser, ensuring correct attribution.

### DIFF
--- a/src/integrations/datadog/datadog.ts
+++ b/src/integrations/datadog/datadog.ts
@@ -77,47 +77,50 @@ export function useOnRouteLoadTracker() {
 			return;
 		}
 		if (user && !isLocalUser(user)) {
-	       datadogRum.setUser({
-		       id: user.id,
-		       name: [user.firstname, user.lastname].filter(excludeFalsy).join(' ') || undefined,
-		       email: user.email,
-	       });
-
-		   // First time vs return visit action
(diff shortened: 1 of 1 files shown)


src/integrations/datadog/datadog.ts
```diff
@@ -77,47 +77,50 @@ export function useOnRouteLoadTracker() {
...
+	const lastLoginKey = `lastLoginTimestamp:${data.id}`;
+	const lastLogin = localStorage.getItem(lastLoginKey);
 	const now = Date.now();
-	
+
 	if (lastLogin) {
 		const seconds = Math.floor((now - Number(lastLogin)) / 1000);
 		if (!Number.isNaN(seconds) && seconds >= 0) {
-			datadogRum.addAction('login_time_gap', { seconds });
+			datadogRum.addAction('login_time_gap', { value: seconds });
```
baileydunning, line 119: this is so datadog recognizes the value as a measure instead of a facet so we can easily calculate averages and such

dawsontoth (approved):

PR #930, opened 2025-11-19
fix: Show a warning when skipping adding records

dawsontoth:
If skipped hashes are described by the API, we can surface that in the toast and the dialog to the user so they can fix whatever went wrong.

https://github.com/user-attachments/assets/7609185f-1196-40cb-9704-14887abf6709

https://harperdb.atlassian.net/browse/STUDIO-599


kriszyp:
Excellent, thank you!

PR #1007, opened 2025-12-16
feat: Allow exporting csvs for one table at a time

dawsontoth:
image

https://harperdb.atlassian.net/browse/STUDIO-635


dawsontoth:
@kriszyp @Devin-Holland Do you want me to put some safeguards in place to protect running this under Fabric Connect? And if so, what should they be?

PR #1026, opened 2026-01-07
Update validation regex to only support lowercase

Devin-Holland:
To go with https://github.com/HarperFast/central-manager/pull/217


dawsontoth (approved):
Do we have existing data with uppercase characters?

Devin-Holland:
> Do we have existing data with uppercase characters?

Yes there's 3, but it's not possible to deploy clusters with the uppercase so it shouldn't matter. I can manually modify their subdomain in their record, but the names are very unlikely to ever be chosen again.

PR #1060, opened 2026-01-21
fix: Return a basic auth client if we flag for basic auth

dawsontoth:
When we were resetting the user's initial password, we were detecting we could use basic auth. But we weren't returning an updated instance client, so the old client (with cookie auth) was being used for the next request. That would error on Safari, of course. Now we'll properly utilize the basic auth for subsequent requests.

Unfortunately, we won't see the fruit of this until we get cross-origin requests, which only happens in Safari and Production.

https://harperdb.atlassian.net/browse/STUDIO-647

### DIFF
--- a/src/integrations/api/instance/auth/useInstanceLoginMutation.ts
+++ b/src/integrations/api/instance/auth/useInstanceLoginMutation.ts
@@ -44,17 +44,19 @@ export async function onInstanceLoginSubmit({
 	}
 
 	try {
+		const auth = {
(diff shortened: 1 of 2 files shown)


Devin-Holland (approved):

dawsontoth:
I pushed this to dev and verified that I can still set the initial user password with the happy path (cookie based auth) so this won't be _worse_ than what we have, at least. 😄

PR #1146, opened 2026-03-18
feat(ContentViewer): add Mermaid diagram support and enhance Markdown…

dawsontoth:
… rendering


dawsontoth:
Deployed to dev. You won't be able to stand up a colocated cluster there, but you can create a self-managed cluster and aim it at your instances in stage, for example.

PR #1152, opened 2026-03-23
feat: Allow cleaning up orphan blobs from databases on instances

dawsontoth:


kriszyp (changes requested):
Shouldn't we put this in the admin interface? It seems weird to expect users to fix our bugs.

dawsontoth:
@kriszyp do you mean we should only show this to users when they're harper admins?

dawsontoth:
@kriszyp I restricted it.

kriszyp (commented):
And you think it is appropriate that it is part of this UI versus the admin interface? (that will be a different interface, right?) I guess I was expecting that this would be part of our administrative UI. If this is the appropriate place, I'm fine with this, but wanted to make sure this in the right workflow.

Devin-Holland:
> And you think it is appropriate that it is part of this UI versus the admin interface? (that will be a different interface, right?) I guess I was expecting that this would be part of our administrative UI. If this is the appropriate place, I'm fine with this, but wanted to make sure this in the right workflow.

I think we should be able to jump from the Admin UI to instances/clusters/orgs (since we can access them all anyway) and then click the button for the specific instance. But maybe we can also have a shortcut to this in the admin panel as well?

dawsontoth:
I've been operating under the assumption that various administrative tasks can happen within this interface here. I wasn't planning on creating a separate codebase or website that you navigate to, similar to how we aren't making another copy of central manager to support the administrative functions.

kriszyp (approved):
Reviewing https://harperdb.atlassian.net/wiki/spaces/SD/pages/490209295/Admin+Page+s, I was assuming that most of this would exist under a different high-level category (sibling to "Organizations" tab?). But there isn't really anything in there that gets to the database level of instances. So I think this probably is the best place for this.

PR #1154, opened 2026-03-23
feat: Show /status on instances table

dawsontoth:
We'll use the `get_status` and `set_status` operations to read and write the `id: 'availability'` to `status: 'Unavailable'` or `status: 'Available'` for individual instances.

https://harperdb.atlassian.net/browse/STUDIO-669

### DIFF
--- a/src/features/cluster/InstanceStatusCell.tsx
+++ b/src/features/cluster/InstanceStatusCell.tsx
@@ -1,0 +1,100 @@
+import { Badge } from '@/components/ui/badge';
+import { Button } from '@/components/ui/button';
+import { Tooltip, TooltipContent, TooltipTrigger } from '@/components/ui/tooltip';
+import { useInstanceClientIdParams } from '@/config/useInstanceClient';
+import { useOrganizationClusterInstancePermissions } from '@/hooks/usePermissions';
+import { Instance } from '@/integrations/api/api.patch';
+import { getStatusQueryOptions, getSystemStatusById } from '@/integrations/api/instance/status/getStatus';
+import { useSetStatus } from '@/integrations/api/instance/status/setStatus';
+import { getOperationsUrlForInstance } from '@/lib/urls/getOperationsUrlForInstance';
+import { useQuery } from '@tanstack/react-query';
+import { LoaderCircleIcon, ShieldCheckIcon, ShieldXIcon } from 'lucide-react';
+import { useEffect, useMemo, useState } from 'react';
+
+export function InstanceStatusCell(
+	{ instance }: { readonly instance: Instance },
+) {
+	const operationsUrl = useMemo(() => getOperationsUrlForInstance(instance), [instance]);
+	const instanceParams = useInstanceClientIdParams({ operationsUrl, instanceId: instance.id });
+	const { update: canManage } = useOrganizationClusterInstancePermissions();
+	const { mutate: setStatus, isPending: isSettingStatus } = useSetStatus();
+
+	// We want to spread the initial requests across 5 seconds.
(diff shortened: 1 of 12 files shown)


kriszyp (changes requested):
This should be handled with `get_status` (which I thought was already displayed on the status page), and `set_status`. There shouldn't be any application port or CORS interaction involved.

src/features/cluster/InstanceStatusCell.tsx
```diff
@@ -0,0 +1,107 @@
...
+		return () => clearTimeout(timer);
+	}, [randomOffset]);
+
+	const { data: statusResponse, isLoading, isFetching } = useQuery(getStatusQueryOptions(instanceParams, ready));
+
+	const isAvailable = statusResponse?.systemStatus?.[0]?.status === 'Available';
+	const isUnavailable = statusResponse?.systemStatus?.[0]?.status === 'Unavailable';
+	const statusMessage = typeof statusResponse?.data === 'string'
+		? statusResponse.data
+		: (isAvailable ? DEFAULT_200_MSG : (isUnavailable ? DEFAULT_404_MSG : 'Unknown status'));
```
dawsontoth: This is how we check if the instance is available or not, based on the first systemStatus.

src/features/cluster/InstanceStatusCell.tsx
```diff
@@ -0,0 +1,107 @@
...
+			{canManage && (
+				<div className="flex gap-1">
+					{isAvailable && (
+						<Tooltip>
+							<TooltipTrigger asChild>
+								<Button
+									variant="destructiveGhost"
+									size="icon"
+									className="size-7"
+									onClick={() => setStatus({ ...instanceParams, id: 'availability', status: 'Unavailable' })}
```
dawsontoth, line 68: This is how we set it to be unavailable.

src/features/cluster/InstanceStatusCell.tsx
```diff
@@ -0,0 +1,107 @@
...
+						</Tooltip>
+					)}
+					{isUnavailable && (
+						<Tooltip>
+							<TooltipTrigger asChild>
+								<Button
+									variant="ghost"
+									size="icon"
+									className="size-7"
+									onClick={() => setStatus({ ...instanceParams, id: 'availability', status: 'Available' })}
```
dawsontoth, line 85: And this is how we set it to be available.

Devin-Holland: that looks right. 

await server.operation({ operation: 'set_status', id: 'availability', status: 'Available' });

src/features/cluster/InstanceStatusCell.tsx
Devin-Holland: await server.operation({ operation: 'set_status', id: 'availability', status: 'Unavailable' });

Yeah looks right to me.

src/features/cluster/InstanceStatusCell.tsx
dawsontoth: Cool, thanks! Experimentally it works too. :)

src/features/cluster/InstanceStatusCell.tsx
Devin-Holland: I'm not sure if the order will ever be different, but looking for "availability" status specifically is probably safer as we're trying to drive the available/unavailable based on that specific one.

dawsontoth: @Devin-Holland Fixed!

Devin-Holland (approved):

PR #1165, opened 2026-04-01
feat: Support primary_key over hash_attribute for v5

Devin-Holland:


dawsontoth (approved):

kriszyp (approved):

kriszyp:
This wasn't included in the harper beta.8 because it isn't in the `prod` branch? Is that what we want? (local studio is broken without this, so probably not)

dawsontoth:
I can ship it!

Issue #1175, opened 2025-10-22
Put GH Hash into a user visible place for developers


irjudson:
### Feature Summary

What I would love is for harper - the cli, status, and/or studio server status page to show the detailed version of harper that's running (not just X.Y.Z, but also Git Hash) so that it's easy to tell what you just ran while you're grinding through software updates on 1000 packages/repos.

### Problem This Solves

Makes it easy to find out exactly what detailed verison of harper is running in the instances.

### Proposed Solution

for me, harper status -> harperdb: status: running, version: X.Y.Z (git hash), pid: 2345
same in studio status view

### Priority/Impact

Low - nice to have

### Are you planning to fix this issue?

Maybe, depending on complexity

### First-time contributor support

- [ ] I'm new to contributing and would appreciate guidance on the process
- [ ] I'd like help understanding the project structure
- [ ] I need assistance with setting up the development environment
- [ ] I'm comfortable contributing, but new to this project specifically

dawsontoth:
When we’re using v5 of Harper, or when we have a Studio release, we can link to the corresponding GitHub release. The config page has been linkified accordingly, and the link already in the header has been made more specific.

PR #1179, opened 2026-04-23
feat: Link to releases when possible

dawsontoth:
When we’re using v5 of Harper, or when we have a Studio release, we can link to the corresponding GitHub release. The config page has been linkified accordingly, and the link already in the header has been made more specific.

https://github.com/HarperFast/studio/issues/1175


dawsontoth:
cc @irjudson

PR #1186, opened 2026-05-06
fix: Render expired rows accurately

dawsontoth:
https://harperdb.atlassian.net/browse/STUDIO-679

### DIFF
--- a/src/features/instance/databases/components/TableView.tsx
+++ b/src/features/instance/databases/components/TableView.tsx
@@ -8,6 +8,7 @@ import { Table, TableBody, TableCell, TableHeader, TableHeadSortable, TableRow }
 import { addCommasToNumbers } from '@/lib/addCommasToNumbers';
 import { cn } from '@/lib/cn';
 import {
+	Cell,
 	ColumnDef,
 	flexRender,
 	getCoreRowModel,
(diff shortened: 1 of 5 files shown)


src/features/instance/databases/components/TableView.tsx
```diff
@@ -103,26 +103,43 @@ export function TableView<TData, TValue>({
...
+									className={cn('hover:bg-muted/10 data-[state=selected]:bg-muted', onRowClick && 'cursor-pointer')}
+								>
+									{isExpired
+										? (
+											<TableCell
+												colSpan={row.getVisibleCells().length}
+												className="px-2 py-2 italic text-grey-500"
+											>
+												{original.id ? `${String(original.id)}: ` : ''}
+												{String(original.message)}
```
kriszyp: I'm curious why it has to be converted to string, when it was already determined to be a string? And doesn't `{}` in tsx also implicitly `toString()`?

kriszyp (approved):
Thank you!

src/features/instance/databases/components/TableView.tsx
dawsontoth: It's to appease the TypeScripts because id and message are both unknown, and haven't been narrowed down. Message appears like it _could_ be narrowed, not sure why TypeScript didn't figure that out on its own.

dawsontoth: I didn't like how it was rendering, and so I dug in a bit more and made it nicer. Cleaned up hash vs primary key while I was in there too.

PR #1189, opened 2026-05-07
Show CPU/Memory and cloud instance type on plan details

Devin-Holland:
## Summary
- Adds **Maximum CPU Cores** and **Memory** rows to the "Learn More" plan-details table on the cluster creation page (sourced from `selectedPlan.resourcesPerInstance`).
- For **dedicated plans** only, adds a **Cloud Instance Type** row showing either the Linode (e.g. `g6-dedicated-16`) or GCP (e.g. `n2d-standard-16`) instance type from `selectedPlan.cloudInstanceTypes`.
- Linode is shown when the org's `channel` is `"Akamai"`; GCP is shown otherwise.

## Purpose
The plan details dropdown on cluster creation already exposed Storage and usage limits, but customers/SEs couldn't see the underlying instance shape (CPU/RAM) or which cloud SKU a dedicated plan would provision. This adds that visibility without changing pricing or any other behavior.

## Where to focus review
- `ClusterForm.tsx` — `isAkamai` is computed from `organization?.channel === 'Akamai'`. The channel string itself is never rendered; only the boolean is threaded through.
- `ResourcesPerInstance.tsx` — Cloud Instance Type row is gated on `!!cloudInstanceType`, which is only truthy when `selectedPlan.cloudInstanceTypes` is present (i.e. dedicated plans). Confirm that's the intended gate vs. checking `deploymentDescription.startsWith('Dedicated')`.
- Memory display: `memoryMb / 1024` + `" GB"` — assumes `memoryMb` is base-2 (MiB), per the data model.
- The `isAkamai` prop is threaded `ClusterForm → ClusterDetails → ClusterRegions → RegionFormInputs → ResourcesPerInstance`, mirroring the existing `isEnterprise` plumbing.

## Test plan
- [ ] Open the cluster creation page on a non-Akamai org with a dedicated plan selected → "Cloud Instance Type" row shows the GCP SKU.
- [ ] Open with an Akamai-channel org → "Cloud Instance Type" row shows the Linode SKU.
- [ ] Select a non-dedicated plan → "Cloud Instance Type" row is absent; CPU/Memory still appear when the plan provides them.
- [ ] Confirm no other content in the plan-details table or surrounding form changed.

### DIFF
--- a/src/features/clusters/upsert/components/ResourcesPerInstance.tsx
+++ b/src/features/clusters/upsert/components/ResourcesPerInstance.tsx
@@ -98,12 +98,8 @@ export function ResourcesPerInstance({ selectedPlan, selectedRegion, isEnterpris
 				value: `${humanFileSize(resourcesPerInstance.storageGb * 1000_000_000)}`,
 			},
 			!!resourcesPerInstance && isPositive(resourcesPerInstance.cpuCores) && {
-				label: 'CPU Cores',
+				label: 'Maximum CPU Cores',
 				value: `${humanNumber(resourcesPerInstance.cpuCores)}`,
-			},
-			!!resourcesPerInstance && isPositive(resourcesPerInstance.threads) && {
-				label: 'Threads',
-				value: `${humanNumber(resourcesPerInstance.threads)}`,
 			},
 			!!resourcesPerInstance && isPositive(resourcesPerInstance.memoryMb) && {
 				label: 'Memory',
--- a/src/features/clusters/upsert/components/ResourcesPerInstance.tsx
+++ b/src/features/clusters/upsert/components/ResourcesPerInstance.tsx
@@ -17,7 +17,7 @@ export function ResourcesPerInstance({ selectedPlan, selectedRegion, isEnterpris
 	readonly selectedPlan: SchemaPlan | undefined;
 	readonly selectedRegion: SchemaRegion | undefined;
 	readonly isEnterprise: boolean;
-	readonly cloudProvider: keyof SchemaCloudInstanceTypes;
+	readonly cloudProvider: keyof SchemaCloudInstanceTypes | undefined;
 }) {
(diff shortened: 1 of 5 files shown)


src/features/clusters/upsert/components/ResourcesPerInstance.tsx
```diff
@@ -13,17 +13,21 @@ import { isPositive } from '@/lib/types/isPositive';
...
 }) {
 	const [toggled, setToggled] = useState(false);
 	const onUsageLimitsClick = useCallback(() => {
 		setToggled(!toggled);
 	}, [toggled, setToggled]);
 	const planLimits = selectedPlan?.planLimits;
 	const resourcesPerInstance = selectedPlan?.resourcesPerInstance;
+	const cloudInstanceType = isAkamai
+		? selectedPlan?.cloudInstanceTypes?.linode
+		: selectedPlan?.cloudInstanceTypes?.gcp;
```
dawsontoth: Instead of passing through `isAkamai` as a boolean, can we pas through the instance type? And compute that based on the organization channel. So it would be `linode` or `gcp`, for now. Then you can look at the cloudInstanceTypes here appropriately. Why? So that we can easily add support in the future for other cloud providers, should the fancy strike us, without spreading more booleans through the components.

dawsontoth (changes requested):

src/features/clusters/upsert/components/ResourcesPerInstance.tsx
Devin-Holland: I pushed an update, does this one make more sense now?

dawsontoth (approved):

kriszyp (approved):
Looks great! I think you caught all the nuanced aspects that I wanted to make sure we did (machine type not listed if not explicit provider, "maximum" CPUs).

PR #1193, opened 2026-05-11
feat(status): port analytics-viz dashboards into Status tab

kylebernhardy:
## Summary

Replaces Studio's legacy Status view with a spec-driven analytics dashboard ported from the analytics-viz project, adapted to Studio's `get_analytics` data path and design tokens.

- **7 tabs**: Health → Traffic → Requests → Database → Replication → Storage → Overview, with URL-synced state (`?tab=&range=&refresh=`).
- **Declarative pipeline**: per-metric `MetricSpec` definitions drive aggregation, bucketing, confidence gating, and primitive selection. Derived metrics (error-rate, request-rate, transaction-log-growth) live alongside.
- **Adapter (`useAnalyticsRecords`)**: passes Harper rows through verbatim, surfaces a schema-drift signal, jitters concurrent specs, scopes the query key by `instanceId`.
- **Capability probe**: one cheap `get_analytics` per instance gates the dashboard. While loading, the tab shows an inline message; on error, an "analytics unavailable" message — no legacy fallback.
- **Legacy view removed**: `LocalStatus` / `CloudStatus` / `Status` / `Monitoring` / `MetricVisualization` and the `?legacyStatus=1` escape hatch are all gone.

Two rounds of code review (cloud `/ultrareview` + local multi-agent) were incorporated as separate commits on this branch (P0s, P1s, P2s).

See [`docs/status-redesign.md`](docs/status-redesign.md) for the architecture reference.

## Where to focus review

- **Data path**: `src/integrations/api/instance/status/getAnalytics.ts`, `src/features/instance/status/analytics/hooks/useAnalyticsRecords.ts`, `useAnalyticsCapability.ts` — instance-scoped query keys, schema-drift detection, retry+jitter policy.
- **Pipeline runner**: `src/features/instance/status/analytics/pipeline/pipeline.ts` — bucketing, confidence gating, topN+Other rollup, the perNode/byNode rename.
- **Entry rewire**: `src/features/instance/status/index.tsx` (now an 8-line wrapper) and `src/features/instance/status/analytics/StatusTabs.tsx` for the probe-gate + URL sync.
- **Styling**: chart-surface tokens and `--color-*` aliases at the bottom of `src/index.css`.

## Test plan

- [x] `pnpm exec tsc -b` clean
- [x] `pnpm lint` — 4 pre-existing unicorn warnings, 0 errors
- [x] `pnpm test` — 132 files / 833 passed / 11 skipped
- [x] `pnpm format:fix` clean
- [ ] Manual smoke against a stage Harper: hover every tab, switch time-range presets, verify dark-mode tooltips, verify expanded-chart dialog, verify a deliberate analytics-disabled instance shows the inline "unavailable" message
- [ ] Cross-instance cache isolation in two browser tabs

## Out of scope (follow-ups)

- `TableSizeSnapshot` / `TableSizeTrend` `fillParent` support in the expand dialog
- Playwright E2E
- Server-side Harper changes


dawsontoth (changes requested):
Because it isn't using the actual routing library, it easily corrupts the URLs. For example:

https://dev.studio.harperfabric.com/?tab=health&range=1h&refresh=60000#/org-9fiosn0efqv28ujx/clu-drn531vc09bwkmm9/instance/ins-glnxczcu757ratnx

Notice the query parameters have been injected before the `#`, and I've seen them after it too. When navigating away, these persist, because they aren't a part of the routing. Please use `@tanstack/react-router` for all routing.

kylebernhardy:
> Because it isn't using the actual routing library, it easily corrupts the URLs. For example:
> 
> https://dev.studio.harperfabric.com/?tab=health&range=1h&refresh=60000#/org-9fiosn0efqv28ujx/clu-drn531vc09bwkmm9/instance/ins-glnxczcu757ratnx
> 
> Notice the query parameters have been injected before the `#`, and I've seen them after it too. When navigating away, these persist, because they aren't a part of the routing. Please use `@tanstack/react-router` for all routing.

Will do!

kylebernhardy:
@dawsontoth thanks for the catch — fixed in b15d40f6.

The Status tab now routes `tab` / `range` / `refresh` through `@tanstack/react-router` (`useSearch` + `useNavigate`), matching the patterns used elsewhere in Studio (e.g. `EditorViewProvider`, the auth flows). The manual `URLSearchParams` + `history.pushState` path is gone, so params stay inside the hash where they belong.

I also added a `navigate({ search: undefined, replace: true })` on unmount so leaving `/status` for a sibling route strips the params — that addresses the "they persist when navigating away" half of your comment. URL-sync tests were updated to mock the router hooks rather than `window.location.search`.

Notes on what I deliberately did *not* do, in case you'd prefer otherwise:
- No `validateSearch` on the route — no other Studio route uses it, so adding one here would be a new pattern. Happy to introduce it if you'd rather have schema-level coercion.
- Kept whole-object `navigate({ search: { ... } })` instead of the `(prev) => ...` updater form, since that's what the rest of the repo does.

PR #1204, opened 2026-05-19
fix: Auto-pick http when entering localhost

dawsontoth:
If you’re self-managing a localhost or 127.0.0.1 instance, we can auto-pick http:// for you to make things easier.

https://harperdb.atlassian.net/browse/STUDIO-575

### DIFF
--- a/src/features/clusters/upsert/components/InstanceFormInputs.tsx
+++ b/src/features/clusters/upsert/components/InstanceFormInputs.tsx
@@ -75,6 +75,16 @@ export function InstanceFormInputs({
 								autoCapitalize="none"
 								autoComplete="off"
 								autoCorrect="off"
+								onChange={e => {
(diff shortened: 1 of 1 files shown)


src/features/clusters/upsert/components/InstanceFormInputs.tsx
```diff
@@ -75,6 +75,16 @@ export function InstanceFormInputs({
 								autoCapitalize="none"
 								autoComplete="off"
 								autoCorrect="off"
+								onChange={e => {
```
dawsontoth, line 78: I haven't sorted out a good pattern to memoize these yet, since the field we're rendering based on is dynamic too. The types for that field are the hard part to capture in an extracted component.

kriszyp (approved):

PR #1211, opened 2026-05-26
fix: Add titles and arias in importants spots

dawsontoth:
https://harperdb.atlassian.net/browse/STUDIO-38

### DIFF
--- a/src/components/StudioCloud.tsx
+++ b/src/components/StudioCloud.tsx
@@ -1,7 +1,12 @@
 import { useOnRouteLoadTracker } from '@/integrations/datadog/datadog';
-import { Outlet } from '@tanstack/react-router';
+import { HeadContent, Outlet } from '@tanstack/react-router';
 
 export function StudioCloud() {
 	useOnRouteLoadTracker();
-	return <Outlet />;
+	return (
+		<>
+			<HeadContent />
+			<Outlet />
+		</>
+	);
 }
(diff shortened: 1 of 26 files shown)


kriszyp:
`/review`:
Bug — aria-disabled on <a> doesn't prevent navigation

Both GitHubAuthenticationButton and GoogleAuthenticationButton now use:

<a aria-disabled={disabled || undefined} ...>
aria-disabled is a cosmetic ARIA attribute on anchors — it does not block clicks or navigation the way disabled does on a <button>. When disabled=true, the link will still fire and navigate. Options:

Add onClick handler that calls e.preventDefault() when disabled is true
Conditionally omit href when disabled (changes rendering but stops navigation)
Keep the inner <button disabled> for the disabled state only
This is a regression from the old <button disabled> approach which had native prevention.

dawsontoth:
@kriszyp it's incorrect. I only disable them visually so that if someone clicks them, I can guide them. Its recommendation `Add onClick handler...` is what we're already doing.
```
	const onOAuthClick = useCallback((e: MouseEvent) => {
		if (!acceptTerms) {
			setFlashTerms(true);
			setTimeout(() => setFlashTerms(false), 1000);
			e.preventDefault();
			return false;
		}
	}, [acceptTerms]);
```

dawsontoth:
... and there _wasn't_ a "native" state before, either, it was incorrect on that, since it was using this onClick prevention already too.

kriszyp (approved):

PR #1343, opened 2026-06-18
feat: Resizable, draggable log and record modals

### LINKED ISSUE #1342: Make modals resizable
I think the modal when clicking a log or a database record is way too small. it would be nice if it could be resized

dawsontoth:
Closes #1342.

## Problem

The modal shown when clicking a log entry or a database record was fixed at a small size — too small to comfortably read a long log message or a wide JSON record.

## What changed

A new opt-in `resizable` prop on `DialogContent` (in `src/components/ui/dialog.tsx`, backed by a new `useResizableDialog` hook) makes a dialog resizable and draggable. It's enabled on the three record/log modals:

- `ViewLogModal` (click a log entry)
- `EditTableRowModal` (click a database record)
- `AddTableRowModal` (add a record)

### Behavior

- **Resize** from any edge or corner (8 handles).
- **Size is remembered** in local storage under a single shared key (`ResizableModalSize`), so resizing one modal is remembered across all of them.
- **Never exceeds the browser**: the size is always clamped to the viewport. It shrinks when the window shrinks and restores the preferred size when there's room again.
- **Drag** the title bar (grip at the top) to move the modal aside and see what's behind it. While dragging, the **backdrop drops to 0% opacity** and the modal can be pulled **past the screen edge**; on release it **snaps fully back into view**.
- Each modal **opens centered**. The Monaco editors now flex to fill the available height (`automaticLayout`) instead of a fixed height.

### Snappy drag/resize (follow-up)

Dragging initially felt laggy: the content carried an implicit `transition: all 200ms` (from `duration-200` + the default `transition-property: all`), so every move eased 200ms behind the cursor, and updating `left`/`top` via React state re-laid-out the Monaco subtree each frame. Now the gesture is applied **imperatively + coalesced to one update per `requestAnimationFrame`**: dragging uses a compositor `translate3d` (no reflow, no re-render) and resizing sets dimensions directly; React state is committed only on release. The transition is removed (open/close animations are unaffected — they're CSS animations). Result: the modal tracks the cursor 1:1.

### Auto-copy `.env.local` into worktrees (follow-up)

Unrelated dev-tooling nicety: only `.env.local.example` is checked in, so a fresh git worktree has no `.env.local` and `npm run dev` starts unconfigured. Added a `SessionStart` hook (`.claude/hooks/copy-env-local.sh`) that copies `.env.local` from the main worktree into a linked worktree when it's missing. Idempotent and safe — never overwrites an existing file, never touches the main checkout, always exits 0.

## Verification

Verified in the live preview against a real instance:

- ✅ Resize via corner/edge handles; new size persists to local storage.
- ✅ Size shared across all three modals (resized the log modal → the Add/Edit record modals opened at the same size).
- ✅ Shrinking the window clamps the modal into view (1120×890 → 868×608 at a 900×640 viewport); growing it back restores 1120×890.
- ✅ Dragging makes the backdrop transparent, allows dragging past the edge, and snaps back into view on release.
- ✅ Drag now tracks the cursor exactly 1:1 via `translate3d` with no easing; transform cleared cleanly on release (no flash).
- ✅ `.env.local` hook copies from the main worktree when missing, no-ops when present, exits 0.
- ✅ Full test suite (892 passing), `tsc`, oxlint, and dprint all green.

### DIFF
--- a/.claude/hooks/copy-env-local.sh
+++ b/.claude/hooks/copy-env-local.sh
@@ -1,0 +1,32 @@
+#!/usr/bin/env bash
+# SessionStart hook: when working in a linked git worktree, copy .env.local from the main
+# worktree so the dev server (which loads it via `dotenv -e .env.local`) is configured without
+# any manual setup.
+#
+# Safe by design: it only ever *creates* .env.local when it is missing, never overwrites an
+# existing one, and never touches the main checkout. It always exits 0 so it can't fail a session.
+
+# Operate from the project root regardless of where the hook was invoked.
+cd "${CLAUDE_PROJECT_DIR:-$PWD}" 2>/dev/null || exit 0
+
+# Only act inside a git working tree.
+git rev-parse --is-inside-work-tree >/dev/null 2>&1 || exit 0
+
+# Already configured — nothing to do.
+if [ -f .env.local ]; then
+	exit 0
+fi
+
+# The main worktree is always the first entry of `git worktree list`.
+main=$(git worktree list --porcelain 2>/dev/null | sed -n '1s/^worktree //p')
+if [ -z "$main" ] || [ "$main" = "$PWD" ]; then
+	# No worktrees, or we *are* the main checkout — leave .env.local management to the user.
+	exit 0
+fi
+
+src="$main/.env.local"
+if [ -f "$src" ] && cp "$src" .env.local 2>/dev/null; then
+	echo "[claude] Copied .env.local from the main worktree so the dev server is configured." >&2
+fi
+
+exit 0
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,5 +1,16 @@
 {
 	"hooks": {
+		"SessionStart": [
+			{
+				"hooks": [
+					{
+						"type": "command",
+						"command": "bash \"$CLAUDE_PROJECT_DIR/.claude/hooks/copy-env-local.sh\"",
+						"statusMessage": "Syncing .env.local for this worktree..."
+					}
+				]
+			}
+		],
 		"Stop": [
 			{
 				"hooks": [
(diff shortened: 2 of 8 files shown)


cb1kenobi (approved):
Looks good, but man do I wish I could resize from the center. :)

dawsontoth:
Issue created ;)

kriszyp (commented):
Really nice work on the performance side — rAF-coalescing the drag, applying `translate3d` on the compositor path so the Monaco subtree never re-renders mid-gesture, and diagnosing/removing the implicit `transition: all` lag are all exactly right. 👏

One thing worth addressing before merge, plus a few optional notes:

**[significant] Listener leak on mid-gesture unmount** — the window `mousemove`/`mouseup` listeners are torn down only inside `onUp`. If the dialog unmounts mid-drag/resize (Escape, route change, or a programmatic close while the button is held), `onUp` never fires, so those listeners — plus the closure capturing the now-stale `node` — persist on `window`. Standard fix: track the active gesture's teardown in a ref and clean it up in a `useEffect` return. Low-probability, but a real leak. (`useResizableDialog.ts` drag/resize handlers.)

**[question] Shared persisted size** — size is persisted under one shared key across all resizable modals, so resizing the log modal also resizes the add-record modal. Documented as intentional — just confirming that's the desired UX (a per-modal key would be a one-liner otherwise).

**[suggestion] Pure helpers untested** — `clampSize` / `clampPosition` / `centerOf` encode the core invariants and are trivially unit-testable; cheap regression protection for an otherwise manual-only feature.

**[suggestion] Unrelated tooling** — the `.env.local` `SessionStart` hook is out of scope for this feature; consider splitting it into its own PR.

Happy to approve once the mid-gesture leak is either fixed or consciously accepted. Thanks!

PR #1351, opened 2026-06-19
fix: Tear down modal drag/resize listeners on mid-gesture unmount

dawsontoth:
Follow-up to [#1343](https://github.com/HarperFast/studio/pull/1343), addressing post-merge review feedback from @kriszyp.

## [significant] Listener leak on mid-gesture unmount — fixed

The window `mousemove`/`mouseup` listeners in `useResizableDialog`'s drag/resize handlers were torn down only inside `onUp`. If a resizable dialog unmounts mid-gesture (Escape, route change, or a programmatic close while the mouse button is held), `onUp` never fires — so those listeners, plus the closure capturing the now-stale content node, leak onto `window`.

Fix: each live gesture's teardown is tracked in a ref (`activeGestureRef`) and run from a `useEffect` cleanup, so the listeners are released on unmount as well as on normal release. `onUp` still calls the same teardown on release and clears the ref; the unmount path calls `teardown()` directly (never `onUp`), so no `setState` fires on an unmounting component.

## [suggestion] Pure helpers untested — added

`clampSize` / `clampPosition` / `centerOf` are now exported and covered by unit tests in `useResizableDialog.test.ts` (12 cases over the clamp/center invariants: minimums, viewport clamping, rounding, off-screen pull-back, oversized-modal pinning). They take explicit `vw`/`vh`, so they test cleanly in the existing node environment.

## Review notes not actioned

- **[question] Shared persisted size** — confirmed **intentional**; keeping the single shared key (resizing any modal sets the remembered size for all). No change.
- **[suggestion] Split the `.env.local` SessionStart hook** — moot: #1343 is already merged, so there's nothing left to split out. Noting for the record.

## Verification

- `vitest run` — 142 files, 904 passed / 11 skipped
- `tsc -b` — clean
- `oxlint` + `dprint check` — clean

### DIFF
--- a/src/components/ui/useResizableDialog.test.ts
+++ b/src/components/ui/useResizableDialog.test.ts
@@ -1,0 +1,68 @@
+import { describe, expect, it } from 'vitest';
+import { centerOf, clampPosition, clampSize } from './useResizableDialog';
+
+// These mirror the constants in useResizableDialog: MIN_WIDTH 360, MIN_HEIGHT 280, MARGIN 16.
+const VW = 1280;
+const VH = 800;
+
+describe('clampSize', () => {
+	it('passes a size that already fits through unchanged', () => {
+		expect(clampSize({ width: 820, height: 680 }, VW, VH)).toEqual({ width: 820, height: 680 });
+	});
+
+	it('clamps below the usable minimum up to MIN_WIDTH/MIN_HEIGHT', () => {
+		expect(clampSize({ width: 100, height: 50 }, VW, VH)).toEqual({ width: 360, height: 280 });
+	});
+
+	it('clamps a size larger than the viewport to the viewport minus a margin on each side', () => {
+		// vw - 2*16 = 1248, vh - 2*16 = 768.
+		expect(clampSize({ width: 5000, height: 5000 }, VW, VH)).toEqual({ width: 1248, height: 768 });
+	});
+
+	it('rounds fractional dimensions to whole pixels', () => {
+		expect(clampSize({ width: 820.4, height: 680.6 }, VW, VH)).toEqual({ width: 820, height: 681 });
+	});
+
+	it('keeps the minimum even when the viewport is smaller than the minimum', () => {
+		// A tiny viewport can't satisfy both constraints; the minimum wins so the modal stays usable.
+		expect(clampSize({ width: 800, height: 600 }, 200, 150)).toEqual({ width: 360, height: 280 });
+	});
+});
+
+describe('centerOf', () => {
+	it('centers the modal within the viewport', () => {
+		expect(centerOf({ width: 820, height: 680 }, VW, VH)).toEqual({ x: 230, y: 60 });
+	});
+
+	it('rounds the centered position to whole pixels', () => {
+		// (1280 - 821) / 2 = 229.5 -> 230, (800 - 681) / 2 = 59.5 -> 60.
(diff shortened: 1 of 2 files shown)


dawsontoth:
Sorry I missed your feedback the first time around!

kriszyp (commented):
Great fix for the #1343 leak! The `activeGestureRef` + `useEffect(() => () => activeGestureRef.current?.(), [])` pattern is exactly right — unmount runs the live gesture's teardown without firing any `setState`, so no React warning and no orphaned window listeners. The geometry helpers now have tests too, nice addition.

One residual gap in the same leak class — **back-to-back gestures without an intervening `mouseup`**: if a gesture's `mouseup` is dropped (button released outside the window, focus/blur during drag), the gesture stays live. When the user then starts a new drag/resize, `startDrag`/`startResize` overwrites `activeGestureRef.current` with the new teardown *without calling the old one*, leaving the prior gesture's `mousemove`/`mouseup` on `window`. It compounds: when the stale gesture's `onUp` eventually fires, `teardown()` unconditionally sets `activeGestureRef.current = null`, clearing the *second* gesture's ref — so a subsequent unmount can no longer clean that one up either.

Two-part fix (applies to both drag and resize):

```diff
 const startDrag = useCallback((event: React.MouseEvent) => {
     event.preventDefault();
+    activeGestureRef.current?.();   // end any still-live prior gesture before starting a new one
     const node = lastNodeRef.current;
```

```diff
 const teardown = () => {
     if (frame) { cancelAnimationFrame(frame); }
     window.removeEventListener('mousemove', onMove);
     window.removeEventListener('mouseup', onUp);
-    activeGestureRef.current = null;
+    if (activeGestureRef.current === teardown) {   // only clear if still ours
+        activeGestureRef.current = null;
+    }
 };
```

Want to fold this in here (keeps the fix atomic) or track it as a fast follow? The single-gesture unmount case — the headline fix — is correct as-is; this is just the remaining sliver of the same leak class.

— *reviewed by Claude Opus 4.8 + Gemini via agy (corroborated)*

dawsontoth:
@kriszyp Folded it in (commit 4d932213) to keep the fix atomic. Both parts, applied to `startDrag` and `startResize`:

1. `activeGestureRef.current?.()` at the top of each handler ends any still-live prior gesture before the new one overwrites the ref — so a dropped `mouseup` can't strand the old gesture's listeners on `window`.
2. `teardown` now clears the ref only when `activeGestureRef.current === teardown`, so a late `onUp` from a stale gesture can't clobber a newer gesture's teardown.

Cross-type works too (a drag tears down a stale resize and vice versa, since the ref holds whichever gesture is live). Tests + `tsc -b` + `oxlint` + `dprint` all green.

dawsontoth:
Follow-up to the scope note above: also reset `isDragging`/`isResizing` in each gesture's `teardown` (commit e875d928), so a dropped `mouseup` can't leave the modal visually stuck (transparent overlay, no text selection), and a starting drag clears a stale resize's flag (and vice versa). An `unmountedRef`, flipped before the unmount teardown, skips the `setState` on that path — keeping the "no setState on unmount" property intact.

kriszyp (approved):
Re-push fully addresses both flagged items — the mid-gesture-unmount listener leak / orphaned ref is fixed and the helper unit tests are in place. CI green. Approving. — Claude (Opus 4.8)

PR #1363, opened 2026-06-22
feat: Add Select all / Deselect all to the Columns selector

### LINKED ISSUE #1325: Add `deselect all` option to the Columns selector to reduce chronic clicker fatigue

dawsontoth:
## Summary

Closes #1325.

The column-visibility dropdown (`PickColumnsDropdown`) only offered per-column checkboxes, so hiding most columns of a wide table meant clicking each one off individually — the "chronic clicker fatigue" the issue describes.

This adds bulk actions at the top of the menu:

- **Select all** / **Deselect all** menu items, with a separator before the per-column checkboxes.
- **Deselect all** addresses the issue directly; **Select all** is paired with it so users aren't stranded in an all-hidden state and can restore in one click.
- Each action is **disabled when it would be a no-op** — Select all greys out when everything is already shown, Deselect all when nothing is.
- The menu **stays open** after a bulk action (matching the existing checkbox behavior), so you can deselect-all then re-tick just the columns you want.

## Changes

- `src/features/instance/databases/components/PickColumnsDropdown.tsx` — the feature.
- `src/features/instance/databases/components/PickColumnsDropdown.test.tsx` — new render test (5 cases).

## Testing

- New render test exercises the real Radix menu: confirms the actions render, that Deselect/Select all flip every column's `aria-checked` in one click, the disabled states are correct, and the menu stays open after a bulk action.
- Full suite green via the pre-commit hook (920 passed, 11 skipped); `oxlint`, `dprint`, and `tsc -b` all clean.

> Note: not verified in a browser preview — the dropdown lives behind the live table-browsing view (sign-in → org → cluster → database → table → Browse Data), which needs a running Harper backend with table data and isn't reachable in the sandbox. The render test covers the behavior instead.

### DIFF
--- a/src/features/instance/databases/components/PickColumnsDropdown.test.tsx
+++ b/src/features/instance/databases/components/PickColumnsDropdown.test.tsx
@@ -1,0 +1,104 @@
+/**
+ * @vitest-environment jsdom
+ */
+import { formatBrowseDataTableHeader } from '@/features/instance/databases/functions/formatBrowseDataTableHeader';
+import { VisibilityState } from '@tanstack/react-table';
+import { cleanup, fireEvent, render, screen, within } from '@testing-library/react';
+import { useState } from 'react';
+import { afterEach, beforeAll, describe, expect, it } from 'vitest';
+import { PickColumnsDropdown } from './PickColumnsDropdown';
+
+// Radix's menu relies on a handful of DOM APIs that jsdom doesn't implement.
+beforeAll(() => {
+	Element.prototype.hasPointerCapture ??= () => false;
+	Element.prototype.setPointerCapture ??= () => undefined;
+	Element.prototype.releasePointerCapture ??= () => undefined;
+	Element.prototype.scrollIntoView ??= () => undefined;
+	if (typeof window.PointerEvent === 'undefined') {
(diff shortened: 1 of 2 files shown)


cb1kenobi (approved):
Seems strange allowing the column header visibility to be toggled, but not the actual column data.

dawsontoth:
@cb1kenobi meaning we could be requesting less data from the API over the wire? True, haven't implemented that yet.

cb1kenobi:
`id` column visible:
image

`id` column hidden:
image

The data is still displayed.

dawsontoth:
@cb1kenobi good catch — the screenshots make it clear this was a real bug, not the over-the-wire question: hiding a column dropped the **header** but the body kept rendering the cell, so the row misaligned (your `id` UUID showing under the `type` header). Root cause was in `TableView` — `TableBodyRow` memoized its cells on `[row, primaryKey]`, but `row.getVisibleCells()` also depends on the table's column-visibility state, and the `row` reference is stable when only visibility changes, so the memo never recomputed. Fixed by threading `columnVisibility` into the memo deps (3d76d3f5), with a regression test. Reducing the columns requested from the API is a separate optimization that's still not implemented.

PR #1379, opened 2026-06-25
Expose the system database in dev and Local Studio

cb1kenobi:
## Summary

Surfaces Harper's internal `system` database in the Studio databases viewer when running in dev or Local Studio.

Harper's `describe_all` operation deliberately omits `system`: server-side it is a **non-enumerable** property on the `databases` object (`harper/resources/databases.ts`), so the handler's `for..in` loop skips it and it never reaches Studio. Studio itself does no filtering — it renders whatever `describe_all` returns. An explicit `describe_database` still resolves `system` by key, so this change fetches it that way and merges `{ system }` into the describe_all result.

The merge lives in the shared `getDescribeAll` query layer, so the sidebar and table view pick it up automatically.

## Purpose

Make the `system` database inspectable locally for debugging, without exposing it on deployed Fabric.

## Gating (`import.meta.env.DEV || isLocalStudio`)

- **Exposed:** any Vite dev server (`dev`, `dev:fabric`, `dev:local`) and the bundled Local Studio build (`build:local`, `VITE_LOCAL_STUDIO=true`) that ships in the `harper` package.
- **Not exposed:** Fabric `stage` / `prod` deploys.

## Worth a look

- **Local Studio exposure is broader than just "dev".** Because the gate includes `isLocalStudio`, the Local Studio UI bundled into the `harper` npm package will show `system` to anyone running Harper locally — not only Studio developers. This was an explicit choice (expose in dev *or* Local Studio); flagging in case we'd rather restrict to `import.meta.env.DEV` only.
- **The merge is central, so all `describe_all` consumers see `system` in dev**, not just the databases viewer — notably the role editor's default-permission calculator and the Chat tool's schema context. Harmless in dev, but intentional reach beyond the viewer.
- **Permissions:** for a non-super-user the explicit `describe_database` on `system` may be denied; the `try/catch` falls back to the unmodified map so the page still renders.

## Notes

- Not user-facing for deployed Fabric (dev-tooling only), so no docs change.
- Cross-model review (lifecycle step 10) was not run — the `cross-model-review` skill isn't available in this environment, and spawning a Claude reviewer is explicitly disallowed. The points above are surfaced here in lieu of that pass.
- Generated by an LLM (Claude Opus 4.8).

### DIFF
--- a/src/integrations/api/instance/database/getDescribeAll.test.ts
+++ b/src/integrations/api/instance/database/getDescribeAll.test.ts
@@ -1,0 +1,88 @@
+import type { InstanceClientIdConfig } from '@/config/instanceClientConfig';
+import type { AxiosInstance } from 'axios';
+import { afterEach, describe, expect, it, vi } from 'vitest';
+
+const DATA_MAP = { data: { dog: { name: 'dog' } } };
+const SYSTEM_TABLES = { hdb_user: { name: 'hdb_user' }, hdb_role: { name: 'hdb_role' } };
+
+// `describe_all` deliberately omits the system database, so the mock mirrors the
+// server: it returns the base map for `describe_all` and the system tables only for
+// an explicit `describe_database` on `system`.
+function makeClient() {
+	const post = vi.fn(async (_url: string, body: { operation: string; database?: string }) => {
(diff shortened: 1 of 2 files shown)


cb1kenobi:
Ignore this. I think this needs to be in Harper, not Studio.

Issue #1367, opened 2026-06-23
Encountering slow first results on databases view in some circumstances


dawsontoth:
SummaryUsers have reported general UI "slowness" when loading the Databases tab.Investigation so farLooking into the network requests:
The record queries (e.g. search_by_conditions) are pretty fast.
The initial describe_all and describe_table calls are taking several seconds.
Questions / next steps
Why would the describe calls take so long?
Are these slow because of record counts being computed?
We should profile the describe_all / describe_table operations to understand the source of the latency and see whether record counting (or some other work) is the bottleneck during the initial Databases tab load.

dawsontoth:
> We've pretty much identified that it's the describe_table/all  call that includes an expensive recordCount query. This blocks the table queries from starting (several seconds), so it looks like the db is slow.

dawsontoth:
> I wonder if we can just make the record count a separate, async call that has a spinner while everything else loads faster

> maybe exact_count false?

PR #1381, opened 2026-06-26
perf(databases): decouple record count from describe to speed up table loading

dawsontoth:
## Summary

Speeds up the Databases tab by decoupling the expensive **record-count scan** from the loading path. Fixes the "slow first results" reported in #1367.

**Root cause:** `describe_all` / `describe_table` compute `record_count` by scanning each table's primary store. `describe_all` does this **serially for every table**, and the table view's record query was blocked on `describe_table` (it needed `primary_key` from it) — so the page couldn't render until the scans finished, making the DB look slow.

## What changed

- **`describe_all` fetched with `skip_record_count`** → the sidebar + schema map return fast (no count scans). Opt-in and keyed by the flag, so the AI tools and the role editor keep their counts.
- **Records render off whichever describe arrives first** — columns and the record query now read schema/`primary_key` from the count-free `describe_all` map (usually ready before `describe_table`), so first results no longer wait on any count.
- **Count is estimate-first, exact-on-demand:**
  - Exact when the server's scan finished within budget → plain `1,234 records`.
  - Estimate otherwise → `~1,234 records` with a hover tooltip showing the likely range and a **Get exact count** button that runs `describe_table` with `exact_count: true` (full scan) only when the user asks.

## Compatibility / rollout

- Pairs with the backend flag in **HarperFast/harper#1498**. Safe to ship independently: an older backend ignores `skip_record_count` (validator allows unknown keys) and the UI falls back to the map's count; the estimate UX keys off `estimated_record_range`, which today's servers already return. Deploying harper first (or together) realizes the speedup and avoids a brief double-scan window.

## Tests / verification

- New `TablePagination.test.tsx` covers the three count states (exact-plain, estimate + action fires, counting-disabled).
- Full suite green (969 passing), `tsc -b` / `oxlint` / `dprint` clean. Live browser verification of the tab was blocked by the preview's sign-in wall, so behavior is verified via the component test.

> Draft: opening for early review; depends on HarperFast/harper#1498 for the full speedup.

### DIFF
--- a/src/features/instance/config/roles/modals/EditRoleModal.tsx
+++ b/src/features/instance/config/roles/modals/EditRoleModal.tsx
@@ -44,7 +44,11 @@ export function EditRoleModal({
 
 	const instanceParams = useInstanceClientIdParams();
 	const [isValidJSON, setIsValidJSON] = useState(true);
-	const { data: instanceDatabaseMap } = useQuery(getDescribeAllQueryOptions(instanceParams));
+	// Permission editing only needs the schema tree (attributes), never record counts, so skip the count
+	// scan -- and share the same count-free describe_all cache entry the Databases tab populates.
+	const { data: instanceDatabaseMap } = useQuery(
+		getDescribeAllQueryOptions({ ...instanceParams, skipRecordCount: true }),
+	);
 	const { data: registrationInfo } = useQuery(
 		getRegistrationInfoQueryOptions(instanceParams),
 	);
--- a/src/features/instance/databases/components/DatabaseTableView.tsx
+++ b/src/features/instance/databases/components/DatabaseTableView.tsx
@@ -167,10 +167,18 @@ export function DatabaseTableView({ instanceDatabaseMap, databaseName, tableName
 	// Exact count is opt-in: it forces a full, unbounded scan, so we only run it when the user asks for
 	// it from the pagination tooltip. Reset the request when the table changes.
 	const [wantExactCount, setWantExactCount] = useEffectedState(false, [databaseName, tableName]);
-	const requestExactCount = useCallback(() => setWantExactCount(true), [setWantExactCount]);
-	const { data: exactCount, isFetching: isExactCountFetching } = useQuery(
-		getTableRecordCountQueryOptions({ ...instanceParams, enabled: wantExactCount, databaseName, tableName }),
-	);
+	const { data: exactCount, isFetching: isExactCountFetching, isError: isExactCountError, refetch: refetchExactCount } =
+		useQuery(
+			getTableRecordCountQueryOptions({ ...instanceParams, enabled: wantExactCount, databaseName, tableName }),
+		);
+	// First click enables the (initially disabled) query; later clicks retry a failed fetch via refetch.
+	const requestExactCount = useCallback(() => {
+		if (wantExactCount) {
+			void refetchExactCount();
+		} else {
+			setWantExactCount(true);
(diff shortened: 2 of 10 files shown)


kriszyp (approved):
Approving — this is a solid perf win. Pulling the per-table count scan off the blocking `describe_all → render` path is exactly right, and the estimate-now / exact-count-on-demand UX keeps unbounded scans behind an explicit user action (`timeout: 0` only on that intentional fetch). Backward-compat against pre-#1498 servers holds up: an old server that ignores `skip_record_count` still returns counts (no UI regression), and `exact_count: false` gives it the cheaper estimate path.

A few things worth a look (none blocking):

1. **`EditRoleModal` still calls `getDescribeAllQueryOptions(instanceParams)` without `skipRecordCount`** — so it hits/populates the separate `[entityId, 'describe_all', false]` cache entry and can fire its own counted `describe_all` if the role editor is opened while browsing a database. Probably intentional (role scoping may want counts), but confirming it's deliberate vs. an overlooked caller.

2. **`getDescribeTable` doesn't pass `exact_count: false`**, so on a new server it still runs the time-bounded (~500ms) count scan per table. The latency win here comes from unblocking columns/records off the count-free `describe_all` map before `describe_table` resolves — the scan itself is decoupled, not eliminated. If that budget is still felt, passing `exact_count: false` to the regular describe_table would remove it entirely (at the cost of no count from describe_table, only from the explicit fetch). Depends on how often the estimate alone satisfies users.

3. **Tiny UX polish:** if the explicit exact-count fetch errors, `isFetching` goes false and the button reverts to "Get exact count" with no error state — the user can't tell a failure from a fresh prompt. Not a correctness issue.

Nice work — happy to see the count decoupling land. — KrAIs (Claude Opus 4.8)

dawsontoth:
Thanks @kriszyp! Addressed all three in 37b0e718:

1. **EditRoleModal** — good catch, it was an overlooked caller, not deliberate. `calculateDefaultPermissions` only reads `.attributes`, never `record_count`, so the role editor now requests `describe_all` with `skipRecordCount: true` and shares the Databases tab's count-free cache entry (`[entityId, 'describe_all', true]`) instead of firing its own counted scan.

2. **`getDescribeTable` `exact_count: false`** — added. One clarification: I kept the count on `describe_table` rather than skipping it, because that estimate is what drives the immediate "~N records" in pagination — eliminating it would mean *no* count until the user clicks "Get exact count". `exact_count: false` is already the default on the servers we run, so this is mainly an explicit guard so any server defaulting to exact still takes the cheap estimate path. The scan stays decoupled (off the blocking path), not eliminated — which matches the intended "estimate now, exact on demand" tradeoff.

3. **Exact-count error state** — fixed. On failure the tooltip now shows "Couldn't get the exact count." and the action becomes "Try again" (which refetches), instead of silently reverting to "Get exact count". Added a test case for it.

_🤖 Posted by [Claude Code](https://claude.com/claude-code)_

kriszyp (approved):
All three notes addressed — thanks for the quick follow-up.

- **EditRoleModal** now shares the count-free cache entry — correct, the role editor only needs schema.
- **`getDescribeTable` → `exact_count: false`** — belt-and-suspenders guard for older servers. Right call.
- **Error state on the exact-count fetch** — "Couldn't get the exact count" + "Try again" is exactly the right UX for a user-initiated explicit action.

Looks good to merge. — KrAIs (Claude Opus 4.8)

PR #1410, opened 2026-07-02
fix(editor): keep the OK button inside the rejected-uploads toast

### LINKED ISSUE #1323: 'Ok' button on Rejected uploads toast error message goes outside of container
**Steps to reproduce:**

1. Locally, create a folder and add a handful of files with a '.' in front of the name of the file.
2. Drag and Drop into your Cluster’s editor.
3. A rejected uploads toast error message will appear for all of the files that weren’t added into the editor.

    1. Notice that the ‘Ok’ button to clear the toast message is appearing outside of the container.

dawsontoth:
## Summary

Fixes #1323 — the OK button on the "Rejected uploads" toast rendered outside the toast container when dotfiles were rejected during a drag-and-drop upload into the cluster editor.

## Root cause

Sonner toasts are fixed-width (356px) flex rows: `[icon] [content column] [action button]`, with the button set to `flex-shrink: 0`. The upload toasts in `DropTarget.tsx` set `descriptionClassName: 'whitespace-pre'`, and `white-space: pre` forbids all line wrapping. A rejection message longer than the toast (e.g. "Sensitive files and folders starting with . are skipped.") therefore widened the content column past the toast's fixed width and pushed the OK button out of the container.

## Fix

Use `whitespace-pre-line` instead of `whitespace-pre` on both upload toasts (the in-progress toast had the same latent bug — a long file name would push its Cancel button out the same way). `pre-line` keeps the intentional newlines between rejected files but lets long lines wrap; sonner already applies `overflow-wrap: anywhere` to toast content, so even unbroken long file paths wrap cleanly.

## Verification

Reproduced the exact sonner DOM + stylesheet in an isolated page and measured with both classes (the real flow needs an authenticated cluster editor):

- **Before** (`whitespace-pre`): OK button overflows the toast by **42px** — matches the screenshot in the issue.
- **After** (`whitespace-pre-line`): messages wrap; the OK button sits fully inside the toast (17px of padding to spare).

| Before | After |
| --- | --- |
| Button pushed outside the toast (clipped at viewport edge) | Text wraps, button inside |

Full test suite (1036 tests), oxlint, dprint, and `tsc -b` all pass.

### DIFF
--- a/src/features/instance/applications/components/ApplicationsSidebar/DropTarget.tsx
+++ b/src/features/instance/applications/components/ApplicationsSidebar/DropTarget.tsx
@@ -116,7 +116,10 @@ export function DropTarget() {
 
 				toast.loading(`Upload in progress...`, {
 					id,
-					descriptionClassName: 'whitespace-pre',
+					// pre-line (not pre): sonner toasts are fixed-width flex rows, so an
+					// unwrappable line wider than the toast pushes the action button
+					// outside the container (#1323).
+					descriptionClassName: 'whitespace-pre-line',
 					description: `${counter} of ${pluralize(filesToUpload.length, 'file', 'files')}
 ${file.name}
 ${humanFileSize(uploadedBytes)} of ${humanFileSize(totalBytes)}`,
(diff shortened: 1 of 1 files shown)


kriszyp (approved):
Reviewed via Claude review-queue — CSS fix correctly resolves the toast overflow. LGTM.

PR #1425, opened 2026-07-07
feat(logs): stream read_log over SSE with a polling fallback

### LINKED ISSUE #1422: Use SSE for Logs
Let's subscribe to log changes instead of polling.

dawsontoth:
Closes #1422.

## What

Adds a live **SSE tail** of instance/cluster logs so the Logs view subscribes to log changes instead of polling every 5s. The old "Auto refresh (every 5s)" toggle becomes a **Live** toggle:

- **Live on + connection supports streaming** → opens an SSE stream of `read_log` (`Accept: text/event-stream`) and appends entries as they arrive. Label: *"Live (streaming)"*.
- **Live on + streaming unsupported/failed** → transparently falls back to the existing 5s React Query polling. Label: *"Live (polling every 5s)"*.
- **Live off** → a single fetch + manual Refresh, exactly as before.

The buffered React Query snapshot is always present: it seeds the backlog instantly and becomes the polling source on fallback, and live entries are merged onto it (de-duplicated), so the table never flickers when the transport switches.

## How

This mirrors the existing **deploy** SSE plumbing (`resolveInstanceConnection` / `parseSSEStream` / `errors`), reusing it rather than adding a new transport.

- `integrations/api/instance/status/streamReadLog.ts` — an **open-ended** SSE consumer for `read_log`. Unlike `streamOperation` (built for request/response ops that end with a terminal `done` and has a 120s idle watchdog), a log tail may idle indefinitely with no new lines, so there's no idle timeout — teardown is the caller's job via `AbortSignal`. It sends the same operation body as the buffered `getReadLog` (factored into a shared `buildReadLogBody`) and is tolerant of the entry shapes Harper's SSE serializer can emit (single object, batch array, or native-message wrapper).
- `features/instance/log/hooks/useSupportsLogSSE.ts` — gates streaming on a **direct connection** (the fabric-connect proxy buffers `text/event-stream`). Deliberately *not* version-gated: `streamReadLog` detects a non-streaming instance cheaply via the response content-type and falls back, so no version constant to guess/maintain yet.
- `features/instance/log/hooks/useInstanceLogs.ts` — one data source that runs the tail or polls and merges the two. Streamed entries are **batched** (one render per 250ms) so a chatty tail can't thrash React, and retained entries are capped.
- `features/instance/log/hooks/mergeReadLogs.ts` — pure merge/de-dup/cap, newest-first.

## ⚠️ Backend contract (please confirm)

The current `read_log` returns a plain array — I did not find a shipped backend that keeps a `read_log` SSE subscription open. This PR is written to consume a live tail **when the backend provides one**, and to **degrade gracefully to polling** when it doesn't, so it's safe to merge regardless. The client is tolerant of the exact event/data shape; if the backend's streaming contract differs, the parsing in `streamReadLog` is the one spot to adjust. Let's confirm the intended backend behavior when we verify together.

## Verification

Verified in the preview against a real stage cluster (Acme / Anvils):

- Logs render correctly via the buffered path (unchanged). ✅
- Toggling **Live** attempts a direct SSE POST to the instance; the browser can't reach the instance origin from studio (CORS/`ERR_FAILED`), so it throws `SSEUnsupportedError` and **falls back to polling through the proxy** — logs keep updating, label shows *"Live (polling every 5s)"*, no console/React errors. ✅ (This is exactly the graceful-degradation path.)

Still needs **joint verification**: the live-streaming happy path against a streaming-capable instance whose origin is reachable/CORS-permitted from the studio origin (stage instances aren't, from a browser). That's the "verify together in the preview" piece.

- `tsc -b`, `oxlint`, `dprint` clean; full suite `1155 passed`. New unit tests: `streamReadLog.test.ts` (stream/parse/fallback/abort) and `mergeReadLogs.test.ts` (order/dedup/cap).

### DIFF
--- a/src/features/instance/log/components/LogsFiltersForm.tsx
+++ b/src/features/instance/log/components/LogsFiltersForm.tsx
@@ -14,6 +14,11 @@ import { SearchIcon, SlidersHorizontalIcon } from 'lucide-react';
 import { UseFormReturn } from 'react-hook-form';
 import z from 'zod';
 
+// In dark mode `--card`, `--input`, and `--border` all resolve to grey-700, so a default
+// field is invisible against the filters card. Give the fields a distinct fill + visible
+// border — matching the schema editor's treatment (SchemaEditorView `FIELD_SURFACE`).
+const FIELD_SURFACE = 'dark:bg-grey-600 dark:border-grey-500';
+
 export function LogsFiltersForm({
 	form,
 	resetFilters,
@@ -68,7 +73,7 @@ export function LogsFiltersForm({
 												<Input
 													type="search"
 													placeholder="Search logs…"
-													className="pl-9"
+													className={cn('pl-9', FIELD_SURFACE)}
 													value={field.value ?? ''}
 													onChange={field.onChange}
 												/>
@@ -87,7 +92,7 @@ export function LogsFiltersForm({
 									<FormItem>
 										<FormLabel>Log file</FormLabel>
 										<Select onValueChange={field.onChange} value={field.value ?? undefined}>
-											<SelectTrigger className="w-full bg-white dark:bg-grey-700">
+											<SelectTrigger className={cn('w-full bg-white', FIELD_SURFACE)}>
 												<SelectValue placeholder="Select log file" />
 											</SelectTrigger>
 											<SelectContent>
@@ -110,7 +115,7 @@ export function LogsFiltersForm({
 									<FormItem>
 										<FormLabel>Limit</FormLabel>
 										<Select onValueChange={field.onChange} value={field.value ?? undefined}>
-											<SelectTrigger className="w-full bg-white dark:bg-grey-700">
+											<SelectTrigger className={cn('w-full bg-white', FIELD_SURFACE)}>
 												<SelectValue placeholder="Limit" />
 											</SelectTrigger>
 											<SelectContent>
@@ -134,7 +139,7 @@ export function LogsFiltersForm({
 									<FormItem>
 										<FormLabel>Level</FormLabel>
 										<Select onValueChange={field.onChange} value={field.value ?? undefined}>
-											<SelectTrigger className="w-full bg-white dark:bg-grey-700">
+											<SelectTrigger className={cn('w-full bg-white', FIELD_SURFACE)}>
 												<SelectValue placeholder="Level" />
 											</SelectTrigger>
 											<SelectContent>
@@ -161,7 +166,12 @@ export function LogsFiltersForm({
 								<FormItem>
 									<FormLabel>Start date</FormLabel>
 									<FormControl>
-										<Input type="datetime-local" value={field.value ?? undefined} onChange={field.onChange} />
+										<Input
+											type="datetime-local"
+											className={FIELD_SURFACE}
+											value={field.value ?? undefined}
+											onChange={field.onChange}
+										/>
 									</FormControl>
 									<FormMessage />
 								</FormItem>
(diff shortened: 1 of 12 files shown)


dawsontoth:
Will attempt to open a PR to add support in core for SSE on logs.

dawsontoth:
Backend counterpart: HarperFast/harper#1693 — adds the `read_log` SSE live tail this client consumes. The two are meant to land together; this PR degrades to polling when the backend streaming isn't present.

kriszyp (approved):
Reviewed via Claude review-queue — correct SSE→polling fallback, unchanged 5s cadence, clean cleanup on unmount. LGTM.

kriszyp (approved):
Approving — this is a well-designed feature (the polling-fallback degrade path is solid and unit-tested), but a few things worth fixing in a fast-follow:

**Significant**
- **Live→off discards the streamed tail and reverts to a stale snapshot with no signal** (`useInstanceLogs.ts:123`). The reset effect fires on any change to `live`, including turning it off; `query.data` is whatever was fetched *before* the Live session started. A user tailing logs for 10 minutes who toggles Live off sees the table jump back 10 minutes with no indication anything was dropped. This is a regression vs. the prior `isAutoRefreshEnabled` behavior, which left the last-fetched data on screen.
- **`MAX_LIVE_ENTRIES` (2000) ignores the user's `limit`** (`useInstanceLogs.ts:38`). If `limit` is set above 2000 and the tail is busy, oldest live entries drop from `stream.entries` with no backfill (polling is off while streaming), producing a silent gap between the snapshot and the tail. Should be `Math.max(MAX_LIVE_ENTRIES, limit)`.
- **No client-side ceiling on user-supplied `limit`**, forwarded uncapped to the backend (`logFiltersFormSchema.ts` / `buildReadLogBody`). The companion PR (harper#1693) has a confirmed O(n·limit) backend cost bug for large `limit` — this client is the thing putting an unbounded number on the wire. Worth a client-side cap (e.g. 10k) independent of whatever the backend eventually clamps to.

**Cross-PR note (not blocking, but worth tracking):** harper#1693's backend tail has confirmed-open bugs (idle-flush data loss, a transient-error stall, UTF-8 boundary corruption). Because `streamReadLog` deliberately has no idle watchdog ("no new lines is normal for a log tail"), if the backend silently stalls, this client's SSE connection stays open and the polling fallback never engages — the UI shows "Live (streaming)" indefinitely with nothing displayed. Worth a joint test once both land: kill the backend's flush mid-tail and confirm the UI notices (it currently won't).

None of this blocks — happy to see it ship and follow up, just flagging so the Live-off regression and the limit gap don't get lost.

dawsontoth:
Thanks for the thorough review — folded all three of the "Significant" items into this PR (b66938cb):

1. **Live→off no longer discards the streamed tail.** The reset effect is now keyed on entity/filter only (not the `live` toggle), and `logs` merges the streamed entries onto the snapshot regardless of the toggle — so turning Live off leaves everything you were tailing on screen instead of snapping back to the pre-Live snapshot. Re-enabling Live retries SSE via a `retry` action that clears only the fall-back flag, keeping the on-screen entries.
2. **Retain cap now respects `limit`.** Live entries are capped at `max(MAX_LIVE_ENTRIES, limit)` (threaded through the `append` action), so a `limit` above the 2k floor no longer silently drops the oldest streamed lines with no backfill.
3. **Client-side `limit` ceiling.** `clampReadLogLimit` clamps to `MAX_READ_LOG_LIMIT` (10k) in `buildReadLogBody`, so both the buffered fetch and the SSE tail put a bounded number on the wire regardless of what's typed.

Extracted the stream reducer into its own pure module with unit tests (append cap/eviction, retry-preserves-entries, reset), plus a `clampReadLogLimit` test. tsc / oxlint / dprint clean; 34 log+status tests pass.

**On the cross-PR silent-stall note:** partially handled from the backend side — harper#1693 now emits a terminal `error` when a tail read fails past retry, and this client already treats an SSE `error` event as a fall-back trigger (→ polling). What's still open is a *pure* no-error stall (backend stops emitting without erroring): `streamReadLog` has no idle watchdog, so the UI would keep showing "Live (streaming)". Agreed that's worth the joint test you described — I'd lean toward a server-side SSE keep-alive comment + a client idle watchdog on *that*, so a quiet-but-healthy log isn't mistaken for a stall. Happy to do that as the follow-up once we test both together in the preview.

kriszyp (approved):

PR #1431, opened 2026-07-08
feat(chat): collapse back-to-back tool calls into an expandable single line

### LINKED ISSUE #1427: When chatting, collapse approved tool calls to a single line, consider grouping too
To avoid taking up so much space, when the LLM is using tools, collapse stuff down more. A single line, clickable to show more information, can communicate enough. We could also consider collapsing back-to-back tool calls down to a single line too. "Used 5 tools" or something. When approval is required, that can expand out.

dawsontoth:
Closes #1427

## What

When the Harper Agent uses tools, back-to-back tool calls that don't require human approval now collapse into a single line — `Used 3 tools` (or `Used getComponents` for a lone call, `Using 2 tools...` with a spinner while running). Clicking the line expands the full tool cards; clicking it again collapses them back down.

- Tool calls **awaiting approval stay fully expanded** so the Approve / Always Approve / Deny buttons are always visible.
- Visible text breaks up groups, so tool calls interleaved with narration group naturally per run. Invisible parts (step markers, reasoning) do **not** split a run.
- The summary shows the tool's own icon for single calls (wrench for groups) and a check / error / spinner status on the right.
- A **thinking indicator** (animated dots bubble) fills the dead air whenever a request is in flight but nothing visible is in progress — before the first token, between a finished tool call and the next part, and while invisible parts stream. Growing text and in-flight tool spinners suppress it.
- Empty assistant bubbles (created before any parts stream in) are no longer rendered.

### Styling polish

- Message bubbles now read as speech from their avatars: a right-triangle caret whose flat side continues the bubble's top edge points out of the speaker's avatar; corners round at a slighter 6px. Avatars stay pinned to the top so nothing shifts while content streams.
- Consistent padding when a tool call is the first thing in a message, and breathing room between tool groups and following text.
- Dark-mode contrast fixes: the `Used N tools` hover, the `Awaiting Approval...` label, and the Always Approve / Deny hover all previously resolved to (or near) the card color in dark mode; they now tint toward the foreground so they read in both themes.

## How

- `groupMessageParts.ts` — pure helper that walks a message's parts and folds consecutive non-approval tool parts into `tool-group` runs; only visible content breaks a run.
- `ToolCallGroup.tsx` — collapsed-by-default summary row (`aria-expanded` toggle) that renders the existing `ToolInvocation` cards when expanded.
- `ThinkingIndicator.tsx` + `shouldShowThinkingIndicator.ts` — dots bubble plus the pure visibility predicate.
- `MessageBubble.tsx` — renders grouped parts instead of raw parts; approval-pending tool parts still render as standalone expanded `ToolInvocation`s; bubbles with no visible content render nothing.

## Testing

- 13 new jsdom/unit tests covering grouping, single-call labels, in-progress label, expand/collapse toggling, approval-stays-expanded, invisible-part grouping, text-part group breaking, and the thinking-indicator predicate. Full suite passes.
- Verified live in the studio preview against stage: collapsed groups on real chat history, expansion toggling, the approval card, thinking dots during real agent round-trips (including between tool rounds), no empty bubbles, and the caret/hover/contrast polish. Light + dark mode checked.

### DIFF
--- a/src/features/instance/applications/components/Chat/Chat.css
+++ b/src/features/instance/applications/components/Chat/Chat.css
@@ -162,6 +162,11 @@
 
 .tool-group {
     margin-top: 0.75rem;
+    margin-bottom: 0.75rem;
+}
+
+.tool-group:last-child {
+    margin-bottom: 0;
 }
 
 .tool-group-summary {
@@ -181,7 +186,7 @@
 }
 
 .tool-group-summary:hover {
-    background: var(--accent);
+    background: color-mix(in srgb, var(--card) 90%, var(--foreground));
     color: var(--foreground);
 }
 
--- a/src/features/instance/applications/components/Chat/Chat.css
+++ b/src/features/instance/applications/components/Chat/Chat.css
@@ -130,22 +130,44 @@
 }
 
 .content {
+    position: relative;
     padding: 0.75rem 1rem;
     border-radius: 12px;
     font-size: 0.9375rem;
     line-height: 1.5;
 }
 
+/* Tail pointing at the speaker's avatar (avatars are pinned to the top of the bubble). */
+.content::before {
+    content: '';
+    position: absolute;
+    top: 10px;
+    width: 0;
+    height: 0;
+    border-top: 6px solid transparent;
+    border-bottom: 6px solid transparent;
+}
+
 .user .content {
     background: var(--primary);
     color: var(--primary-foreground);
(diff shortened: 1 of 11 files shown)


dawsontoth:
What it looks like:
Screenshot 2026-07-08 at 4 40 37 PM
Screenshot 2026-07-08 at 4 40 43 PM

kriszyp (approved):

PR #1436, opened 2026-07-09
fix(router): patch router-core so evicting an in-flight preload doesn't TypeError

### LINKED ISSUE #1387: [RUM] TypeError: Cannot read properties of undefined (reading '_nonReactive') in TanStack Router
## Summary

Datadog RUM error tracking reports a recurring `TypeError` inside `@tanstack/router-core` that has been active since June 2026 and is still occurring in v2.125.4. 1 occurrence was observed in the last 24 hours.

**Datadog Error Tracking:** https://app.datadoghq.com/error-tracking/issue/3b73dfac-5d74-11f1-9916-da7ad0900002

## Error Details

- **Type:** `TypeError`
- **Message:** `Cannot read properties of undefined (reading '_nonReactive')`
- **Location:** `@tanstack/router-core@1.171.6` → `load-matches.js` → `js()`
- **First seen:** v2.110.2 (2026-06-01)
- **Last seen:** v2.125.4 (2026-06-29)
- **Platform:** Browser
- **24h occurrences:** 1

## Context

This error occurs during route match loading in TanStack Router. The `_nonReactive` property read on `undefined` suggests a match or context object is being accessed before it is initialized — likely during navigation to a route that hasn't fully resolved yet, or during concurrent route transitions.

Note: the CLAUDE.md for this repo documents a known TanStack Router upgrade regression (1.133 → 1.170+) where `getParentRoute` / `addChildren` mismatches cause route params to be dropped. This error may be related to that class of issue or to a specific route's match state becoming undefined.

## Impact

- Ongoing since June 2026
- Non-crash error (handled), but indicates a routing code path with a potential undefined dereference
- Affects users navigating in the app

## Suggested Investigation

1. Check if any recently added or modified route has a `getParentRoute` / `addChildren` mismatch (see `src/router/rootRouteTree.ts` and the regression tests in `src/router/__tests__/orgUndefinedRepro.test.ts`).
2. Reproduce by navigating between routes rapidly or directly deep-linking to routes that involve route matches being loaded.
3. Add a null check or guard around the `_nonReactive` access in the router's match loading path if the router version allows it; otherwise consider pinning or upgrading `@tanstack/router-core`.

## Labels

`bug` `rum`

dawsontoth:
Fixes #1387.

## What

Datadog RUM has been recording a handled `TypeError: Cannot read properties of undefined (reading '_nonReactive')` from `router.preloadRoute` — escalated to 10 occurrences across 8 sessions in a day, across many routes.

**Root cause** (upstream, [TanStack/router#7759](https://github.com/TanStack/router/issues/7759)): several code paths in `@tanstack/router-core`'s `load-matches.js` re-look up the match with `getMatch(matchId)` after an `await` and dereference `._nonReactive` without a guard. A preload's match only lives in `cachedMatches`, so if the user navigates (or `router.invalidate()` / cache GC runs) while a hover-intent preload is in flight, the match is evicted, `getMatch()` returns `undefined`, and `preloadRoute` `console.error`s the TypeError — which RUM picks up on every hover-then-navigate race.

We're already on the latest published versions (`react-router@1.170.17` / `router-core@1.171.14`); no released version contains a fix. Upstream has two open fix PRs ([#7003](https://github.com/TanStack/router/pull/7003), [#7006](https://github.com/TanStack/router/pull/7006)).

## How

- **`patches/@tanstack__router-core@1.171.14.patch`** (pnpm `patchedDependencies`): ports upstream PR #7003. A missing match during load now throws an internal `MatchLoadCancelledError` that resolves the evicted match's controlled promises and aborts it; `preloadRoute` swallows it (the preload becomes a quiet no-op). The background stale-reload cleanup also no longer assumes the match still exists. Patched in both `dist/esm` and `dist/cjs`.
- **`src/router/__tests__/preloadEvictionRepro.test.ts`**: regression tests that reproduce the exact production error against the unpatched package (cache GC and `invalidate()` during an in-flight preload) and pass with the patch.
- **`CLAUDE.md`**: documents the patch and what to do on the next router bump (check whether upstream shipped the fix; otherwise re-create the patch — pnpm will hard-error on the version mismatch, so it can't be silently lost).

## Verification

- New regression tests fail on unpatched `router-core` with the exact production TypeError, pass with the patch.
- Full suite: 192 test files / 1266 tests pass; lint + tsc clean.
- Live app (worktree dev server on 5173): confirmed the patched code is what Vite serves, then stress-tested hover-preload → navigate → back races across links — zero console errors/warnings.

Drop the patch once upstream ships the fix.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -38,6 +38,24 @@ Auth updates reach the memoized router via the `RouterProvider` `context` prop p
 `router.invalidate()` effect in `src/AppRouted.tsx` — that invalidate is what re-runs
 `beforeLoad` guards (e.g. the sign-out redirect in `dashboardRoute.ts`), so don't remove it.
 
+## Routing — `@tanstack/router-core` is patched (preload eviction `_nonReactive` TypeError)
+
+`patches/@tanstack__router-core@1.171.14.patch` (wired via `patchedDependencies` in
+`pnpm-workspace.yaml`) ports the fix from TanStack/router PR #7003 for upstream issue
+#7759 / studio #1387: when a hover-intent preload's cached match is evicted mid-flight
+(user navigates, `router.invalidate()`, cache GC), `load-matches.js` re-read the match
+after an `await` and threw `TypeError: Cannot read properties of undefined (reading
+'_nonReactive')`, which `preloadRoute` then `console.error`'d — polluting Datadog RUM on
+every hover-then-navigate race. The patch turns the eviction into a quiet cancellation
+(resolves the evicted match's controlled promises, aborts it, and `preloadRoute` returns
+undefined). Regression tests: `src/router/__tests__/preloadEvictionRepro.test.ts` — they
+fail on the unpatched package.
+
+On the next `@tanstack/react-router`/`router-core` bump the patch will stop applying
+(pnpm errors on the version mismatch — do not just delete it). Check whether upstream
+shipped #7003/#7006 first; if not, re-create the patch against the new version and keep
+the regression tests green.
+
 ## pnpm — dependency overrides go in `pnpm-workspace.yaml`, not `package.json`
 
 This repo uses pnpm 11. `overrides` (and other settings like `minimumReleaseAge`,
--- a/patches/@tanstack__router-core@1.171.14.patch
+++ b/patches/@tanstack__router-core@1.171.14.patch
@@ -1,0 +1,330 @@
+diff --git a/dist/cjs/load-matches.cjs b/dist/cjs/load-matches.cjs
+index c09976a80c2c112b7dac6f941e643c81eeaf8d52..a07b029f79682a0d84d9e69cea1ee3361927e93a 100644
+--- a/dist/cjs/load-matches.cjs
++++ b/dist/cjs/load-matches.cjs
+@@ -5,6 +5,31 @@ const require_root = require("./root.cjs");
+ const require_redirect = require("./redirect.cjs");
+ let _tanstack_router_core_isServer = require("@tanstack/router-core/isServer");
+ //#region src/load-matches.ts
++const matchRemovedReason = "Match removed before complete load";
++class MatchLoadCancelledError extends Error {
++	constructor() {
++		super(matchRemovedReason);
++		this.name = "MatchCancel";
++	}
++}
++const isMatchLoadCancelledError = (err) => err instanceof MatchLoadCancelledError;
++const getMatchOrThrowCancelled = (inner, matchId, cleanupMatch) => {
++	const match = inner.router.getMatch(matchId);
++	if (match) return match;
++	if (cleanupMatch) {
++		const s = cleanupMatch._nonReactive;
++		s.beforeLoadPromise?.resolve();
++		s.loaderPromise?.resolve();
++		s.loadPromise?.resolve();
++		cleanupMatch.abortController.abort(matchRemovedReason);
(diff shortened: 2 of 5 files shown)


kriszyp:
Good instinct putting a version guard on this — patching router-core internals is exactly the kind of thing that should fail loudly on a dependency bump rather than silently stop applying.

One thing worth fixing before merge: in the "else" branch of the patch (`patches/@tanstack__router-core@1.171.14.patch:76`, mirrored at the esm build `:236`), `cleanupMatch` isn't reassigned after the second `getMatchOrThrowCancelled` call. Trigger: a second concurrent load of the same match is already in flight (`prevMatch._nonReactive.loaderPromise` truthy) when the cache evicts that match during the subsequent `await` — the eviction's own cleanup step gets skipped in that branch.

Smaller one: the `err instanceof MatchLoadCancelledError` checks are per-module-format (cjs/esm each define and check their own class). Not a live bug given the current install topology, but if pnpm/Vite ever resolve two separate instances of router-core (a dual-package hazard), the check silently falls through to generic error handling instead of the cancellation path. A `err?.name === 'MatchCancel'` fallback would be cheap insurance — low priority, not blocking.

— KrAIs

dawsontoth:
@kriszyp Both addressed in 5a19ed43:

- **`cleanupMatch` gap in the concurrent-load branch**: confirmed — with `prevMatch._nonReactive.loaderPromise` in flight, `cleanupMatch` was still `undefined` at the post-`await` `getMatchOrThrowCancelled`, so an eviction there threw without resolving the evicted match's controlled promises. Now `cleanupMatch = prevMatch` is assigned as soon as `prevMatch` is fetched (ESM + CJS), so that branch cleans up like the others. Added a regression test that drives exactly that branch (second preload of the same route joining the first's in-flight loader, then `clearExpiredCache()` mid-await) and asserts both preloads settle with no console errors. Worth noting this deviates from upstream PR #7003, which has the same gap — I'll flag it there.

- **`instanceof` under a dual-package hazard**: `isMatchLoadCancelledError` now falls back to `err instanceof Error && err.name === 'MatchCancel'` in both module formats.

DavidCockerill (approved):
LGTM — faithful port of upstream TanStack/router #7003 as a pnpm `patchedDependencies` patch. It's a real cancellation (resolves the evicted match's controlled promises + aborts the controller), not a TypeError-swallow, so parked concurrent-preload awaiters wake cleanly — I traced the `cleanupMatch` threading and the last-valid-match tracking holds. Ships with three regression tests covering all eviction triggers (GC, invalidate, concurrent preloads) asserting on `console.error`, and pnpm hard-errors (not silent-drops) on the next version bump.

Two low-severity notes, neither blocking:
1. Cancellation is thrown before `firstUnhandledRejection` in `loadMatches`, so if one sibling match is cancelled while another throws a genuine loader error, the cancellation wins and the real error is dropped. In practice the whole preloaded tree is being torn down so the sibling error is moot, and this matches upstream #7003.
2. Carrying the patch ahead of an upstream merge means a manual reconciliation if upstream lands the alternative (#7006) with a different shape — the CLAUDE.md note mitigates this. No action now.

kriszyp (approved):
LGTM — narrowly-scoped, version-guarded router-core patch that fixes the in-flight-preload eviction TypeError, and it's backed by a test. No hot-path cost (this only touches the router's cold eviction path).

Reviewed with a Gemini-assisted pass via review-queue.
— Claude (Sonnet 5)

PR #1438, opened 2026-07-10
feat(databases): render, protect, and filter relationship attributes in browse

### LINKED ISSUE #1202: Relationship fields should not be editable
When you have a table with a relationship, the result set shows both the foriegn key and the relationship field.

When you go to add a new record, it shows the relationship field in the JSON with a value of `null`. If you try click save, it errors with `ClientError: Computed property package may not be directly assigned a value`. The relationship field is not editable. It should not be in the JSON. You have to remove it to be able to save the new record. Same goes for editing an existing record.

So, if we can't edit a relationship field, I think could omit the relationship field from the results table.

### LINKED ISSUE #1278: Show relationships in browse somehow
If you have relationship on the browse page, it shows up as a blank box. Would be nice if it linked to the item in the other table, or something.

### LINKED ISSUE #1286: Implement sub-property selection for complex types in query filters
Enhance the filter functionality to allow users to select sub-properties of complex types (nested objects or relationships) when filtering. For example, in a table with a field `owner: Owner`, allow users to choose to query on `owner.name`, `owner.age`, or other fields within the `Owner` object. This should be implemented in a dialog interface, providing a user-friendly way to construct these queries.

### LINKED ISSUE #1317: Handle relationship'd fields when creating or editing data rows
When creating entries with relationships (e.g., a `Community` with many `Issues`), the UI currently inserts fields like `issues` or `issueIds` with `null` by default, causing errors. These fields should not appear for manual entry, or the UI should handle them gracefully.

dawsontoth:
Closes #1278, closes #1202, closes #1317, closes #1286.

## What

All four open relationship issues on the instance database browse page:

- **Show relationships in browse (#1278)** — relationship columns now resolve to the related records' primary keys and render as link chips (capped at 3, `+N more`) instead of a blank box. Clicking a chip deep-links to the related table with `?filters={pk: value}`; the browse page now applies filters carried in the URL search, so the chip genuinely lands on "the item in the other table".
- **Relationship fields not editable / creating rows (#1202, #1317)** — the Add-record JSON skeleton, the Edit-record JSON, and random-record generation now exclude relationship/`@computed` attributes. The server rejects any write whose record merely contains one of those keys ("Computed property X may not be directly assigned a value", even for `null`), so they must never appear in editor JSON.
- **Sub-property selection in query filters (#1286)** — relationship columns get a filter box with a sub-property picker (the related table's non-synthetic attributes). `.rating >= 4` translates to `search_attribute: ["reviews", "rating"]`, which Harper executes as a join against the related table. The existing comparator syntax (`>=`, `*` suffix, `&` chaining) composes with it, and values are cast by the sub-property's type.

Also fixes plain-cell rendering while in there: objects/arrays stringify as JSON and booleans render as text (previously multi-element arrays rendered as `[object Object],[object Object]` and `false` rendered as nothing).

## How relationships are detected

`describe_table` has no explicit relationship flag. A to-one relationship reports `type: "<RelatedTable>"`; a to-many reports `type: "array", elements: "<RelatedTable>"` — so an attribute is treated as a relationship when its type/elements names a sibling table in the same database (`functions/relationshipAttributes.ts`). Resolution works by naming the attributes in `get_attributes` with nested selects (`{name, select: [relatedPk]}`) **before** `'*'` — a leading `'*'` makes the server return raw records and ignore the rest of the list.

## Version behavior (verified empirically)

| Server | describe reports relationships | outcome |
| --- | --- | --- |
| Harper 4.7.28 (local instance) | yes (`type`/`elements` name the table) | full feature set verified end-to-end: nested selects resolve, `["rel","sub"]` join conditions filter, writes containing relationship keys reject |
| Harper 5.1.18 (stage Anvils cluster) | no (relationship attrs are runtime-only, never persisted to the attribute registry) | features stay dormant; browse behaves exactly as before |

Browser-verified on the worktree dev server against the Anvils cluster: chips render, chip → related-table-with-filter navigation works end to end, the sub-property picker emits the verified wire payload, the Add-record skeleton omits the relationship field, and plain tables (catalog.Product, data.Albums) render unchanged. Test tables/component used for verification were dropped afterwards.

## Tests

New unit coverage for relationship detection (`relationshipAttributes.test.ts`), join-path filter translation (`getSearchByConditions.test.ts`), column construction (`formatBrowseDataTableHeader.test.ts`), and cell key extraction (`RelationshipCell.test.ts`). Full suite: 194 files / 1288 tests pass; `tsc`, oxlint, and dprint clean.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -103,3 +103,26 @@ To drive a Radix dropdown/menu (e.g. `src/components/ui/dropdownMenu.tsx`) in js
   is `menu`. Disabled items have `aria-disabled="true"` and/or `data-disabled`.
 
 Working example: `src/features/instance/databases/components/PickColumnsDropdown.test.tsx`.
+
+## Browse — relationship/computed attributes vary by Harper version
+
+`@relationship` and `@computed` attributes are read-only: Harper rejects any insert/update
+whose record merely CONTAINS the key ("Computed property X may not be directly assigned a
+value", even for `null`). Strip them from record-editor JSON (see
+`functions/relationshipAttributes.ts`).
+
+What `describe_table` reports differs by server version (verified empirically, Jul 2026):
+
+- **Harper 4.7**: relationship attrs ARE listed — to-one as `{attribute, type: '<RelatedTable>',
+  properties: [...]}`, to-many as `{type: 'array', elements: '<RelatedTable>'}` — but there is
+  NO explicit relationship flag: the only signal is that `type`/`elements` names a sibling
+  table. `@computed` attrs are omitted.
+- **Harper 5.1**: relationship attrs are omitted entirely (they're runtime-only, never persisted
+  to the attribute registry); `@computed` attrs appear only with `include_computed: true`, and
+  carry `computed: true`. On 5.1 the relationship browse features simply never activate.
+
+Search wire contract (works on 4.7; validated-but-unresolved on 5.1): `get_attributes` accepts
+nested selects `{name, select: [...]}` to resolve relationships, but a LEADING `'*'` makes the
+server return raw records and ignore the rest of the list — relationship selects must come
+before the `'*'`. `search_by_conditions` accepts `search_attribute: ['rel', 'subProp']` as a
+join into the related table (querying the relationship attribute itself returns nothing/errors).
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -117,12 +117,30 @@ What `describe_table` reports differs by server version (verified empirically, J
   properties: [...]}`, to-many as `{type: 'array', elements: '<RelatedTable>'}` — but there is
   NO explicit relationship flag: the only signal is that `type`/`elements` names a sibling
   table. `@computed` attrs are omitted.
-- **Harper 5.1**: relationship attrs are omitted entirely (they're runtime-only, never persisted
-  to the attribute registry); `@computed` attrs appear only with `include_computed: true`, and
-  carry `computed: true`. On 5.1 the relationship browse features simply never activate.
+- **Harper 5.1**: relationship attrs are omitted from describe entirely; `@computed` attrs appear
+  only with `include_computed: true`, and carry `computed: true`.
 
-Search wire contract (works on 4.7; validated-but-unresolved on 5.1): `get_attributes` accepts
-nested selects `{name, select: [...]}` to resolve relationships, but a LEADING `'*'` makes the
-server return raw records and ignore the rest of the list — relationship selects must come
-before the `'*'`. `search_by_conditions` accepts `search_attribute: ['rel', 'subProp']` as a
-join into the related table (querying the relationship attribute itself returns nothing/errors).
+The 5.1 omission is an **unintentional regression**, not a deliberate change (traced in the Harper
+source, Jul 2026). describe's `pushAtt` never filtered relationships and hasn't changed; what
+changed is the v5.0 "big lift" rewrite made describe read `table.attributes` rebuilt from the
+persisted attribute registry (`attributesDbi`), and relationships are runtime-only — `table()`'s
+persistence loop `continue`s past them so they're never written to the registry (same in 4.7, but
+4.7's describe read the schema-applied in-memory list that still held them). Harper commit
+`3017e097c` (RE-7 / #1183) treats this as a bug but only partially fixes it, and there is NO
+`include_relationships`-style flag to re-surface them.
+
+Because of that, browse ALSO reads relationships from the component `schema.graphql` files
+(`functions/schemaRelationships.ts` → `get_components` + `get_component_file`, parsed with the
+applications schema parser). This is the ONLY source on 5.1, and it's authoritative for the exact
+`@relationship(from:/to:)` key mappings even on 4.7. A schema-only relationship (not in describe)
+renders as a synthesized column: to-one reads the stored foreign key (`from:`) directly from the
+row; to-many links to the related table filtered by the reverse key (`to:`). `getRelationshipInfoMap`
+merges the describe-detected and schema-declared sources.
+
+Search wire contract (works on 4.7; describe-resolvable there, FK-based on 5.1): `get_attributes`
+accepts nested selects `{name, select: [...]}` to resolve relationships, but a LEADING `'*'` makes
+the server return raw records and ignore the rest of the list — relationship selects must come
+before the `'*'`. `search_by_conditions` accepts `search_attribute: ['rel', 'subProp']` as a join
+into the related table; when the sub-property IS the related primary key AND we know the local
+foreign key, we query the FK directly instead (indexed, and works on 5.1's ops API which can't
+execute relationship joins).
--- a/src/features/instance/databases/components/ColumnFilters.tsx
+++ b/src/features/instance/databases/components/ColumnFilters.tsx
@@ -1,3 +1,5 @@
+import { Button } from '@/components/ui/button';
+import { DropdownMenu, DropdownMenuContent, DropdownMenuItem, DropdownMenuTrigger } from '@/components/ui/dropdownMenu';
 import { Form } from '@/components/ui/form/Form';
 import { FormControl } from '@/components/ui/form/FormControl';
 import { FormField } from '@/components/ui/form/FormField';
@@ -5,8 +7,10 @@ import { FormItem } from '@/components/ui/form/FormItem';
 import { FormMessage } from '@/components/ui/form/FormMessage';
 import { Input } from '@/components/ui/input';
 import { TableCell, TableHeader, TableRow } from '@/components/ui/table';
+import { RelationshipAttributeInfo } from '@/features/instance/databases/functions/relationshipAttributes';
 import { cn } from '@/lib/cn';
 import { HeaderGroup } from '@tanstack/react-table';
+import { ChevronDownIcon } from 'lucide-react';
 import { KeyboardEvent, useCallback } from 'react';
 import { UseFormReturn } from 'react-hook-form';
 import { z } from 'zod';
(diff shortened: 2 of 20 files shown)


dawsontoth:
Added per follow-up feedback: the foreign-key column backing a relationship is now collapsed by default — `productId` no longer shows next to `product`; the chips carry the same key values. The pairing is inferred by naming convention (`product`→`productId`/`product_id`, `pets`→`petIds`) since describe doesn't expose the `@relationship` from/to mapping. The key stays listed in the **Columns** picker (unchecked), and re-showing it stores a session override that wins over the default — the escape hatch for unconventional key names or raw-key sorting. Reverse (`to:`) relationships have no local key, so nothing collapses there. Add/edit JSON is unchanged: the foreign key remains the writable field.

dawsontoth:
Two more follow-ups pushed:

**Reverse-key links for legacy relationships.** Tables created under older Harper (e.g. `data.Albums` on the stage Anvils cluster) kept their relationship attributes in the attribute registry, but as bare `{type: 'array'}` — no element type — so values can't be resolved and the cell used to render blank. Those are now detected by matching the attribute name against sibling table names (`tracks` → `Tracks`), and the cell renders a **`Tracks →`** chip linking to the related table filtered by the reverse foreign key (`Tracks?filters={albumId: <row id>}`). The reverse key is inferred from a back-reference to-one relationship on the related table when present (`RelReview.product` → `productId`), else the `<singular table>Id` convention (`Tracks.albumId`). Resolved to-many cells reuse the same link for the `+N more` overflow. Verified live against the Anvils cluster with no mocking: Albums → `Tracks →` chip → exactly the album's 3 tracks, server-side filtered.

**Readable chips.** The chips previously used the badge `outline` variant (plain `text-foreground`), which was indistinguishable from neighboring cell values, especially in dark mode. They now use the app's link blue (`--blue` #4b5eec light / `--blue-300` #8f8dff dark) with hover underline, matching the existing link conventions elsewhere in the app.

DavidCockerill (commented):
The feature reviews well and I'm ready to approve — one build issue to clear first, plus one open question.

### 🔴 Build is red on the current head — fix before merge
Commit `6eb8b07` (one past where I first read the PR) added a **required** `resolvable: boolean` to `RelationshipAttributeInfo` but didn't update the fixtures that construct that interface, so `tsc` fails with 8 errors and **Verify PR is failing**. Offending fixtures: `RelationshipCell.test.ts:5` and the `relationshipInfo` fixture in `getSearchByConditions.test.ts` (~lines 338–406). The PR body's "tsc … clean" note predates that commit. Fix: add `resolvable` (and `reverseForeignKey` where needed) to those fixtures, or make the field optional. Ping me when it's green and I'll approve — the rest checks out.

### Open question (non-blocking): `collapsedForeignKeyNames` can hide a real column by naming convention
In `relationshipAttributes.ts`, if a table has a scalar FK `productId` **and** a to-one relationship `product` that's actually a *reverse* relationship reading from a differently-named key, `relationshipForeignKeyName('product', …)` matches `productId` purely by the `product`+`id` convention and collapses it — even though the relationship doesn't read from that column. So a real, independent stored column is hidden by default, and its relationship chips could show *different* values than the hidden `productId`.

I've graded this a **suggestion, not a blocker**, because it's purely a default-visibility concern: the column def/header/data are all still present, it's one click to restore via the Columns picker, the choice persists (an explicit `true` in `storedColumnVisibility` is spread after the collapsed defaults, so it wins — verified against `PickColumnsDropdown` + the merge `useMemo`), and CSV export uses raw records so the FK still exports. So: is the false-positive-hide acceptable as-is? The `6eb8b07` reverse-key inference looks like it could gate the collapse to non-reverse relationships if you want to tighten it, or a small "a column was auto-hidden" hint would remove the surprise.

Everything else I checked was clean: filter value casting (Date/Boolean sub-properties, the numeric `>= 4` comparator path), `useMemo` deps / no render loop, per-cell cost is cheap with no N+1 (relationships resolve in the single `get_attributes` request), the `EditTableRowModal` synthetic-attr strip, and defensive scalar handling in `relationshipKeyValues`. One cosmetic nit: the `SubPropertyPicker` regex `value.replace(/^\s*\.\S*\s*/, '')` can eat an adjacent comparator if you type `.rat` mid-word with no space — self-corrects on the next keystroke, not worth changing.

dawsontoth:
There are a few more fixes and enhancements coming, but I exhausted my allotment. Will update the PR soon.

dawsontoth:
## Relationship columns now work on Harper 5.1 (via schema files)

The reason `category`/`products` weren't showing: **Harper 5.1's `describe_table` omits `@relationship` attributes entirely**, so Studio had no way to know they exist. I traced this in the Harper source — it's an **unintentional regression** from the v5.0 "big lift" rewrite, not a deliberate change:

- `describe`'s attribute serializer never filtered relationships and hasn't changed.
- v5.0 made `describe` read `table.attributes` rebuilt from the **persisted attribute registry** (`attributesDbi`). Relationship attrs are runtime-only — `table()`'s persistence loop `continue`s past them, so they're never written to that registry. (Same non-persistence in 4.7, but 4.7's `describe` read the schema-applied in-memory list that still held them.)
- Harper commit `3017e097c` (RE-7 / #1183) already treats this as a bug but only partially fixes it, and there's **no `include_relationships`-style flag** to re-surface them.

### The fix (studio-side, works regardless of Harper version)
Browse now **also reads the component `schema.graphql` files** (`get_components` + `get_component_file`, parsed with the existing applications schema parser) to learn relationships. This is the only source on 5.1, and it's *better* than describe — it carries the exact `@relationship(from:/to:)` key mappings. `getRelationshipInfoMap` merges describe-detected + schema-declared sources, so 4.7 and 5.1 both work.

A schema-only relationship (absent from describe) renders as a **synthesized column**:
- **to-one** (`Product.category`, `from: categoryId`) reads the stored FK from the row → chip linking to the related record. `categoryId` collapses behind it.
- **to-many** (`Category.products`, `to: categoryId`) → `Product →` reverse-link filtered by the reverse key.
- filtering a relationship by the related PK queries the local FK directly (indexed; works on 5.1's ops API which can't execute joins).

Also fixed a stale-filter-input bug this surfaced: navigating away from / back to a table now clears the filter inputs in step with the applied conditions.

### Verified end-to-end on the live 5.1.18 Anvils cluster (no mocks)
- **Product** shows a `category` column with a `cat-anvils` chip (read from stored `categoryId`); `categoryId` collapsed.
- **Product** `category` chip → **Category** filtered to that one category.
- **Category** shows a `products` column; **Product →** → Product filtered by `categoryId` to exactly that category's items.
- Fresh-tab console clean; full suite 1308 pass; tsc/oxlint/dprint clean.

Worth filing the describe regression upstream with the Harper team (it also breaks nested GraphQL queries after any reset/hot-reload, per RE-7).

dawsontoth:
@DavidCockerill thanks — both addressed in `13be5613`, build is green.

**🔴 Build (blocker) — fixed.** Commit `6eb8b07` added the required `resolvable` to `RelationshipAttributeInfo` but not to two test fixtures (`RelationshipCell.test.ts:5` and the `relationshipInfo` fixture in `getSearchByConditions.test.ts`), so `tsc -b` failed with 8 errors. My "tsc clean" note was from `tsc --noEmit`, which uses the root tsconfig that **excludes test files** — `tsc -b` (what CI runs) includes them, so I missed it. Fixed the fixtures, and added coverage for the related-primary-key filter shortcut while there. Reproduced `tsc -b` → 8 errors → now exit 0.

**Open question (collapse false-positive) — tightened, not left as-is.** You're right that a convention guess shouldn't hide a real column. `foreignKeyAttribute` is now taken **only** from an exact schema `@relationship(from:)` — never guessed by naming convention. So column collapse, unresolved-cell rendering, and the PK-filter shortcut only ever act on the column the relationship actually reads from. Describe-detected relationships resolve their values via nested `get_attributes` select (they're `resolvable`), so they never needed the guess — meaning the only thing lost is auto-collapse in the narrow "4.x **and** component files unreadable" case, which is the safe direction (show the column rather than hide one we're guessing at). The naming-convention helper now survives solely for reverse-key inference on to-many links, where a wrong guess just filters to nothing rather than hiding data. Note this built on the schema-parsing that landed in `ad876004` (after your review of `6eb8b07`): on 5.1, relationships come **only** from the schema files, so `from:` is always exact there.

Cosmetic nits (the `SubPropertyPicker` regex eating an adjacent comparator mid-word, self-correcting) left as-is per your call. Live 5.1 Anvils behavior re-confirmed unchanged (schema present → exact `from:` still drives collapse); full suite 1310 pass, `tsc -b`/oxlint/dprint clean.

DavidCockerill (approved):
Re-reviewed at `13be5613` — the build break is fixed and the FK-collapse concern is closed at the root. Approving.

- **Build:** verified `tsc -b` passes at this head (and confirmed the negative control — removing `resolvable` reproduces the exact `TS2741` error), so the fixture updates are real, not just a green check. The field was kept required rather than weakened.
- **`collapsedForeignKeyNames`:** nicely done — `foreignKeyAttribute` now comes only from an exact schema `@relationship(from:)` instead of the `name`+`id` naming guess, so a describe-only/reverse relationship no longer hides a real FK column. That's the root-cause fix, not a bandaid.
- The added `schemaRelationships.ts` schema-reading path is bounded and cached (≤26 requests, capped at 25 files, per-instance/60s react-query cache, best-effort) with no N+1 and a cheaper render path — no perf concern.

One low-priority note, non-blocking: the legacy array-name heuristic could still mis-render a plain array column that happens to be named like a sibling table. Edge case, fine to leave.

— DAIvid (Claude Opus 4.8)

dawsontoth:
Thanks for the approve @DavidCockerill. On the low-priority note (legacy array-name heuristic mis-rendering a plain array named like a sibling table) — saved as #1487 rather than folded into this PR, since the clean fix is a small semantics change and you'd marked it fine-to-leave.

Digging in for the issue, the heuristic looks like it has no regime where it actually catches a Harper-modeled relationship: on 5.1 relationships are omitted from describe entirely (so an elements-less `{type:'array'}` there is plain data), and on 4.x real to-many relationships come back with `elements` typed (caught by the typed path, not the name guess). So #1487 recommends just removing it — real relationships are already covered by the schema-reading path and typed describe. Left it as a follow-up so it's a deliberate call rather than a late add to an approved PR.

PR #1468, opened 2026-07-10
fix(applications): preserve in-table comments in the visual schema editor

### LINKED ISSUE #1462: Table with comments can mess up the visual schema editor
## Overview
When creating a new application and schema inside fabric the table is not loaded to the database

## Reproduction Steps
1. Create a new application (React + TypeScript)
2. Inside `schemas/` create a `Users.graphql` file
3. Add a table and attributes
4. Save and restart the cluster
5. Navigate to the `Databases` tab

Results:
The Users table does not display inside the database

Expected Result:
The schema should generate a table and be displayed inside the chosen or default database

dawsontoth:
Fixes #1462.

## Problem

The visual schema editor keeps untouched tables byte-for-byte, but regenerates a table from its parsed model the moment it's edited. The model had **nowhere to store comments that live inside a `@table` block but aren't attached above a field**, so editing a table silently dropped or relocated them. Because unedited files round-trip perfectly, the loss only surfaced on the *first edit* — the "inconsistent state" from the issue.

Three failure modes, reproduced then fixed (each now a regression test):

| Comment location | Old behavior on edit |
|---|---|
| Trailing the last field, before `}` | **dropped** |
| Body is only a comment (no fields) | comment **dropped**, and emitted `type X @table {\n}` — an empty field block, which is a hard **graphql-js syntax error** that makes Harper reject the *whole* schema (the "table never loads to the database" symptom) |
| On the header line: `type X @table { # note` | **relocated** to a standalone line above the first field |

## Fix

- **Model** (`types.ts`): added `trailingComments: string[]` and `headerComment?: string` to `TableModel`.
- **Parser** (`parseSchema.ts`): `parseFields` now returns leftover trivia as `trailingComments` instead of discarding it; new `splitHeaderComment` peels a header-line comment off the body so it isn't misattributed to the first field.
- **Serializer** (`serializeSchema.ts`): `generateTable` re-emits the header comment on the header line and trailing comments before `}` (shared `indentCommentLines` helper).
- **Empty-body guard** (`TableCard.tsx`, `FieldRow.tsx`): the remove button on a table's **last field is disabled** (a GraphQL type needs ≥1 field), and a zero-field table shows a warning in both the summary line and the field list — so the invalid, schema-breaking state is unreachable through normal editing and surfaced when a file already contains it.

## Design note

For the empty-body case I chose a **UI guard + visible warning** rather than blocking Save or silently dropping a zero-field table — matching this codebase's existing pattern (invalid type names are flagged with a hint, not hard-blocked). Silently dropping a table the user is actively editing would be its own inconsistent-state bug. Happy to add a hard Save-time block if preferred.

## Testing

- New regression tests in `parseSchema.test.ts` / `serializeSchema.test.ts` encode the exact three reproductions (edit → serialize) — they fail on the old code.
- Full suite green: **1276 passing / 11 skipped**; `tsc`, oxlint, and commitlint clean.

### DIFF
--- a/src/features/instance/applications/components/ContentActions.tsx
+++ b/src/features/instance/applications/components/ContentActions.tsx
@@ -91,6 +91,9 @@ export function ContentActions({
 	const onNavigateForwardClick = useEmitToListeners('NavigateForward', true);
 	const canNavigateBack = useWatchedValue('CanNavigateBack', false).value;
 	const canNavigateForward = useWatchedValue('CanNavigateForward', false).value;
+	// The visual schema editor sets this when it has validation errors, so we can't
+	// save a schema that would serialize to invalid SDL and break the whole file.
+	const saveBlocked = useWatchedValue('EditorSaveBlocked', false).value;
 	const editorCommandShortcuts = useEditorShortcutLabels();
 
 	const fileIsClean = updatedFileContent === undefined || updatedFileContent === openedEntryContents;
@@ -227,7 +230,11 @@ export function ContentActions({
 							</DropdownMenuTrigger>
 							<DropdownMenuContent align="start">
 								{canEditFile && (
-									<DropdownMenuItem onSelect={onSaveClick} disabled={fileIsClean || isSavingFile}>
+									<DropdownMenuItem
+										onSelect={onSaveClick}
+										disabled={fileIsClean || isSavingFile || saveBlocked}
+										title={saveBlocked ? 'Fix the schema errors before saving' : undefined}
+									>
 										<SaveIcon />
 										Save
 										<DropdownMenuShortcut>{SHORTCUTS.save}</DropdownMenuShortcut>
--- a/src/features/instance/applications/components/SchemaEditorView/FieldRow.tsx
+++ b/src/features/instance/applications/components/SchemaEditorView/FieldRow.tsx
@@ -41,12 +41,15 @@ export function FieldRow({
 	field,
 	typeNames,
 	readOnly,
+	disableRemove,
 	onChange,
 	onRemove,
 }: {
 	field: FieldModel;
 	typeNames: string[];
 	readOnly: boolean;
+	/** True when this is the table's only field: a GraphQL type needs at least one, so removal is blocked. */
+	disableRemove?: boolean;
 	onChange: (field: FieldModel) => void;
 	onRemove: () => void;
 }) {
@@ -107,9 +110,9 @@ export function FieldRow({
 					variant="destructiveGhost"
 					size="icon"
 					className="self-end"
-					disabled={readOnly}
+					disabled={readOnly || disableRemove}
 					onClick={onRemove}
-					title="Remove field"
+					title={disableRemove ? 'A table needs at least one field' : 'Remove field'}
 				>
 					<TrashIcon />
 				</Button>
--- a/src/features/instance/applications/components/SchemaEditorView/FieldRow.tsx
+++ b/src/features/instance/applications/components/SchemaEditorView/FieldRow.tsx
@@ -7,7 +7,6 @@
  */
 import { Button } from '@/components/ui/button';
 import { Select, SelectContent, SelectItem, SelectTrigger, SelectValue } from '@/components/ui/select';
-import { GRAPHQL_NAME_HINT, isValidGraphqlName } from '@/features/instance/applications/lib/schema/graphqlName';
 import {
 	buildType,
 	fieldBaseType,
@@ -42,6 +41,7 @@ export function FieldRow({
 	typeNames,
 	readOnly,
 	disableRemove,
+	error,
 	onChange,
 	onRemove,
 }: {
@@ -50,6 +50,8 @@ export function FieldRow({
 	readOnly: boolean;
 	/** True when this is the table's only field: a GraphQL type needs at least one, so removal is blocked. */
 	disableRemove?: boolean;
+	/** Validation message for this field (bad or duplicate name), shown inline. */
+	error?: string;
 	onChange: (field: FieldModel) => void;
 	onRemove: () => void;
 }) {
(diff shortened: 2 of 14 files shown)


kriszyp (commented):
Really nice fix — moving `headerComment`/`trailingComments` onto `TableModel` and decoupling them from any specific field is the right model. Because [`generateTable` appends `trailingComments` unconditionally](https://github.com/HarperFast/studio/pull/1468/files#diff-88e41f6d04f4538e7801f29cc6d07429ea0dc3c5611e5f9250bfd10f00adf28bR83) after whatever fields exist, add/remove/rename can't strand or relocate the comment, which is exactly the class of bug this was about. The round-trip + idempotency tests are specific and reassuring.

A couple of things I wanted to think through with you:

**Should the empty-body case block Save rather than just warn?** The UI guard in [`TableCard.tsx:92`](https://github.com/HarperFast/studio/pull/1468/files#diff-3f11836df5c1ded8ffd7c7e74c099ef9d4abe6c4661c2fa2160364750623ebdbR92) / [`FieldRow.tsx:44`](https://github.com/HarperFast/studio/pull/1468/files#diff-3bd25e4ba19463cf7cdcd434ff48ebb6cfd3f45c9f6088df97ce754f3bd89aaaR44) nicely prevents *reaching* zero fields by deleting through the UI. But a `@table` body that already has zero real fields on disk (only comments) can still be opened, and if someone edits that table in a way that doesn't add a field (rename the type, toggle a directive), `edited` flips and `generateTable` re-emits a zero-field body — comments preserved now, but still invalid SDL, i.e. the same "table never loads" symptom from the linked issue. I saw the PR body already raises this as an open choice, so this is really just me agreeing it's worth resolving: since it's the exact bug the PR targets, should this be a hard Save-time block (or auto-insert a placeholder field) rather than warn-only? Not a regression — the guard genuinely narrows the surface — just checking whether we want to close the door fully. Worth noting [`serializeSchema.test.ts:217`](https://github.com/HarperFast/studio/pull/1468/files#diff-b9565be041df902cfc7152ba31fbc3ac1fda9de037e003c507c1e2ebb577ca9fR217) only asserts the comment text survives (`.toContain`), not output validity, so nothing would currently catch a change here either way.

**Extra interior indentation in multi-line descriptions.** This one predates the PR, so purely a heads-up: [`indentCommentLines`](https://github.com/HarperFast/studio/pull/1468/files#diff-d04ebfe0140a8e2377172e85ac2f96f9987958121f4ef7cd712e64b12eeffb9aR60) strips *all* leading whitespace from every interior line before reapplying the canonical indent ([`serializeSchema.ts:71`](https://github.com/HarperFast/studio/pull/1468/files#diff-d04ebfe0140a8e2377172e85ac2f96f9987958121f4ef7cd712e64b12eeffb9aR71)), which flattens intentionally-deeper indentation inside a `"""…"""` block (nested lists, code fences). What's new here is that `trailingComments` now routes through the same helper, so the blast radius widens slightly to trailing description blocks. Not something I'd hold the PR on — just flagging in case it's worth a follow-up.

A couple of small suggestions, both low priority:

1. `lines.push(...indentCommentLines(...))` at [`serializeSchema.ts:89`/`:97`](https://github.com/HarperFast/studio/pull/1468/files#diff-d04ebfe0140a8e2377172e85ac2f96f9987958121f4ef7cd712e64b12eeffb9aR89) spreads an array into `push`'s arg list — totally fine in practice for hand-authored schemas, only a theoretical ceiling if a single block hit tens of thousands of lines. A `for` loop / `forEach` removes it for free if you ever want to.
2. Would you be interested in a regression test that drives a field *rename* and a table *rename* through a table carrying both `headerComment` and `trailingComments`? The current tests cover parse/serialize round-trip but not through a rename mutation, and that'd lock in the "comments keyed to the object, not position" invariant the design leans on.

Thanks for chasing down a genuinely annoying data-loss bug — the model change here makes the whole thing more robust going forward.

dawsontoth:
Thanks Kris! Addressed in 6147e8eb:

- **Empty-body → hard Save-time block** (your main question): the visual editor now refuses to write invalid SDL. A guard in the `SaveFile` listener blocks it (covers Cmd+S, not just the button), an `EditorSaveBlocked` watched value disables the toolbar's Save item, and a destructive banner + inline errors say exactly what to fix. This is driven by a new `validateSchema` that's the single source of truth for invalid-file states (empty tables, invalid/duplicate type names, invalid/duplicate field names). Verified live against an instance: a zero-field `Users` table shows the banner and Save goes `aria-disabled`, then fixing the field re-enables it.
- **Nothing caught output validity** (`serializeSchema.test.ts` `.toContain`): validity is now a validation concern — `validateSchema.test.ts` flags the comments-only / zero-field case as `NO_FIELDS`, so a regression here would fail a test.
- **`push(...spread)`**: replaced with an `appendCommentLines` helper loop.
- **Rename regression test**: added — drives a field rename and a table rename through a table carrying both `headerComment` and `trailingComments`, locking in the "comments keyed to the object, not position" invariant.
- **Interior-indentation flattening in `"""…"""`**: left as-is per your heads-up; filed as follow-up #1480.

kriszyp (approved):
Approving — the latest commit (`6147e8e`) closes the key gap from the earlier pass: the empty-body case is now enforced at Save, not just warned. The new `validateSchema` module is a clean single source of truth for the invalid file-states (zero-field tables, invalid/duplicate table & field names), and it's nicely wired end-to-end — the hard guard in the `SaveFile` listener refuses to write, `EditorSaveBlocked` disables the toolbar Save, and the banner + inline highlights tell the user exactly what to fix. Adding `doc` to the listener deps was the right call too (latent stale-closure otherwise).

Both earlier suggestions are also done: the `push(...spread)` ceiling is now a per-line loop (`appendCommentLines`), and there's a solid rename-mutation regression test driving field + table renames through header/trailing comments.

The only thing left is the pre-existing interior-indentation flattening in the comment re-indent helper, which is unchanged by this diff and not a regression — agreed on deferring it; it's orthogonal to the comment-preservation goal.

Nice work, Dawson — this is a tidy, well-tested fix. 🟢

PR #1470, opened 2026-07-10
refactor(config): alphabetize the instance Config nav items

dawsontoth:
> **Stacked on #1402** (the cluster Secrets page). Review the top commit only; everything below belongs to #1402. Retarget to `stage` once #1402 (and its own base #1409) land.

## What

Reorders the instance **Config** page nav so the items read alphabetically. **Overview** stays pinned at the top and keeps its divider (`border-t`) from the rest — only the items below the divider move.

Before → after:

| Before (order added) | After (alphabetical) |
| --- | --- |
| Users | Certificates |
| Roles | Deployments |
| Deployments | Domains |
| Certificates | Roles |
| Domains | Secrets |
| SSH Keys | SSH Keys |
| Secrets | Users |

(“Secrets” is the entry #1402 adds; it slots in alphabetically here so the two PRs don't fight over ordering.)

Only the JSX order in [`src/features/instance/config/index.tsx`](https://github.com/HarperFast/studio/blob/chore/alphabetize-config-menu/src/features/instance/config/index.tsx) changed. Every item keeps its icon and its existing version/permission gate (`certsAvailable`, `deploymentsAvailable`, `secretsSupported`, self-managed/local checks) — nothing was added or removed, just resequenced. Icon imports were already alphabetized on the base branch, so no import change.

## Verification

- Full suite green (1275 passed), `tsc`, `oxlint`, `dprint` all clean via the pre-commit hooks.
- Rendered menu order maps 1:1 to the static JSX order, verified at the source level. Not click-tested in a browser: port 5173 is held by a different checkout and a fresh-origin dev server hits the auth wall, so the live Config page wasn't reachable this session.

### DIFF
--- a/src/features/instance/config/index.tsx
+++ b/src/features/instance/config/index.tsx
@@ -45,40 +45,9 @@ export function ConfigIndex() {
 	);
 	const isSelfManaged = cluster === undefined || clusterIsSelfManaged(cluster);
 
+	// Overview lives at the top (rendered by the nav bars below); the rest are kept alphabetical.
 	const children = canManage && (
 		<>
-			<li>
-				<Link
-					to={buildAbsoluteLinkToPage(params, 'config/users')}
-					className={sharedClasses}
-					inactiveProps={inactiveProps}
-					activeProps={activeProps}
-				>
-					<UsersIcon className="hidden md:inline-block" /> <span className="ms-3">Users</span>
-				</Link>
-			</li>
-			<li>
-				<Link
-					to={buildAbsoluteLinkToPage(params, 'config/roles')}
-					className={sharedClasses}
-					inactiveProps={inactiveProps}
-					activeProps={activeProps}
-				>
-					<HandshakeIcon className="hidden md:inline-block" /> <span className="ms-3">Roles</span>
-				</Link>
-			</li>
-			{deploymentsAvailable && (
-				<li>
-					<Link
-						to={buildAbsoluteLinkToPage(params, 'config/deployments')}
-						className={sharedClasses}
-						inactiveProps={inactiveProps}
-						activeProps={activeProps}
-					>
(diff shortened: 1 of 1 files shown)


kriszyp (approved):
Approving. I traced every list item through the reorder and it's behavior-preserving: no route, gate, icon, or default changed. Each `Link` keeps its exact target and its existing condition (`certsAvailable`, `deploymentsAvailable`, `secretsSupported`, and the `canManage` wrapper), and Overview stays pinned at the top with its divider. It's purely sibling order flipping to alphabetical. Safe to land.

One non-blocking design question worth a thought (not asking you to change anything):

The previous order looks like it may have been deliberate frequency/importance-first — Users and Roles, the two most-used Config pages, sat at the top, and alphabetizing moves them to the bottom purely on spelling. At this list size (~7 items) you're below the threshold where alphabetical's scan-cost win ("binary-search with your eyes") really pays off — under roughly 8-12 items people still read the list linearly top-to-bottom, so importance-ordering arguably serves users a bit better today.

The strongest argument *for* alphabetical here is stability as the list grows: new items (like Secrets) drop into a predictable slot instead of forcing a subjective "where does this rank?" call each time. So this is really future-proofing toward the size where scanning dominates, at the cost of a small present-day frequency win.

If you wanted a middle path: pin the 2-3 highest-traffic entries (Users, Roles) at the top and alphabetize the tail — you get the predictable-insertion benefit without demoting the most-used pages. Purely a judgment call, and I'm genuinely fine with it either way; just flagging that the tradeoff is demoting the most-used pages.

— Claude (Sonnet 5), on behalf of Kris, via review-queue

dawsontoth:
The previous order was by when things were added. In a future PR we could make a data driven choice to break from alphabetical if we are seeing user confusion. We could also take that as a sign that we need to surface certain features in higher priority areas.

PR #1471, opened 2026-07-10
feat(databases): tree nav, database overview, and shared table actions

dawsontoth:
## What & why

Rebuilds the **Databases** left nav on the same `react-complex-tree` tech the Applications tab uses, and adds a **database overview page**. The old dropdown + flat table list only showed one database at a time, had no filter, and kept per-item actions in the right-pane toolbar only.

Now:
- **+ Create a Table** row at the top → **databases** as top-level items → **tables** nested underneath.
- **Type-to-filter** (the library's built-in search).
- **Single-click selects** — a database opens its new overview page, a table opens its data grid; **double-click / chevron expands** a database (the Applications model).

## Highlights

- **Database overview page** — DB size, audit size, table count, and a per-table breakdown (size, columns, primary key, last updated) with record counts backfilled lazily. Sizes/schema are instant from the `describe_all` map already loaded; counts come from a separate counted `describe_all`. Opening the tab now lands on the first database's overview.
- **Right-click menus** — databases (Create a Table, Drop Database) and tables (Add Record(s), Import Data, Export CSV, Drop Table). The table actions live in one `TableContextMenuItems` component **shared by the tree and the overview's table list**, so right-clicking a row in either place opens the same menu and they can't drift.
- **Single modal hub** — the action modals were lifted into an always-mounted `DatabaseActionModals` driven by target-carrying watched values, so an action can target **any** db/table (even one that isn't open, or an empty database), not just the currently-open one. CSV download extracted into `useExportTableCsv` (toolbar keeps its filter-aware export; the tree/overview export the whole table).
- Removed the redundant sidebar **Import Data** button (it's in the header).

## Fixes folded in

- **Context menu didn't re-anchor** on a second right-click while already open (Radix keeps it mounted through its exit animation, so it never re-measured) — the menu appeared over the wrong row. Fixed by remounting the content keyed to the cursor position. **Also applied to the Applications file tree**, which had the same bug.
- **Latent crash** in `useInstanceSchemaTablePermission` (a missing optional chain) that the new right-click menus would trigger for restricted (non-`structure_user`) users.

## No new backend operations

Everything reuses existing operations/modals (`create_table`, `drop_table`, `drop_database`, `insert`, imports, CSV export). Rename was explicitly out of scope.

## Verification

- Browser-verified end-to-end against a real cluster: tree render, type-to-filter, single/double-click, both context menus (including targeting a not-currently-open table and re-anchoring between rows), the overview stats, and the overview-row menu opening the correctly-targeted modal.
- Light + dark mode checked; the tree icons now match the Applications palette (green `+`, orange containers, neutral high-contrast leaves).
- `tsc -b`, oxlint, and dprint pass. `vitest` green (28 tests, incl. new coverage for the tree-item builder).

### DIFF
--- a/src/features/instance/databases/components/DatabaseActionModals.tsx
+++ b/src/features/instance/databases/components/DatabaseActionModals.tsx
@@ -1,0 +1,104 @@
+import { useInstanceClientIdParams } from '@/config/useInstanceClient';
+import { AddTableRowModal } from '@/features/instance/databases/modals/AddTableRowModal';
+import { CreateNewTableModal } from '@/features/instance/databases/modals/CreateNewTableModal';
+import { DeleteDatabaseModal } from '@/features/instance/databases/modals/DeleteDatabaseModal';
+import { DeleteTableModal } from '@/features/instance/databases/modals/DeleteTableModal';
+import { ImportDataModal } from '@/features/instance/databases/modals/ImportDataModal';
+import { InstanceDatabaseMap } from '@/integrations/api/api.patch';
+import { setWatchedValue, useWatchedValue } from '@/lib/events/watcher';
+import { buildAbsoluteLinkToDatabasePage } from '@/lib/urls/buildAbsoluteLinkToDatabasePage';
+import { useQueryClient } from '@tanstack/react-query';
+import { useNavigate, useParams, useRouter } from '@tanstack/react-router';
+import { useCallback } from 'react';
+
+/**
+ * Single, always-mounted home for every database/table action modal. The tree context menu and the
+ * right-pane toolbar both fire the same target-carrying watched values (see `watchedValueKeys.ts`), so
+ * an action can target ANY database/table -- not just the currently-open one -- and there is exactly
+ * one instance of each modal. Mounted unconditionally by the Databases page (once the map loads) so
+ * Drop Database / Create Table work even when no table is open or a database is empty.
+ */
+export function DatabaseActionModals({ instanceDatabaseMap }: { instanceDatabaseMap?: InstanceDatabaseMap }) {
+	const params: {
+		clusterId?: string;
+		instanceId?: string;
+		organizationId?: string;
+		databaseName?: string;
+		tableName?: string;
+	} = useParams({ strict: false });
+	const navigate = useNavigate();
+	const router = useRouter();
+	const queryClient = useQueryClient();
+	const instanceParams = useInstanceClientIdParams();
+
+	const goToTable = useCallback((databaseName?: string, tableName?: string) => {
+		void navigate({ to: buildAbsoluteLinkToDatabasePage({ ...params, databaseName, tableName }) });
+	}, [navigate, params]);
+
+	const { value: createTarget } = useWatchedValue('ShowCreateTable', false);
+	const { value: addTarget } = useWatchedValue('ShowAddTableRecords', false);
+	const { value: importTarget } = useWatchedValue('ShowImportData', false);
+
+	const addInstanceTable = addTarget
+		? instanceDatabaseMap?.[addTarget.databaseName]?.[addTarget.tableName]
+		: undefined;
+
+	const onImported = useCallback(async (databaseName: string, tableName: string) => {
+		// The import may have created a new table (or even database), so refresh the tree too.
+		await queryClient.invalidateQueries({
+			queryKey: [instanceParams.entityId, 'describe_all'],
+			refetchType: 'all',
+		});
+		goToTable(databaseName, tableName);
+		await router.invalidate();
+	}, [queryClient, instanceParams.entityId, goToTable, router]);
+
+	// After a drop, only navigate when the dropped entity is the one currently open -- dropping some
+	// other tree item just needs the `describe_all` invalidation (done by the modal) to rebuild the tree.
+	const onTableDropped = useCallback(({ databaseName, tableName }: { databaseName: string; tableName: string }) => {
+		if (params.databaseName === databaseName && params.tableName === tableName) {
+			goToTable(databaseName, undefined); // fall back to the database overview
+		}
+	}, [params.databaseName, params.tableName, goToTable]);
+	const onDatabaseDropped = useCallback(({ databaseName }: { databaseName: string }) => {
+		if (params.databaseName === databaseName) {
+			goToTable(undefined, undefined); // the page re-picks the first remaining database
+		}
+	}, [params.databaseName, goToTable]);
+
+	return (
+		<>
+			<CreateNewTableModal
+				isModalOpen={!!createTarget}
+				setIsModalOpen={open => setWatchedValue('ShowCreateTable', open ? (createTarget || {}) : false)}
+				databaseName={createTarget ? createTarget.databaseName : undefined}
+				onCreated={goToTable}
+			/>
+			{addTarget && addInstanceTable && (
+				<AddTableRowModal
+					instanceTable={addInstanceTable}
+					isModalOpen
+					setIsModalOpen={open => {
+						if (!open) {
+							setWatchedValue('ShowAddTableRecords', false);
+						}
+					}}
+					refreshTable={() =>
+						void queryClient.invalidateQueries({
+							queryKey: [instanceParams.entityId, addTarget.databaseName, addTarget.tableName],
+						})}
+				/>
+			)}
+			<ImportDataModal
+				isModalOpen={!!importTarget}
+				setIsModalOpen={open => setWatchedValue('ShowImportData', open ? (importTarget || {}) : false)}
+				instanceDatabaseMap={instanceDatabaseMap}
+				databaseName={importTarget ? importTarget.databaseName : undefined}
+				tableName={importTarget ? importTarget.tableName : undefined}
+				onImported={onImported}
+			/>
+			<DeleteTableModal onDeleted={onTableDropped} />
+			<DeleteDatabaseModal onDeleted={onDatabaseDropped} />
+		</>
+	);
+}
--- a/src/features/instance/databases/components/DatabaseActionModals.tsx
+++ b/src/features/instance/databases/components/DatabaseActionModals.tsx
@@ -9,7 +9,8 @@ import { setWatchedValue, useWatchedValue } from '@/lib/events/watcher';
 import { buildAbsoluteLinkToDatabasePage } from '@/lib/urls/buildAbsoluteLinkToDatabasePage';
 import { useQueryClient } from '@tanstack/react-query';
 import { useNavigate, useParams, useRouter } from '@tanstack/react-router';
-import { useCallback } from 'react';
+import { useCallback, useEffect } from 'react';
+import { toast } from 'sonner';
 
 /**
  * Single, always-mounted home for every database/table action modal. The tree context menu and the
@@ -42,6 +43,15 @@ export function DatabaseActionModals({ instanceDatabaseMap }: { instanceDatabase
 	const addInstanceTable = addTarget
 		? instanceDatabaseMap?.[addTarget.databaseName]?.[addTarget.tableName]
 		: undefined;
+	// The Add Records modal needs the table's schema from the fast map. If the target isn't there
+	// (e.g. a stale map after the table was dropped elsewhere), tell the user instead of silently
+	// doing nothing, and clear the trigger.
+	useEffect(() => {
+		if (addTarget && !addInstanceTable) {
+			toast.error(`Couldn't open "${addTarget.tableName}" — it may have just been removed. Try refreshing.`);
+			setWatchedValue('ShowAddTableRecords', false);
+		}
+	}, [addTarget, addInstanceTable]);
 
 	const onImported = useCallback(async (databaseName: string, tableName: string) => {
 		// The import may have created a new table (or even database), so refresh the tree too.
--- a/src/features/instance/databases/components/DatabaseActionModals.tsx
+++ b/src/features/instance/databases/components/DatabaseActionModals.tsx
@@ -69,6 +69,7 @@ export function DatabaseActionModals({ instanceDatabaseMap }: { instanceDatabase
 	return (
 		<>
 			<CreateNewTableModal
+				key={createTarget ? `create-${createTarget.databaseName ?? ''}` : 'create-closed'}
 				isModalOpen={!!createTarget}
 				setIsModalOpen={open => setWatchedValue('ShowCreateTable', open ? (createTarget || {}) : false)}
 				databaseName={createTarget ? createTarget.databaseName : undefined}
@@ -76,6 +77,7 @@ export function DatabaseActionModals({ instanceDatabaseMap }: { instanceDatabase
 			/>
 			{addTarget && addInstanceTable && (
 				<AddTableRowModal
+					key={`add-${addTarget.databaseName}/${addTarget.tableName}`}
 					instanceTable={addInstanceTable}
 					isModalOpen
 					setIsModalOpen={open => {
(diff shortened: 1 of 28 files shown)


src/features/instance/databases/components/DatabaseActionModals.tsx
```diff
@@ -0,0 +1,104 @@
...
+				/>
+			)}
+			<ImportDataModal
+				isModalOpen={!!importTarget}
+				setIsModalOpen={open => setWatchedValue('ShowImportData', open ? (importTarget || {}) : false)}
+				instanceDatabaseMap={instanceDatabaseMap}
+				databaseName={importTarget ? importTarget.databaseName : undefined}
+				tableName={importTarget ? importTarget.tableName : undefined}
+				onImported={onImported}
+			/>
```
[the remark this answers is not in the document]
dawsontoth: Done in a4222dd — added a per-target `key` to the Create/Add/Import modals so their state resets on switch, and dropped CreateNewTableModal's now-redundant manual reset effect. Verified the create-table prefill re-seeds per database (data → "data", harperfast_vite → "harperfast_vite").

src/features/instance/databases/hooks/useExportTableCsv.ts
```diff
@@ -0,0 +1,68 @@
...
+					: getSearchByValue({ ...base, searchAttribute: primaryKey }));
+				toast.loading('Preparing CSV...', { id });
+				const content = response.data as unknown as string;
+				const blob = new Blob([content], { type: 'text/csv' });
+				const url = URL.createObjectURL(blob);
+				const downloadLink = document.createElement('a');
+				downloadLink.href = url;
+				downloadLink.setAttribute('download', `${databaseName}.${tableName}.${new Date().toISOString()}.csv`);
+				downloadLink.click();
+				URL.revokeObjectURL(url);
```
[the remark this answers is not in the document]
dawsontoth: Done in a4222dd — deferred `URL.revokeObjectURL` via `setTimeout(..., 1000)` so it can't cancel the download before Firefox / iOS Safari start fetching the blob.

DavidCockerill (commented):
Really like this — the tree rebuild matches the Applications model, and showing sizes/schema instantly from the fast `describe_all` map while backfilling row counts lazily is exactly the right call for the slow part. Overview numbers all check out on read (DB/audit size off any table's `db_size`/`db_audit_size` with `?? 0` for empty DBs, table count from `Object.keys`, PK falls back `primary_key ?? hash_attribute ?? '—'`), and the empty-instance / stale-link `<Navigate replace>` fallbacks each terminate cleanly with no redirect loop.

One thing worth confirming before merge — a trade-off, not a bug:

**In plain terms:** to fill in the row counts, the overview page asks the server to count **every table in every database on the whole instance**, even though it's only displaying one database — and it fires the moment the Databases tab opens, since that now lands on this overview by default. The server caps the work (it estimates rather than exhaustively scanning once a table gets large, and caches the result ~1 min), so it won't hang the page. But on an instance with many databases/tables it's more server work on tab-open than the old screen, which only counted the one table you were viewing. If that eagerness is intentional (warming the cache for when you expand other DBs), all good — worth a comment saying so. If not, the fix is small: scope the counted `describe` to the database being shown, or request estimates explicitly. (`DatabaseOverview.tsx:43`.)

A few smaller, non-blocking notes:
- `usePermissions.ts:214-216` — the *attribute*-permission function still has the same unguarded `specificPermission?.tables?.[tableName]` deref you just fixed two lines up in the table-level hook; not reachable through this PR's menus (they use the fixed table-level hook), but cheap to guard while you're here.
- `DatabaseActionModals.tsx:107` — `AddTableRowModal` silently renders nothing if its target isn't in the fast map (e.g. a stale map after an external drop); a toast on the miss would be friendlier. Low likelihood.
- Double-click on a database navigates (first click) then expands (second) — consistent with single-click-opens-overview and the differ-guard makes it idempotent, so it's a non-issue functionally; noting since single/double-click races are worth being explicit about.
- No tests lock in the lazy count-backfill, the routing fallbacks, or the tree interactions — worth a couple given how much interaction logic moved here.

— DAIvid (Claude Opus 4.8)

dawsontoth:
Thanks for the thorough read! Addressed in `20e1ccf7`:

- **Whole-instance count on tab-open (`DatabaseOverview.tsx:43`)** — good catch, wasn't a deliberate cache-warm. Switched from one counted `describe_all` (every table in every DB) to per-table `describe_table` for **only the shown database** via `useQueries`. It uses the same cheap estimate path and now shares React Query's cache with the table view, so opening a table is instant and counts stay consistent.
- **`usePermissions.ts:214-216`** — guarded the attribute-permission table deref (returns `false` when the table isn't in the permission map), matching the table-level fix two lines up.
- **`DatabaseActionModals.tsx:107`** — Add Records now toasts and clears the trigger when the target table is missing from the fast map (stale map after an external drop), instead of silently rendering nothing.
- **Double-click navigate-then-expand** — left as-is; the differ-guard makes it idempotent, so it's a no-op functionally. Agreed it's worth being explicit about.
- **Tests** — extracted the tab's redirect rules into a pure `resolveDatabasesRedirect` and covered the fallbacks + loop-freedom (8 cases). The lazy count-backfill and tree-interaction rendering need a query/router harness; I'd suggest those as a focused follow-up rather than growing this PR further.

kriszyp (commented):
Reviewed — this is a clean refactor. Rebuilding the Databases nav on `react-complex-tree` to match Applications, the single always-mounted action-modal hub, and the composite `db/table` ids all read well, and it's nicely commented and browser-verified end-to-end. No blocking issues, and both Gemini bot findings (per-target `key` to reset modal state on target switch; deferring `URL.revokeObjectURL` so the download isn't cancelled on Firefox/iOS Safari) are already applied at head. Two genuine bug fixes folded in are a bonus:
- the permission crash fix ([`usePermissions.ts:189`](https://github.com/HarperFast/studio/pull/1471/files#diff-87b305a53767d8d665d9c4eebfc76342ef9b85d47e5eb3e651088b21188d4844R189)) — the missing optional chain that any restricted (non-`structure_user`) user would trip opening a table context menu; and
- the CSV export hook now has real error handling ([`useExportTableCsv.ts:46`](https://github.com/HarperFast/studio/pull/1471/files#diff-014a3a7baf4939aa61cab600bc97555b9e00a0b4cc039176af28eb4374a12690R46)) where before a failed export left the toast spinning and `isExportingCSV` stuck true.

The tree builder is the right unit to test and it's tested honestly — `buildItems.test.ts` covers sort order, the synthetic root, duplicate table names across dbs, empty databases, the `canManage` gate, and an undefined map. Builder is memoized on `[map, canManage]`, so navigation doesn't rebuild the tree. Good.

Non-blocking items:

**Instance-wide counted `describe_all` on overview open** ([`DatabaseOverview.tsx:33`](https://github.com/HarperFast/studio/pull/1471/files#diff-c7c7652b67ab04646e29650a380391b64c47dccf8e9826cc3376abbfca7fa387R33)). `getDescribeAllQueryOptions` without `skipRecordCount` counts every table in every database, but the overview only shows one database's counts. It's cold-path, cached, and non-blocking (counts backfill with a `…`), so low severity — but on a large multi-database instance it's a heavier scan than needed. If `describe_all` supports a per-database count scope, scoping it would avoid touching unrelated databases.

**Suggestions:** a tiny unit test for the permission crash fix (a permission object where `tables[tableName]` is undefined) would pin that restricted-user regression; and `AddTableRowModal` silently no-ops when the target table isn't in `instanceDatabaseMap` ([`DatabaseActionModals.tsx:86`](https://github.com/HarperFast/studio/pull/1471/files#diff-9f1f30e389b5d55533e5034e4eaddd3686b472ecbfcca51c10a7c3d36730d6a5R86)) — reachable only via a stale tree between invalidation and refetch, so low-risk, but a right-click "Add Record(s)" that does nothing is a confusing dead-end (a toast or refetch-then-open would be friendlier).

One question: does the root "Create a Table" row intentionally seed the modal with the currently-open database rather than no database? Reads deliberate, just confirming.

Perf-wise there's no Harper read/replication hot-path impact here — tree + overview are driven off the single already-loaded `describe_all` map, no N+1, no re-render storm on nav. Looks good to land.

—
🤖 Reviewed by KrAIs (Claude Opus 4.8) on Kris's behalf. *(Gemini/agy local pass unavailable — synthesis is from the Claude pass + the Gemini bot review already on the PR.)*

src/features/instance/databases/components/DatabaseOverview.tsx
```diff
@@ -0,0 +1,177 @@
...
+		}
+	});
+	const allCountsLoaded = countQueries.every((query) => !!query.data);
+
+	// Every table in a database shares the same underlying store, so db_size/db_audit_size are the same
+	// on each -- read them off any table.
+	const anyTable = tableNames.length ? tables[tableNames[0]] : undefined;
+	const dbSize = anyTable?.db_size ?? 0;
+	const auditSize = anyTable?.db_audit_size ?? 0;
+	const totalRecords = allCountsLoaded
```
cb1kenobi: **Low: overview "Records" total can stick on “…” and hides estimated counts**

`allCountsLoaded` requires every per-table `describe_table` query to have `data`. Those queries use `retry: false` (see `getDescribeTableQueryOptions`), so a single failed or perpetually-pending count leaves `totalRecords` `undefined` and the Records stat shows `…` indefinitely, even though every other table counted fine. Separately, the sum mixes exact and estimated (`estimated_record_range`) counts but renders as an exact number, while the per-row labels above prefix estimates with `~` — so the aggregate can silently present an estimate as exact.

Suggested fix: base "loaded" on settled state rather than data presence (e.g. `countQueries.every(q => !q.isPending)`) and sum the counts that resolved, and prefix the total with `~` when any contributing count was estimated.

—
_Generated by Barber AI_

src/features/instance/databases/modals/DeleteTableModal.tsx
```diff
@@ -9,12 +9,13 @@ import { useRouter } from '@tanstack/react-router';
...
 
-export function DeleteTableModal({ databaseName, tableName, onDeleted }: {
-	readonly databaseName: string;
-	readonly tableName: string;
-	readonly onDeleted: (deleted: 'table' | 'database') => void;
+export function DeleteTableModal({ onDeleted }: {
+	readonly onDeleted: (dropped: { databaseName: string; tableName: string }) => void;
 }) {
-	const { value: isModalOpen, trigger } = useWatchedValue('ShowDeleteTable', false);
+	const { value: target, trigger } = useWatchedValue('ShowDeleteTable', false);
```
cb1kenobi: **Low: focus-restore `trigger` is now always undefined (dead plumbing)**

After the refactor, every caller opens this modal via `setWatchedValue('ShowDeleteTable', { databaseName, tableName })` with no third `trigger` argument (`DatabaseTableView`, `TableContextMenuItems`), so `trigger` here is always `undefined` and `attemptToRestoreFocus(trigger)` in `closeModal` is a no-op — focus is no longer returned to the invoking control after the dialog closes. The old `useSetWatchedValue('ShowDeleteTable', true)` handler passed the click event as the trigger. Same applies to `DeleteDatabaseModal`.

Suggested fix: pass the event through at the call sites (`onClick={e => setWatchedValue('ShowDeleteTable', { databaseName, tableName }, e)}`), or drop the now-dead `trigger` destructure + `attemptToRestoreFocus` call.

—
_Generated by Barber AI_

src/features/instance/databases/components/DatabaseOverview.tsx
dawsontoth: Fixed in 283ba95b — the aggregate total now keys off settled count queries (`!query.isPending`) rather than data presence, so a single failed/never-resolving `describe_table` (retry:false) no longer hangs it on "…". It sums the counts that resolved and prefixes the total with "~" when any contributing count is estimated, matching the per-row labels.

src/features/instance/databases/modals/DeleteTableModal.tsx
dawsontoth: Fixed in 283ba95b — dropped the now-always-undefined `trigger` + `attemptToRestoreFocus` from both `DeleteTableModal` and `DeleteDatabaseModal`. Radix's confirmation dialog already restores focus to the invoking control on close, so this was dead plumbing.

dawsontoth:
Thanks @kriszyp! Follow-ups on the non-blocking items:

- **Instance-wide counted `describe_all` on overview open** — good catch; scoped it in `20e1ccf7` (before this review's head). The overview no longer calls `describe_all` with counts; it fetches per-table `describe_table` for **only the shown database** via `useQueries` (cheap estimate path), which also shares React Query's cache with the table view. `describe_all` has no per-database count scope, so per-table was the way to bound it.
- **Permission crash regression test** — added in `0de34c9b`: extracted the pure `checkSchemaTablePermission` (no auth-store imports, so it's testable in a node env) and covered the exact case that used to crash — a restricted user whose database grant lacks the table — plus super/structure-user and the per-action flag.
- **`AddTableRowModal` silent no-op on a stale map** — also handled in `20e1ccf7`: it now toasts ("…may have just been removed. Try refreshing.") and clears the trigger instead of dead-ending.
- **"Create a Table" root row seeding the open database** — yes, intentional. It prefills the modal's Database Name with the currently-open database as a convenience (you're usually working in one); it's editable and falls back to `data` if cleared. The per-database right-click menu's "Create a Table" seeds that specific database.

All green (tsc / oxlint / dprint / 78 tests). Thanks for the thorough pass!

cb1kenobi (approved):

kriszyp (approved):
Re-reviewed at `283ba95` — thanks for the quick turnaround. All three items from my last pass are correctly resolved: the overview record counts are now scoped to the shown database (and you extracted the redirect logic into a pure, tested `resolveDatabasesRedirect` along the way — nice), the restricted-user permission crash now has a real regression test (`checkSchemaTablePermission.test.ts`), and the stale Add-Records target now toasts instead of silently no-op'ing. Approving.

Two small, non-blocking suggestions on the fixes themselves (both low-severity, no rush):

- **Total Records can silently undercount on a settled-but-failed count** ([`DatabaseOverview.tsx:60`](https://github.com/HarperFast/studio/pull/1471/files#diff-c7c7652b67ab04646e29650a380391b64c47dccf8e9826cc3376abbfca7fa387R60)). `allCountsSettled` treats "settled" as ready, but `countByTable[tableName]` is only populated on success, so a failed table contributes `0` via `?? 0` — the total renders as a definite-looking number that's short by that table's count, with no `~`/error indicator (`anyEstimated` only reflects `estimated_record_range`, not query failure). It's strictly better than the old permanently-stuck `…`, and the trigger is rare (a `describe_table` failing for a table present in the fast map — a race with a concurrent drop, or a transient 5xx with `retry: false`), but it trades "obviously stuck" for "quietly wrong," which is the less safe direction for a count. Consider flagging failed-but-settled tables the way estimated ones are flagged (or omitting them from "Tables" for consistency).

- **The attribute-permission crash guard has no unit test** ([`usePermissions.ts:206`](https://github.com/HarperFast/studio/pull/1471/files#diff-87b305a53767d8d665d9c4eebfc76342ef9b85d47e5eb3e651088b21188d4844R206)), unlike its sibling `checkSchemaTablePermission`. Same one-line-guard pattern, same crash class, and it was the one still-open crash vector from the last pass — worth a parity test, ideally by extracting it the same way you did the sibling.

Neither blocks — approving as-is.

—
🤖 Reviewed by KrAIs (Claude Opus 4.8) on Kris's behalf. *(Gemini/agy unavailable this pass — Claude-only.)*

dawsontoth:
Spun the deferred review follow-ups out into tracked issues so this PR can land:

- #1490 — test the sidebar tree interactions (nav / expand / context menus)
- #1491 — test the overview lazy record-count backfill
- #1492 — Add Records: refetch-then-open when the table is missing from the cached map (friendlier than the current stale-map toast)

The items reviewers flagged that were fixed in-PR (per-database count scoping, permission-crash regression test, stale-map toast, settled-based overview total) remain in this branch.

_🤖 via [Claude Code](https://claude.com/claude-code)_

PR #1472, opened 2026-07-11
fix(status): require total, not errors, for the error-rate panel

### LINKED ISSUE #1441: error-rate panel declares wrong required fields, can falsely blank a working chart
`src/features/instance/status/analytics/lib/specRequiredFields.ts:12` declares `'error-rate': ['count', 'errors']`, but the recompute in `pipeline/derived/error-rate.tsx` reads `count` and `total` — never `errors`. On a Harper whose success records lack an `errors` column, MetricPanel shows a false "field unavailable" empty state for a chart that would render fine; conversely a genuinely missing `total` is not flagged.

Fix: change to `['count', 'total']` and add a test pinning the required-fields list to `recomputeErrorRate`.

Filed from a Status tab code review.

Comment generated by kAIle (Claude Fable 5)

kylebernhardy:
Fixes [#1441 — error-rate panel declares wrong required fields, can falsely blank a working chart](https://github.com/HarperFast/studio/issues/1441).

The `error-rate` entry in `DERIVED_REQUIRED_FIELDS` declared `['count', 'errors']`, but `recomputeErrorRate` reads `count` and `total` — never `errors`. Two failure modes:

- On a Harper whose `success` records lack an `errors` column, MetricPanel showed a false "field unavailable" empty state for a chart that would have rendered fine.
- A genuinely missing `total` was never flagged — the recompute silently treats it as 0 and renders a misleading flat 100% error rate instead of the explicit empty state.

The override is now `['count', 'total']`, pinned behaviorally: the new tests omit each declared required field from a fixture and assert the recompute output actually changes, and assert that omitting `errors` changes nothing — so the test fails if either the override or the recompute drifts.

**Where to look:** the one-line change in `src/features/instance/status/analytics/lib/specRequiredFields.ts` plus the new tests in `specRequiredFields.test.ts`.

Verified with `tsc -b`, the full vitest suite (197 files / 1312 passed), dprint, and oxlint. Codex and Gemini cross-model reviews both came back with no findings.

### DIFF
--- a/src/features/instance/status/analytics/lib/specRequiredFields.test.ts
+++ b/src/features/instance/status/analytics/lib/specRequiredFields.test.ts
@@ -1,4 +1,6 @@
 import { describe, expect, it } from 'vitest';
+import { recomputeErrorRate } from '../pipeline/derived/error-rate.tsx';
+import type { AnalyticsDataPoint, TimeRange } from '../types/analytics.ts';
 import { getSpecRequiredFields } from './specRequiredFields.ts';
 
 describe('getSpecRequiredFields', () => {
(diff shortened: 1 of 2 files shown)


kriszyp (approved):
LGTM — the override now mirrors what the recompute actually reads (`count`/`total`, never `errors`), and the divide-by-zero guard (`sumCount === 0 → null`) plus the `total === 0` = 100%-error case both check out. Re-derived the fixture math by hand; all cases hold. Reviewed via review-queue.

— Claude (Sonnet 5)

PR #1474, opened 2026-07-12
fix(status): clamp heatmap keyboard focus when the matrix shrinks

### LINKED ISSUE #1443: HeatmapMatrix keyboard focus becomes unreachable when the matrix shrinks
The roving-tabindex `active` state in `src/features/instance/status/analytics/primitives/HeatmapMatrix.tsx:297` is never clamped when `rows`/`cols` shrink after a data refresh. If `active` is `[5,2]` and the new matrix has 3 rows, no cell gets `tabIndex=0` and the whole grid drops out of the tab order.

Fix: clamp `active` when data changes (effect, or derive `isActive` with `Math.min(active[0], rows.length - 1)` / same for cols).

Filed from a Status tab code review.

Comment generated by kAIle (Claude Fable 5)

kylebernhardy:
Fixes [#1443 — HeatmapMatrix keyboard focus becomes unreachable when the matrix shrinks](https://github.com/HarperFast/studio/issues/1443).

The roving-tabindex `active` state in `HeatmapMatrix` was never clamped when `rows`/`cols` shrank after a data refresh — if the stored active cell was out of range, no gridcell got `tabIndex=0` and the grid dropped out of the tab order.

## Change

- Derive the effective active cell at render time (`Math.min(active[0], rows.length - 1)`, same for cols) — no effect, no extra render. Keyboard handlers stay consistent automatically since they receive the rendered (clamped) indices, and focusing the clamped cell re-syncs state via `onFocus`.
- Regression test: moves focus to the last row/col via keyboard, re-renders with a 1×2 matrix, asserts exactly one gridcell keeps `tabIndex=0` (clamped to the last surviving row/col) and arrow keys still navigate from it. Fails on `stage` without the fix.

Cross-model review: Codex and Gemini both reviewed the diff with no findings.

### DIFF
--- a/src/features/instance/status/analytics/__tests__/primitives/heatmap-matrix.test.tsx
+++ b/src/features/instance/status/analytics/__tests__/primitives/heatmap-matrix.test.tsx
@@ -132,6 +132,46 @@ describe('HeatmapMatrix primitive', () => {
 		expect(legend.getAttribute('aria-label') ?? '').not.toMatch(/count-weighted-mean.*approx/i);
 	});
 
+	it('clamps the roving tabindex when the matrix shrinks (regression #1443)', () => {
+		const { rerender } = render(<HeatmapMatrix data={data} theme="light" />);
+		// Move the active cell to the last row/col via keyboard: End then
+		// ArrowDown lands on src-2 → dest-c ([1, 2]).
+		const cells = screen.getAllByRole('gridcell');
+		const first = cells.find((c) => /src-1.*dest-a/.test(c.getAttribute('aria-label') ?? ''))!;
+		expect(first).toBeTruthy();
+		(first as HTMLElement).focus();
+		fireEvent.keyDown(first, { key: 'End' });
+		const rowEnd = cells.find((c) => /src-1.*dest-c/.test(c.getAttribute('aria-label') ?? ''))!;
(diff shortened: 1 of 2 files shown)


kriszyp (approved):
Nice, clean a11y fix. The render-time clamp is correct: `Math.min(active[0], rows.length - 1)` caps at the true last index (no off-by-one), passes through unchanged on grid growth, and the empty-grid `-1` is provably inert since the clamped indices are only ever compared (never used to index an array), an empty grid renders no gridcell to match, and `focusCell` guards its ref lookup. Deriving `isActive` from the clamped values guarantees exactly one surviving cell keeps `tabIndex=0`, so keyboard users can always re-enter the grid.

The regression test genuinely exercises the shrink path — it moves focus to the corner `[1,2]`, re-renders as 1x2, and asserts exactly one tabbable cell that it's the clamped last-surviving cell (src-1 -> dest-b), plus that arrow-nav still works from it. Verifies direction, not just non-crash.

One non-blocking suggestion: `Math.max(0, Math.min(...))` would normalize the empty-grid case to `0` and make the derived indices unconditionally valid array positions — reads more obviously safe to a future editor. Purely optional; the current `-1` is safe as written.

LGTM.

Issue #1447, opened 2026-07-10
Collapse the ~14 near-identical metric spec files with factory functions


kylebernhardy:
Of the ~25 per-metric files under `analytics/pipeline/`, 10 are pure `DimensionSelectorRenderer` wrappers (db-read, db-write, db-message, duration, transfer, cache-resolution, cache-hit, cpu-usage, success, response-200) and 4 are pure `TrafficByTypeRenderer` wrappers (bytes-sent, bytes-received, connections, database-size) — byte-identical except 3–8 strings (`diff db-read.tsx db-write.tsx` differs in 8 of 37 lines, all title/label strings). The mqtt-traffic-sent/received derived pair collapses the same way.

Suggested shape: `makeDimensionSelectorMetric({ id, title, description, label, ariaLabel, thresholds?, ... })` and `makeTrafficByTypeMetric(...)` fed by a config table. Removes ~450 lines and makes "add a metric" a table row instead of a new file.

Include in the same pass: the local `RendererProps` interface is redeclared in 17 files while `SpecRegistryRendererProps` (`types/analytics.ts:290`) already exists — and they've drifted (several omit `fillParent`). The factory makes this fall out for free.

Filed from a Status tab code review.

Comment generated by kAIle (Claude Fable 5)

PR #1483, opened 2026-07-13
refactor(status): generate wrapper metric specs from factory tables

kylebernhardy:
Implements [#1447 — Collapse the ~14 near-identical metric spec files with factory functions](https://github.com/HarperFast/studio/issues/1447).

## What collapsed

- **New `pipeline/wrapperMetrics.tsx`** — factory functions + one declarative table for all 14 pure-wrapper metrics:
  - `makeDimensionSelectorMetric({ ariaLabel, ...spec })` — the 10 `DimensionSelectorRenderer` wrappers (db-read, db-write, db-message, duration, transfer, cache-resolution, cache-hit, cpu-usage, success, response_200).
  - `makeTrafficByTypeMetric({ typeField, defaultViewMode?, ...spec })` — the 4 `TrafficByTypeRenderer` wrappers (bytes-sent, bytes-received, connections, database-size). `defaultViewMode: 'aggregate'` reproduces database-size's `viewMode={props.viewMode ?? 'aggregate'}`.
  - Two sub-factories for the families that were identical modulo strings: `makeLatencyQuantileMetric` (the six per-path p95/quantile-selector line charts — db-read/write/message, duration, transfer, cache-resolution) and `makeByteRateMetric` (bytes-sent/received).
- **New `pipeline/derived/mqtt-traffic.tsx`** — `makeMqttTrafficDerived(...)` builds the sent/received derived pair from one definition.
- **`pipeline/index.ts`** now registers wrapper metrics straight from the table — adding a pure-wrapper metric is a table entry plus one registry line, and a missing table key is a compile error.
- The 14 old file paths (plus `derived/mqtt-traffic-sent/received.tsx`) remain as 3–6-line re-exports, so every existing import path — including `tabs/ConnectionsPanel.tsx` and all pipeline tests — is untouched.
- The per-file `RendererProps` redeclaration is gone for these files; the factories use the canonical `SpecRegistryRendererProps` from `types/analytics.ts` (which restores the `fillParent` typing several copies had dropped). The four bespoke renderers (`connection`, `memory`, `main-thread-utilization`, `replication-latency`) are deliberately untouched — a parallel PR owns those files.

## Preservation guarantees & how they were verified

- **Registry keys unchanged** (including the intentionally snake_case `response_200`); spec contents, titles, descriptions, thresholds, aggregators, and renderer prop wiring (ariaLabel, typeField, spread order) are identical.
- **Existing pipeline tests pass unchanged** — the only change under `__tests__/` is one new file. They import each spec from its original path and pin per-metric pipeline behavior, so they double as the regression net for this refactor.
- **Deep-equal parity check** — during development, the origin/stage spec files were materialized side-by-side and every factory-built spec was asserted `toStrictEqual` against the old literal (15/15 passed; scaffolding not committed).
- **New `spec-registry.test.ts`** pins the full set of registry keys (spec + derived), that every entry has a complete spec, and which entries carry a Renderer — so a factory-table typo can't silently drop a metric.
- `tsc -b`, full `vitest run` (201 files / 1334 passed), `dprint fmt`, `oxlint` all clean.

## Net line delta

21 files: **+643 / −712 (net −69)**; excluding the new 85-line registry test, product source is net **−154**.

## Open concerns

None. Cross-model review: Codex's first pass raised one P2 (keep the received-side mqtt re-export file) — addressed. One intermediate Codex pass and one Gemini pass returned off-target output (they described files not in this diff) and were discounted and re-run; the final passes from both models found no drift and no correctness issues, matching the deep-equal parity check.


kriszyp (approved):
Nicely done — this is the kind of mechanical DRY refactor that's easy to get subtly wrong, and this one holds up. I verified all 15 wrapper specs and both MQTT-traffic derived specs are field-equivalent to the deleted literals (id/registry-key, title, description, unit, formatter, aggregator, threshold, confidence gate, renderer prop wiring all identical).

The correctness-sensitive bits from the #1472/#1475 line survive intact:
- `success` and `response_200` still compute the denominator via FieldExpr `total/count` (the `{ kind: 'op', op: '/', left: total, right: count }` expression) rather than Harper's optional `ratio` field, so the ~34% dropped-record bias stays fixed.
- `response_200` keeps its distinct `temporal: 'mean'` aggregator (vs `success`'s count-weighted-mean), with cross-node still count-weighted.
- confidence gating (`greyBelow: 40 / suppressBelow: 100`) and `fallbackMs: 60000` period-field bucketing preserved on every entry.

The `satisfies Record<string, WrapperMetric>` on the registry is a nice touch — a table-key typo becomes a compile error rather than a silently dropped metric, and the new spec-registry.test.ts pins the key set as a proper regression net.

One cosmetic, non-blocking note: for `bytes-sent`/`bytes-received`/`connections` the factory renderer emits `viewMode={props.viewMode ?? undefined}` where the originals passed no `viewMode` prop at all. Behaviorally identical in React (the child's `?? 'per-node'` default still applies), so no change needed — flagging only for completeness.

Approving.

PR #1501, opened 2026-07-14
feat(secrets): teach secrets.subscribe() and link docs in Secrets config guidance

dawsontoth:
Enhances the inline "How to read it in your component" guidance on the cluster **Secrets** config page (the add/edit dialogs) in two ways.

## 1. Practical `secrets.subscribe()` usage (harper#1787)

The **scoped-tier** example now demonstrates the live-change API: read the value immediately, then subscribe in a fire-and-forget background task so module load isn't blocked — with a guard so the first (current-value) yield doesn't trigger a redundant rebuild:

```js
import { secrets } from 'harper';

// Use the current value right away — read at your module's top level.
let value = secrets.STRIPE_KEY;

// Then react to rotations without blocking startup. Subscribing in a background
// task lets module load finish, so the rest of your app keeps initializing.
(async () => {
	for await (const next of secrets.subscribe('STRIPE_KEY')) {
		if (next === value) continue; // the first yield is the current value you already have
		value = next; // rotated — rebuild anything bound to the secret (e.g. an API client) here
	}
})();
```

The **global / `process.env`** tier example is unchanged — per harper#1787's tier semantics, `subscribe` there is reload-only (not live), so a hot-swap example would mislead. Name-aware bracket notation is preserved (`secrets['deploy.my-app']`).

## 2. Outward documentation links

- A contextual "Learn more about secrets in the Harper docs" link right under the code example (covers the config page and any delivery-enabled surface).
- A page-level "Secrets Docs" link in the Secrets config toolbar, matching the existing SSH-keys / certificates convention.

Both point to `https://docs.harperdb.io/reference/v5/security/secrets` (the current canonical docs structure). Plain `<a target="_blank" rel="noreferrer">` — the repo's dominant external-link pattern — rather than TanStack `Link`, so the isolated dialog test stays router-free.

## Docs

The docs page that backs the `subscribe()` example is **HarperFast/documentation#584** (stacked on the secrets-store docs PR #581). It predated harper#1787, so #584 adds the `secrets.subscribe()` section, the live-accessor note, and the per-tier change table.

## Verification

- `accessExample.test.ts` and `SecretsManager.delivery.test.tsx` updated; the delivery test renders the actual dialog DOM and asserts the subscribe line, the change-guard, and the doc-link href. **14/14 secrets tests pass**; `tsc --noEmit`, `dprint check`, and `oxlint` all clean.

> Note: committed unsigned — this environment has no TTY for gpg pinentry, so signing failed. Happy to amend with a signed commit if the branch protection needs it.


src/features/instance/secrets/accessExample.ts
```diff
@@ -46,16 +50,24 @@ export function buildSecretAccessExample(name: string, tier: SecretTier): string
...
+		`let value = ${read};`,
+		'',
+		'// Then react to rotations without blocking startup. Subscribing in a background',
+		'// task lets module load finish, so the rest of your app keeps initializing.',
+		'(async () => {',
+		`  for await (const next of secrets.subscribe(${quote(key)})) {`,
+		'    if (next === value) continue; // the first yield is the current value you already have',
+		'    value = next; // rotated — rebuild anything bound to the secret (e.g. an API client) here',
+		'  }',
+		'})();',
```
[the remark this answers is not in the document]
dawsontoth: Wrapped the subscribe loop in `try/catch` so a stream error can't become an unhandled rejection. Used `console.error` rather than an empty catch, so the failure is surfaced instead of silently swallowed (the last value keeps working until reload).

src/features/instance/config/secrets/index.tsx
```diff
@@ -141,7 +141,16 @@ export function ConfigSecretsIndex() {
 						)
 					);
 				}}
-			/>
+			>
+				<a
```
cb1kenobi: **Nit: toolbar docs link position differs from the SSH-keys/certificates convention it's meant to match**

`SecretsManager` renders `{children}` after the Refresh/Add buttons, so this "Secrets Docs" link lands at the far right of the toolbar. On the sshKeys and certificates config pages the docs `<Link>` is the first toolbar child, so it renders at the left, before Refresh/Add. Minor visual inconsistency with the stated convention; functionally fine.

—
_Generated by Barber AI_

dawsontoth: Good catch — added a dedicated `docsLink` slot to `SecretsManager` that renders before Refresh/Add and moved the "Secrets Docs" link into it, so it's now first/left like the sshKeys/certificates pages. The `children` slot (the .env editor's raw-editor button) still renders after Add. Added a regression test asserting the order.

Issue #1262, opened 2026-06-18
Instance Page /Browse v2


dawsontoth:
Stuff we've done in other PRs (worth sanity checking):
- [x] Upload CSV
- [x] Upload JSON
- [x] Export CSV
- [x] Hide and show columns
- [x] Letting the user resize column widths
- [x] UI vs API filtering and sorting toggle (the “only cached” button?)

Future improvements:

- [ ] Event subscription button (the “auto” switch? instead of polling, let's watch for changes)
- [ ] Export JSON (instead of just CSV)
- [ ] Altering schema (cross link to the schema in the Applications tab, if this table is defined by a schema file)
- [ ] Filtering dialog (by existing fields, and building complex queries)
- [ ] Hover over column to show the type, computed, relational properties

PR #1502, opened 2026-07-14
feat(databases): resizable browse-table columns + draggable sidebar

### LINKED ISSUE #1287: Allow resizing database table columns
- [ ] Allow resizing the columns

DavidCockerill:
Two related enhancements to the Databases browse view, one per commit (plus one review fix).

Closes #1287 and partially fulfills #1262 

## What

1. **Resizable table columns** — drag a column header's handle to resize; double-click the handle to auto-fit the column to its widest value (capped at 500px so a giant value can't blow it out); widths persist per table (sessionStorage). Columns hold their width via `table-fixed`, and a trailing filler column absorbs the leftover so rows still reach the edge instead of stretching the real columns.
2. **Draggable sidebar** — drag the sidebar's right edge to resize the DB/table tree pane; width clamps to `200px .. half the viewport` and persists (localStorage). Mirrors the existing applications file-tray resize pattern.

## Why

Make the Studio browse view more dynamic — resizable columns first, then a resizable sidebar so long table names aren't stuck truncated.

## Where to look

- **`ui/table.tsx`** is the bulk of commit 1: the resize handle, the double-click auto-fit (measured with a `Range` so it works even when the cell is wider than its content), and the drag-preview clamp. It's a shared component — see the review fix below.
- **Layout change** (commit 2): the databases page moved from a 12-col grid to flex so the sidebar can take an explicit, draggable width. The sticky, full-height sidebar behavior is preserved; worth confirming it feels right across widths and on mobile (handle is md+ only; mobile stays stacked full-width).
- **Default column widths** are per-type defaults; `id`/`__*time__` resolve to 200px because Harper reports them as `Any` — resize/auto-fit is the escape hatch rather than clever type sizing.

## Notes

- **Cross-model review done.** Codex flagged one real issue — because `TableHeadSortable` is shared with `SimpleBrowseDataTable` (roles / users / invoices / config / secrets), gating the handle on TanStack's `getCanResize()` (defaults enabled) put resize handles on *those* tables too. Fixed in `788edad7` by only rendering handles when the table opts in via `enableColumnResizing === true` (only `TableView` does). **Open caveat:** the Gemini (`agy`) leg failed to run (timeout, twice), so there was **no second-model perf/maintainability pass** — a manual skim of the drag handlers + auto-fit measurement would be a reasonable backstop.
- No docs change (Studio UI micro-interaction).
- Verified live (drag / auto-fit / persistence / clamp) + full unit suite green. Generated by an LLM — model: Claude Opus 4.8.

### DIFF
--- a/src/components/ui/table.tsx
+++ b/src/components/ui/table.tsx
@@ -145,7 +145,8 @@ export function TableHeadSortable<TData extends RowData>({
 			style={{ width: `${size}px`, maxWidth: `${size}px` }}
 			// relative (not overflow-hidden) so the absolute resize handle can render past the edge
 			// while dragging; the inner content div does the truncation instead.
-			className={cn('relative', enableSorting ? 'px-0' : 'px-2', className)}
+			// select-none on resizable headers so dragging the handle can't text-select the title.
+			className={cn('relative', enableResizing && 'select-none', enableSorting ? 'px-0' : 'px-2', className)}
 		>
 			{/* pr leaves room for the right-edge resize handle so the title never sits under it. */}
 			<div className={cn('flex items-center min-w-0 overflow-hidden', enableResizing && 'pr-4')}>
--- a/src/components/ui/table.tsx
+++ b/src/components/ui/table.tsx
@@ -5,6 +5,14 @@ import { ArrowDown, ArrowUp, ArrowUpDown, GripVerticalIcon } from 'lucide-react'
 import * as React from 'react';
 import { useCallback } from 'react';
 
+// Upper bound for double-click auto-fit so a very long value (e.g. a stringified object) can't
+// stretch a column across the whole viewport.
+const AUTO_FIT_MAX_SIZE = 500;
+// Horizontal cell padding (px-2 => 8px each side) added back after measuring bare content.
+const CELL_HORIZONTAL_PADDING = 16;
+// Width of the right-edge resize handle strip; reserved when auto-fitting so it never overlaps the title.
+const RESIZE_HANDLE_WIDTH = 16;
+
 export interface TableProps extends React.ComponentProps<'table'> {
 	containerClassName?: string;
 	containerRef?: React.Ref<HTMLDivElement>;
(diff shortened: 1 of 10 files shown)


src/components/ui/table.tsx
```diff
@@ -79,58 +87,105 @@ export function TableHeadSortable<TData extends RowData>({
...
-					>
-						<GripVerticalIcon />
-					</Button>
-				)}
+					: <span className="truncate">{content}</span>}
 			</div>
+			{enableResizing && (
+				<div
+					aria-hidden
+					className="absolute top-0 right-0 z-10 flex h-full w-3 cursor-col-resize items-center justify-center text-gray-600 hover:text-foreground"
```
dawsontoth: Two things on the resize handle here, both about discoverability (verified live on the preview in dark mode):

1. **Hit area is small** — the strip is `w-3` (~12px wide). For a drag affordance that's fiddly to grab. Bumping to ~`w-4`/`w-5` (and the header's `pr-3` reserve to match) would make it much easier to hit without changing the visible grip.

2. **Invisible at rest in dark mode** — the grip is a fixed `text-gray-600` that only appears via `hover:text-foreground`. `gray-600` reads fine on the light card but is near-invisible on the dark `black-dark` card, so you can't see the handle until you're already hovering it (chicken-and-egg for discoverability). Consider a theme-aware resting color (add a `dark:` variant) or an always-on subtle divider line that darkens on hover.

src/components/ui/table.tsx
```diff
@@ -79,58 +87,105 @@ export function TableHeadSortable<TData extends RowData>({
...
+					onMouseDown={header.getResizeHandler()} // for desktop
+					onTouchStart={header.getResizeHandler()} // for mobile
+					onDoubleClick={autoFitColumn}
+					style={{
+						transform: header.column.getIsResizing() ? `translateX(${previewOffset}px)` : '',
```
dawsontoth, line 186: Could we add a clearer visual indicator *during* the drag? With `columnResizeMode: 'onEnd'` the column width doesn't change until release, so mid-drag the only feedback is this ~12px low-contrast grip sliding via `translateX`. A full-height vertical guide line rendered while `header.column.getIsResizing()` is true (positioned at the same `previewOffset`) would show exactly where the new edge will land — and would pair naturally with the handle-visibility tweak above.

src/features/instance/databases/hooks/useResizableDatabasesSidebar.ts
```diff
@@ -0,0 +1,88 @@
...
+const MIN_SIDEBAR_WIDTH = 200;
+
+/** Clamp to a usable range: never below the minimum, never past half the viewport. */
+function clampWidth(width: number, viewportWidth: number): number {
+	const max = Math.max(MIN_SIDEBAR_WIDTH, Math.floor(viewportWidth / 2));
+	return Math.round(Math.min(Math.max(width, MIN_SIDEBAR_WIDTH), max));
+}
+
+/**
+ * Drag-to-resize state for the databases sidebar. Mirrors the applications file tray
```
dawsontoth, line 23: Non-blocking: this hook is ~90% identical to `useResizableSidebar` in `ApplicationsSidebar/SidebarResizeHandle.tsx` (same persisted-width / `isResizing` / clamp / mousemove-mouseup structure, and `clampWidth` mirrors `clampSidebarWidth`). The only real difference is the delta-tracking for an in-flow vs left-anchored panel. Might be worth extracting a shared `useResizablePanel({ anchored })` so the two don't drift. Fine to defer.

dawsontoth (commented):
Reviewed live on the preview at `788edad7` (worktree served on :5173). The core feature is solid — columns hold width via `table-fixed`, long values truncate cleanly, the trailing filler column absorbs the leftover, sizing persists per table, and the `Range`-based auto-fit is a nice touch. I also confirmed the clamp behavior: `getSize()` re-clamps to `minSize` (80) and `previewOffset` keeps the drag preview aligned, so the title never collides.

✅ **The `788edad7` scope fix works** — verified **0** resize grips on the org Roles page (`SimpleBrowseDataTable`) and **9** on the Product browse table. Gating on `enableColumnResizing === true && getCanResize()` is the right call.

_Optional, non-mandatory:_ if you ever want the `SimpleBrowseDataTable` tables resizable too, they'd need the same `table-fixed` + size-driven cell width/persistence plumbing — otherwise resizing there redistributes oddly and doesn't stick. Totally fine to leave as-is.

Inline notes below on the handle UX (hit area, dark-mode visibility, during-drag indicator) and one non-blocking DRY note on the sidebar hook.

A few more minor / non-blocking items with no single line to anchor:
- **A11y**: the table handle is `aria-hidden` + pointer-only, and the sidebar `role="separator"` has no keyboard handler — neither is keyboard-resizable. Matches the existing applications pattern, so not new.
- **Persist quirk (mirrors the existing pattern)**: on window resize the persisted width is permanently clamped down, so re-widening the window doesn't restore the earlier width; it also writes localStorage on every resize event. Same as `useResizableSidebar`, low priority.
- **Tests**: the new test only asserts the handle renders; the auto-fit DOM-measurement and persistence paths aren't covered (understandable — jsdom has no layout).

_Review assisted by Claude Code (Opus 4.8)._

src/components/ui/table.tsx
dawsontoth: light mode has the same issue, fwiw

src/features/instance/databases/hooks/useResizableDatabasesSidebar.ts
```diff
@@ -0,0 +1,88 @@
...
+			setWidth(clampWidth(persistedWidth, window.innerWidth));
+		}
+	}, [persistedWidth, isResizing]);
+
+	// If the viewport shrank since the width was stored, pull it back into range.
+	useEffect(() => {
+		const onResize = () => setPersistedWidth((current) => clampWidth(current, window.innerWidth));
+		window.addEventListener('resize', onResize);
+		return () => window.removeEventListener('resize', onResize);
+	}, [setPersistedWidth]);
```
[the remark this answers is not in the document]
DavidCockerill: Fixed in 174c4ca7 — the resize handler now re-clamps only the *rendered* width and never overwrites the persisted preference, so shrinking then re-widening the window restores it, and it no longer writes localStorage on every resize event. Also added a hook test for exactly this. Thanks! — Claude (Opus 4.8)

src/features/instance/databases/hooks/useResizableDatabasesSidebar.ts
```diff
@@ -0,0 +1,88 @@
...
+		window.addEventListener('mouseup', onMouseUp);
+		return () => {
+			document.body.style.userSelect = previousUserSelect;
+			window.removeEventListener('mousemove', onMouseMove);
+			window.removeEventListener('mouseup', onMouseUp);
+		};
+	}, [isResizing, setPersistedWidth]);
+
+	return { width, isResizing, startResizing };
+}
```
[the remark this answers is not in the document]
DavidCockerill: Added in 174c4ca7 — the hook now returns `handleKeyDown` (ArrowLeft/Right nudge the persisted width in 10px steps). — Claude (Opus 4.8)

src/features/instance/databases/index.tsx
```diff
@@ -25,6 +28,10 @@ export function Databases() {
 		getDescribeAllQueryOptions({ ...instanceParams, skipRecordCount: true }),
 	);
 
+	const { width: sidebarWidth, isResizing, startResizing } = useResizableDatabasesSidebar();
+	// Drive the width through a CSS variable so it only applies at md+ (mobile stays full-width, stacked).
+	const sidebarWidthVar = { '--db-sidebar-width': `${sidebarWidth}px` } as CSSProperties;
```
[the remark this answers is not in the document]
DavidCockerill: Done in 174c4ca7 — `handleKeyDown` is destructured and passed to the separator. — Claude (Opus 4.8)

src/features/instance/databases/index.tsx
```diff
@@ -39,11 +46,26 @@ export function Databases() {
...
+						role="separator"
+						aria-orientation="vertical"
+						aria-label="Resize sidebar"
+						onMouseDown={startResizing}
+						className={cn(
+							'hidden md:block absolute top-0 right-0 bottom-0 w-1 z-40 cursor-col-resize',
+							'hover:bg-violet-400/60 dark:hover:bg-violet-500/60 transition-colors',
+							isResizing && 'bg-violet-400/60 dark:bg-violet-500/60',
+						)}
+					/>
```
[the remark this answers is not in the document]
DavidCockerill: Done in 174c4ca7 — the separator now has `tabIndex={0}`, `aria-valuenow/min/max`, `onKeyDown`, and a `focus-visible` ring. — Claude (Opus 4.8)

src/components/ui/table.tsx
DavidCockerill: Great catch — fixed in 174c4ca7: bumped the hit area to `w-4` (header `pr-4` to match) and switched the resting color to `text-muted-foreground` so the grip is visible before hover in both light and dark. Verified live in dark mode. — Claude (Opus 4.8)

src/components/ui/table.tsx
DavidCockerill: Added in 174c4ca7 — a full-height guide line renders at the target edge while `getIsResizing()` is true (clamped to match the handle's own preview), and clears on release. Verified live. Nice suggestion. — Claude (Opus 4.8)

src/features/instance/databases/hooks/useResizableDatabasesSidebar.ts
DavidCockerill: Agreed this is worth doing. Deferring to its own follow-up since a shared `useResizablePanel({ anchored })` also touches the applications `useResizableSidebar` — leaving this thread open to track it rather than folding an applications-side refactor into this PR. — Claude (Opus 4.8)

DavidCockerill:
Pushed `174c4ca7` addressing the review — all inline threads resolved except the DRY-extraction note, which I left open to track as a follow-up.

- **Persist bug** (Gemini + Dawson): window resize now re-clamps only the *rendered* width, never the saved preference — shrinking then re-widening restores it, and it no longer writes localStorage on every resize event.
- **Handle discoverability** (Dawson): wider `w-4` hit area (header `pr-4` to match) + `text-muted-foreground` resting color, so the grip is visible before hover in both light and dark.
- **During-drag guide** (Dawson): a full-height guide line renders at the target edge while resizing and clears on release.
- **Sidebar a11y** (Gemini): the separator is keyboard-resizable (Arrow keys, 10px steps) with `tabIndex`, `aria-valuenow/min/max`, and a focus-visible ring.
- **Tests**: added hook unit tests — clamp bounds, drag-persist, resize-without-clobber (the fix above), and Arrow-key nudging.

Verified live on the preview: grip visible at rest, guide renders and clears, keyboard resize persists. Re-requesting review — thanks both!

_(Generated by Claude Code — Opus 4.8.)_

src/features/instance/databases/hooks/useResizableDatabasesSidebar.ts
DavidCockerill: Tracked as a separate follow-up (shared `useResizablePanel({ anchored })` spanning this hook + the applications `useResizableSidebar`). Resolving here since it's non-blocking and captured for its own PR — thanks for the pointer! — Claude (Opus 4.8)

dawsontoth:
Pushed two small follow-ups (`d43f5935`, `d692f954`) — a couple of things we caught while poking at the branch in the live preview:

**`d43f5935` — visible resting edge on the sidebar.** The resize handle only shows its highlight on hover/focus/drag. That mirrors the applications file tray — but that tray's pane already has a resting `border-r`, so its edge is always visible; the databases sidebar had no border, so there was nothing to see at rest (you had to hunt for the handle). Added `md:border-r border-border` — the same divider token the browse table already uses — so the seam is visible in both light and dark, with the violet hover/drag highlight still layering on top.

**`d692f954` — no text-selection while resizing a column.** Mousedown-drag on a column's resize handle would text-select the header title (e.g. selecting `name`). TanStack's `getResizeHandler` doesn't suppress selection the way the sidebar hook does, so added `select-none` to resizable headers. Gated on `enableResizing`, so only the browse table is affected.

Both verified live (light + dark; drag no longer selects the label). Format/lint/`TableView` test pass locally.

_(Pushed by Dawson, paired with Claude Code — Opus 4.8.)_

dawsontoth (approved):

PR #1523, opened 2026-07-15
fix(auth): sign out and redirect on 401 from the CM API

DavidCockerill:
> **Draft** — companion to the central-manager fabric-admin session work (`fabric-admin-auth-hardening`, adds a `FABRIC_ADMIN_SESSION_MAX_AGE_HOURS` cap). Keeping this in draft until that lands. Useful on its own, though — it fixes the client reaction to *any* lost session.

## Problem
The CM `apiClient` ([apiClient.ts](src/config/apiClient.ts)) has no response interceptor, and the dashboard guard ([dashboardRoute.ts](src/router/dashboardRoute.ts)) only redirects when the cached `authStore` user is already `null`. So when a cloud session is lost mid-use — normal cookie expiry, server-side revocation, or a fabric admin hitting the new max-age cap — the SPA keeps rendering off the **stale cached user** while every data call fails with `Not allowed` errors. The user isn't bounced to sign-in until they manually refresh (which re-runs `getCurrentUser`, gets 401, and empties the store).

Verified live on dev: after a fabric-admin session was invalidated server-side, the UI kept navigating (cached `authStore`) and only redirected on a hard refresh.

## Fix
Add a response interceptor on `apiClient`: on **401**, clear the cached cloud auth (`authStore.setUserForEntity(OverallAppSignIn, null)`). That triggers the existing `AppRouted` `router.invalidate()` effect → guards re-run → redirect to `/sign-in`. No imperative router access needed.

**Only 401** triggers it. A dead session yields 401 from `GET /User/current` (its `allowRead` returns false when there's no user). **403 is left alone** — it's a legitimate "authenticated but not permitted" response and must not log the user out.

## Where to look
- `src/lib/unauthorizedResponseHandler.ts` — the 401 detection, factored app-import-free so it unit-tests in node env (importing `authStore` pulls in `localStorage` at module load).
- `src/lib/installApiUnauthorizedRedirect.ts` — wires it to `apiClient` with the authStore-clearing callback.
- `src/main.tsx` — installed at startup alongside the other `install*` guards.

## Testing
- `vitest run src/lib/unauthorizedResponseHandler.test.ts` — 3 passing (401 clears + rejects; 403 doesn't; network error doesn't).
- Not yet covered: the full interceptor→guard→redirect path in a running app. Worth a manual check once merged with the CM cap (sign in, let the session expire, confirm auto-redirect without a manual refresh).

## Open question for reviewers
403s from a dead session (admin endpoints return `Not allowed`/403, not 401) won't trigger this. Keying on 401 only is the safe choice (avoids logging out on genuine permission denials), but it means a data call that only ever 403s won't itself force logout — the redirect comes from the next 401-returning call (e.g. `getCurrentUser` on focus/refetch). Flagging in case we want 403-on-`/User/current` handling too.

_Generated with the assistance of an LLM (Claude Opus 4.8)._

### DIFF
--- a/src/lib/installApiUnauthorizedRedirect.ts
+++ b/src/lib/installApiUnauthorizedRedirect.ts
@@ -2,14 +2,35 @@ import { apiClient } from '@/config/apiClient';
 import { authStore, OverallAppSignIn } from '@/features/auth/store/authStore';
 import { makeUnauthorizedResponseHandler } from '@/lib/unauthorizedResponseHandler';
 
+// The unauthenticated auth flows 401 for a bad credential or a stale
+// reset/verify token (CM's verifyToken), not for a lost session — a stale
+// email link opened in a second tab must not sign out a live session.
+const AUTH_FLOW_PATHS = [
+	'/Login/',
+	'/ForgotPassword/',
+	'/ResetPassword/',
+	'/VerifyEmail/',
+	'/ResendVerificationEmail/',
+];
+
+let installed = false;
+
 /**
  * Install the 401 -> clear-auth interceptor on the CM api client. Call once at
- * app startup, before React mounts. On a lost session the cached cloud user is
+ * app startup, before React mounts; re-installs (HMR) are no-ops so duplicate
+ * interceptors can't stack. On a lost session the cached cloud user is
  * cleared, which re-runs the route guards and redirects to /sign-in.
  */
 export function installApiUnauthorizedRedirect(): void {
+	if (installed) {
+		return;
+	}
+	installed = true;
 	apiClient.interceptors.response.use(
 		(response) => response,
-		makeUnauthorizedResponseHandler(() => authStore.setUserForEntity(OverallAppSignIn, null)),
+		makeUnauthorizedResponseHandler(
+			() => authStore.setUserForEntity(OverallAppSignIn, null),
+			(url) => AUTH_FLOW_PATHS.some((path) => url.startsWith(path)),
+		),
 	);
 }
--- a/src/lib/installApiUnauthorizedRedirect.ts
+++ b/src/lib/installApiUnauthorizedRedirect.ts
@@ -1,0 +1,15 @@
+import { apiClient } from '@/config/apiClient';
+import { authStore, OverallAppSignIn } from '@/features/auth/store/authStore';
+import { makeUnauthorizedResponseHandler } from '@/lib/unauthorizedResponseHandler';
+
+/**
+ * Install the 401 -> clear-auth interceptor on the CM api client. Call once at
+ * app startup, before React mounts. On a lost session the cached cloud user is
+ * cleared, which re-runs the route guards and redirects to /sign-in.
+ */
+export function installApiUnauthorizedRedirect(): void {
+	apiClient.interceptors.response.use(
+		(response) => response,
+		makeUnauthorizedResponseHandler(() => authStore.setUserForEntity(OverallAppSignIn, null)),
+	);
+}
--- a/src/lib/installApiUnauthorizedRedirect.ts
+++ b/src/lib/installApiUnauthorizedRedirect.ts
@@ -1,4 +1,5 @@
 import { apiClient } from '@/config/apiClient';
+import { clearAuthStateLocally } from '@/features/auth/handlers/clearAuthStateLocally';
 import { authStore, OverallAppSignIn } from '@/features/auth/store/authStore';
 import { makeUnauthorizedResponseHandler } from '@/lib/unauthorizedResponseHandler';
 
@@ -13,24 +14,46 @@ const AUTH_FLOW_PATHS = [
 	'/ResendVerificationEmail/',
 ];
 
+// Config key stamped by the request interceptor with the cloud identity at send
+// time, so the response handler can tell whether the session has changed since.
+const AUTH_STAMP = 'authUserAtSend';
+
+// The cloud user's stable id, or null when signed out. LocalUser has no id.
+function currentCloudUserId(): string | null {
+	const user = authStore.getConnectionById(OverallAppSignIn).user;
+	return user && 'id' in user ? user.id : null;
+}
+
 let installed = false;
 
 /**
  * Install the 401 -> clear-auth interceptor on the CM api client. Call once at
  * app startup, before React mounts; re-installs (HMR) are no-ops so duplicate
- * interceptors can't stack. On a lost session the cached cloud user is
- * cleared, which re-runs the route guards and redirects to /sign-in.
+ * interceptors can't stack. On a lost session the cached auth is fully cleared
+ * (locally), which re-runs the route guards and redirects to /sign-in.
  */
 export function installApiUnauthorizedRedirect(): void {
 	if (installed) {
 		return;
 	}
 	installed = true;
+
+	// Stamp each request with the cloud identity at send time (see AUTH_STAMP).
+	apiClient.interceptors.request.use((config) => {
+		(config as typeof config & { [AUTH_STAMP]?: string | null })[AUTH_STAMP] = currentCloudUserId();
+		return config;
+	});
+
 	apiClient.interceptors.response.use(
 		(response) => response,
 		makeUnauthorizedResponseHandler(
-			() => authStore.setUserForEntity(OverallAppSignIn, null),
+			clearAuthStateLocally,
 			(url) => AUTH_FLOW_PATHS.some((path) => url.startsWith(path)),
+			// Only clear if the identity hasn't changed since the request was sent —
+			// don't let a slow 401 from the old session clear a fresh re-login.
+			(config) =>
+				(config as (typeof config & { [AUTH_STAMP]?: string | null }) | undefined)?.[AUTH_STAMP]
+					=== currentCloudUserId(),
 		),
 	);
 }
--- a/src/lib/unauthorizedResponseHandler.test.ts
+++ b/src/lib/unauthorizedResponseHandler.test.ts
@@ -1,0 +1,29 @@
+import type { AxiosError } from 'axios';
+import { describe, expect, it, vi } from 'vitest';
+import { makeUnauthorizedResponseHandler } from './unauthorizedResponseHandler';
+
+const errorWithStatus = (status?: number) =>
+	({ response: status === undefined ? undefined : { status } }) as AxiosError;
+
+describe('makeUnauthorizedResponseHandler', () => {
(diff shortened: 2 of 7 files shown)


dawsontoth:
sounds good to me, lemme know when you're out of draft :D

DavidCockerill:
@dawsontoth once my CM work relating to this is done i'll move it out of draft. Just in case i need to modify anything

src/lib/unauthorizedResponseHandler.ts
```diff
@@ -0,0 +1,26 @@
...
+ * default node env without pulling in browser-only globals.
+ */
+export function makeUnauthorizedResponseHandler(clearAuth: () => void) {
+	return (error: AxiosError): Promise<never> => {
+		if (error.response?.status === 401) {
+			clearAuth();
+		}
+		// Re-reject so individual callers still see (and can handle) the error.
+		return Promise.reject(error);
+	};
```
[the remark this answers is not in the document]
DavidCockerill: Applied in ba6a7f88 — good defensive catch. One deviation from the suggestion: kept the `AxiosError` typing with `error?.response?.status` rather than widening the parameter to `any`, which would drop type safety for no extra robustness. (Today this handler is the only response interceptor on `apiClient`, and axios itself always rejects with an `AxiosError`, so the guard is future-proofing for interceptors added later.)

— Claude Opus 4.8 (on behalf of @DavidCockerill)

src/lib/unauthorizedResponseHandler.test.ts
```diff
@@ -0,0 +1,29 @@
...
+		await expect(makeUnauthorizedResponseHandler(clearAuth)(error)).rejects.toBe(error);
+		expect(clearAuth).not.toHaveBeenCalled();
+	});
+
+	it('does not clear auth on a network error with no response', async () => {
+		const clearAuth = vi.fn();
+		const error = errorWithStatus(undefined);
+		await expect(makeUnauthorizedResponseHandler(clearAuth)(error)).rejects.toBe(error);
+		expect(clearAuth).not.toHaveBeenCalled();
+	});
```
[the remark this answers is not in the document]
DavidCockerill: Partially applied in ba6a7f88. Note the first suggested test ("does not clear auth on a network error with no response") already exists in this file — see the third `it` block. Added the second one (undefined rejection reason passes through without throwing), which became meaningful once the guard landed.

— Claude Opus 4.8 (on behalf of @DavidCockerill)

src/lib/unauthorizedResponseHandler.ts
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * Kept free of app imports (auth store, api client) so it unit-tests in the
+ * default node env without pulling in browser-only globals.
+ */
+export function makeUnauthorizedResponseHandler(clearAuth: () => void) {
+	return (error: AxiosError): Promise<never> => {
+		// `error?.`: axios itself always rejects with an AxiosError, but an upstream
+		// interceptor added later could reject with anything (even undefined) — don't
+		// let this handler replace the rejection reason with its own TypeError.
+		if (error?.response?.status === 401) {
```
kriszyp: Question, not a blocker: `apiClient` also carries `/Login/`, `/ResetPassword/`, `/VerifyEmail/`, and `/ForgotPassword/` — none of which represent a lost CM session. For a signed-out visitor `clearAuth()` is a no-op today (already null), but if the CM API ever models an invalid/expired reset or verify-email token as `401` (rather than `400`/`410`), a user who's signed in in one tab and opens a stale verify/reset link in another tab would get signed out of their live session as a side effect. Could you confirm those endpoints don't 401 for "bad token" cases? If they can, might be worth scoping this interceptor away from the unauthenticated auth-flow requests.

src/lib/installApiUnauthorizedRedirect.ts
```diff
@@ -0,0 +1,15 @@
...
+import { authStore, OverallAppSignIn } from '@/features/auth/store/authStore';
+import { makeUnauthorizedResponseHandler } from '@/lib/unauthorizedResponseHandler';
+
+/**
+ * Install the 401 -> clear-auth interceptor on the CM api client. Call once at
+ * app startup, before React mounts. On a lost session the cached cloud user is
+ * cleared, which re-runs the route guards and redirects to /sign-in.
+ */
+export function installApiUnauthorizedRedirect(): void {
+	apiClient.interceptors.response.use(
```
kriszyp, line 47: Nit: the other `install*.ts` helpers in this directory (`installBrowserTranslationDomGuard`, `installStaleDeployReload`) explicitly guard/document being safe to call more than once (HMR, accidental re-invocation). This one registers the interceptor unconditionally, so a second call would stack a duplicate (harmless today since `main.tsx` calls it once and `clearAuth()` is idempotent, but worth matching the sibling convention for consistency).

kriszyp (approved):
Nice

dawsontoth (approved):
I like Kris' suggestions, mind folding those in to this PR?

src/lib/unauthorizedResponseHandler.ts
DavidCockerill: Great catch — and it's not hypothetical: CM's `verifyToken` (sharedHelpers.js) throws `ERROR_CODE.UNAUTHORIZED` for a bad/expired token, and both `ResetPassword` and `VerifyEmail` route through it — plus `/Login` 401s on bad credentials. So a stale verify/reset link in a second tab **would** have signed out a live session today. Fixed in 70fb4dcf: the handler now takes an exemption predicate and the installer scopes it away from the unauthenticated auth flows (`/Login/`, `/ForgotPassword/`, `/ResetPassword/`, `/VerifyEmail/`, `/ResendVerificationEmail/`), with tests for exempt/non-exempt/missing-config cases.

— Claude Opus 4.8 (on behalf of @DavidCockerill)

src/lib/installApiUnauthorizedRedirect.ts
DavidCockerill: Applied in 70fb4dcf — module-level `installed` flag, matching the sibling `install*` convention; re-invocations are no-ops.

— Claude Opus 4.8 (on behalf of @DavidCockerill)

DavidCockerill:
@dawsontoth folded in (70fb4dcf) — Kris's first point turned out to be a live bug, not just a question: CM 401s stale reset/verify-email tokens, so a stale link in another tab would have signed out an active session. The interceptor is now scoped away from the unauthenticated auth flows, and install is idempotent. Still holding draft until the related CM session work (central-manager#459) lands.

— Claude Opus 4.8 (on behalf of @DavidCockerill)

src/lib/installApiUnauthorizedRedirect.ts
```diff
@@ -0,0 +1,36 @@
...
+ */
+export function installApiUnauthorizedRedirect(): void {
+	if (installed) {
+		return;
+	}
+	installed = true;
+	apiClient.interceptors.response.use(
+		(response) => response,
+		makeUnauthorizedResponseHandler(
+			() => authStore.setUserForEntity(OverallAppSignIn, null),
```
kriszyp: This clears only the cloud-user slot, so it is not equivalent to the existing logout path. After A's CM session expires, the in-memory entity connections/Fabric tokens, persisted basic/Fabric flags, and React Query cache remain; if B signs in without a reload, `instanceLayoutRoute` can trust A's still-populated entity connection before checking B's cloud permissions, and queries can reuse A's cached data. Could we centralize a local-only full-sign-out cleanup (all auth connections/tokens/flags plus user-scoped query/session data, without retrying the failed CM logout) and use it here? An A → 401 → B test should assert those stores are empty before B's routes load.

src/lib/unauthorizedResponseHandler.ts
```diff
@@ -0,0 +1,34 @@
...
+ */
+export function makeUnauthorizedResponseHandler(
+	clearAuth: () => void,
+	isExemptUrl: (url: string) => boolean = () => false,
+) {
+	return (error: AxiosError): Promise<never> => {
+		// `error?.`: axios itself always rejects with an AxiosError, but an upstream
+		// interceptor added later could reject with anything (even undefined) — don't
+		// let this handler replace the rejection reason with its own TypeError.
+		if (error?.response?.status === 401 && !isExemptUrl(error.config?.url ?? '')) {
```
kriszyp: Could we guard this with the auth generation that was current when the request was sent? A normal expiry sequence can otherwise clear a fresh login: requests A/B leave with the expired cookie, A's 401 redirects to sign-in, `/Login/` succeeds and stores the new user, then slower B returns its old 401 and this unconditional handler clears the new session. Stamping the request config in a request interceptor (generation or user identity) and only clearing when it still matches would close the race; a deferred-response test resolving B after simulated login would cover it.

kriszyp (commented):
Looks good, maybe a race condition to address.

src/lib/installApiUnauthorizedRedirect.ts
DavidCockerill: Fixed in a1fd078d. You're right the partial clear wasn't equivalent to logout. Added `authStore.signOutAllLocally()` — a local-only teardown that reuses the existing per-entity `signOutLocally()` for every entity in `potentiallyAuthenticated` (clears the in-memory connection, the stored basic-auth creds, the Fabric Connect flag **and** its in-memory token), then clears the residual maps (`fabricConnectAuth`/`fabricConnectInFlight`/`operationTokenRefreshInFlight`) and the cloud slot — no per-instance or CM network logout. The 401 path now calls a `clearAuthStateLocally()` handler that runs that plus **both** React Query caches (Query + Mutation — logout only cleared QueryCache) and storage. A test sets up A's cloud + instance connections with a Fabric flag and asserts they're all null/false after sign-out, so B can't inherit them.

— Claude Opus 4.8 (on behalf of @DavidCockerill)

src/lib/unauthorizedResponseHandler.ts
DavidCockerill: Fixed in a1fd078d, via identity stamping as you suggested. A request interceptor now stamps the cloud user id at send time onto the request config; the handler takes an `isStillCurrent(config)` guard and only clears when the stamped identity equals the current one. So the sequence you described — A expires → 401 clears + redirect → re-login as B → B's earlier in-flight request returns its stale 401 — no longer clears B: at response time the current identity is B but the stale request was stamped A, so it's ignored. Two tests cover it: identity-unchanged (normal expiry → clears) and identity-changed-after-relogin (stale 401 → does not clear).

— Claude Opus 4.8 (on behalf of @DavidCockerill)

PR #1529, opened 2026-07-16
fix(browse): handle rows with an undefined primary key (#1199)

### LINKED ISSUE #1199: 500 'hash_values' must be strings or numbers
it's somehow possible to have a `undefined` primary key. When I click a row (https://github.com/HarperFast/studio/blob/431cf712c96f26de145ac3317f45da7176e61cce/src/features/instance/databases/components/DatabaseTableView.tsx#L303) with an `undefined` key, it assembles a POST body (https://github.com/HarperFast/studio/blob/stage/src/integrations/api/instance/database/getSearchById.ts#L19) with `[undefined]` which after being JSON.stringified becomes `[null]`.

It seems `onRowClick` should sanity check the record has a defined primary key and if not, show a nicer error message.

I wonder if it even needs to re-fetch the record?

If the primary key is `undefined`, you can't edit or delete the record. The only option currently is to drop and recreate the table, which is not ideal. I wonder if there's a way to delete or update a record by the result set index?

dawsontoth:
Closes #1199.

## The state (yes, it's reproducible)

A record can end up with **no value for its declared primary key**. The path: change a table's `@primaryKey` on a *populated* table (or formalize a previously-dynamic table with a natural key, e.g. `id: ID` → `email: String @primaryKey`, then redeploy). Harper accepts the change but does **not** re-key the existing rows — they keep their original stored key (the old `id`), and `describe` now reports the new primary-key attribute the old rows have no value for.

So `rowData.original[primaryKey]` is `undefined`. Clicking the row did:

```
setSelectedIds([undefined]) → search_by_id { ids: [null] }   // JSON.stringify([undefined]) === "[null]"
→ 500 'hash_values' must be strings or numbers
```

…and the edit modal just spun on `<Loading />` forever (the query has `retry: false` and no error toast).

## What this PR does

- **`onRowClick`** detects the missing primary key, skips the doomed `search_by_id`, and opens the modal on the row the list query already returned (answering the issue's "does it even need to re-fetch?" — no).
- **`EditTableRowModal`** shows a warning banner ("This row has no primary key value… usually means the table's primary key was changed after the row was created"), renders the row **read-only**, and hides the Save/Delete actions.
- **`getSearchByIdOptions`** filters `null`/`undefined` ids so a null hash value is never sent (belt-and-suspenders for any caller).
- The Delete guard now uses `!= null` instead of a truthy check, so a record with a falsy-but-valid key (`0`, `""`) can still be deleted.

Unit test added for the id filtering; full suite (1688 tests) green.

## The scarier thing I found

While tracing the delete path I found that a Harper `delete` with `hash_values: [null]` **deletes the entire table** while reporting `"1 of 1 record successfully deleted"` (verified on 4.7.23 and 5.0.15). Studio's delete button happens to guard against it, but that's core data-loss — fixed here: **HarperFast/harper#1837**.

Full investigation and reproduction steps: #1199.

### DIFF
--- a/src/features/instance/databases/modals/EditTableRowModal.tsx
+++ b/src/features/instance/databases/modals/EditTableRowModal.tsx
@@ -1,9 +1,10 @@
 import { Loading } from '@/components/Loading';
+import { Alert, AlertDescription, AlertTitle } from '@/components/ui/alert';
 import { Button } from '@/components/ui/button';
 import { Dialog, DialogContent, DialogFooter, DialogHeader, DialogTitle } from '@/components/ui/dialog';
 import { useMonacoTheme } from '@/hooks/useMonacoTheme';
 import { Editor } from '@/lib/monaco/MonacoEditor';
-import { Save, Trash } from 'lucide-react';
+import { Save, Trash, TriangleAlert } from 'lucide-react';
 import { useCallback, useMemo, useState } from 'react';
 
 export function EditTableRowModal({
@@ -12,6 +13,7 @@ export function EditTableRowModal({
 	setIsModalOpen,
 	isModalOpen,
 	primaryKey,
+	missingPrimaryKey,
 	syntheticAttributes,
 	data,
 	onSaveChanges,
@@ -24,6 +26,9 @@ export function EditTableRowModal({
 	setIsModalOpen: (open: boolean) => void;
 	isModalOpen: boolean;
 	primaryKey: string;
+	/** The clicked row has no value for the declared primary key, so it can't be looked up, edited,
+	 * or deleted by id (see #1199). We still show its contents read-only, with an explanation. */
+	missingPrimaryKey?: boolean;
 	/** Relationship/computed attribute names — read-only, so they are hidden from the editable JSON
 	 * (saving a record that assigns one fails, even with null). */
 	syntheticAttributes?: string[];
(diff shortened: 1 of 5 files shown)


src/features/instance/databases/modals/EditTableRowModal.tsx
```diff
@@ -68,8 +76,30 @@ export function EditTableRowModal({
 					: undefined}
 			>
 				<DialogHeader>
-					<DialogTitle>{canEditRecords ? 'Edit' : 'View'} Row</DialogTitle>
+					<DialogTitle>{isReadOnly ? 'View' : 'Edit'} Row</DialogTitle>
 				</DialogHeader>
+				{missingPrimaryKey && (
```
cb1kenobi: **Low: core missing-primary-key UI behavior is untested**

The added unit test covers `getSearchByIdOptions` id-filtering, but the PR's central user-facing change has no test: that a missing-PK row opens the modal read-only, shows this warning banner, and hides Save/Delete (`isReadOnly` / `!missingPrimaryKey` gating here and in `DatabaseTableView.onRowClick`). The repo already has ~33 `*.test.tsx` component tests, so there's precedent to render `EditTableRowModal` with `missingPrimaryKey` and assert the editor is read-only and the Save/Delete buttons are absent — the branch most likely to regress silently.

—
_Generated by Barber AI_

dawsontoth: Done in e589c41 — added `EditTableRowModal.test.tsx` (jsdom) that renders the modal with `missingPrimaryKey` and asserts the editor is read-only, the banner names the missing PK (`email`), and Save/Delete are absent, plus a normal-row case that keeps both actions. Monaco is stubbed since it can't load in jsdom.

PR #1532, opened 2026-07-17
fix(clusters): hide overlapping version tags in the version picker

### LINKED ISSUE #1531: Hide Overlapping Version Tags
If in the version editing experience for clusters we get 2 versions that are identical but have different tags, don't show both of them to the user.

For example, if we get `5.1.21` stable and `5.1.21` next... only show `stable`. So rough order of the tags I'm aware of, most preferred to least:

1. stable
2. next
3. beta
4. alpha

Those should be preferred over any unknown values that come in, if we need to decide "should I show...stable or foobarbaz?"

dawsontoth:
## What & why

Closes #1531.

The `HarperVersions` endpoint can return the same version under more than one release tag — e.g. `5.1.21` tagged both `stable` and `next`. The cluster version picker ([`ClusterVersion.tsx`](src/features/clusters/upsert/fields/ClusterVersion.tsx)) mapped the raw list straight through, so it offered the same version twice and, because it uses the version string as the React `key`, the duplicates also collided on their keys.

Per the issue, when two entries are identical but carry different tags we should surface only the most-preferred one.

## Change

Dedupe in [`getHarperVersionsQuery.ts`](src/features/clusters/queries/getHarperVersionsQuery.ts) itself, so **both** the create and edit flows are covered (the edit-only `useMemo` in `upsert/index.tsx` does its own version filtering but is skipped entirely when creating a new cluster).

- New pure helper `dedupeHarperVersionsByTag` collapses entries sharing a version string to a single entry.
- Tag preference, most- to least-preferred: **stable > next > beta > alpha**, and any of those beats an unrecognized tag (`HARPER_VERSION_TAG_PREFERENCE`).
- Kept entries retain their original position in the list (backed by `Map` insertion order).
- Widened the `HarperVersion.name` type from the incomplete `'stable' | 'next' | 'current'` union to `string`, reflecting that the endpoint also emits `beta`/`alpha` and potentially unknown tags.

## Testing

- New unit tests in `getHarperVersionsQuery.test.ts` cover the exact issue example (`5.1.21` stable+next → only stable), the full preference order, known-beats-unknown, a lone unknown tag, order preservation, and the empty case.
- Full suite green (1727 passed), plus `tsc -b`, `oxlint`, and `dprint`.

> Note: the live endpoint doesn't reliably emit overlapping tags on demand, so this was verified via the pure-function unit tests rather than the browser flow; the picker's rendering is unchanged.

### DIFF
--- a/src/features/clusters/queries/getHarperVersionsQuery.ts
+++ b/src/features/clusters/queries/getHarperVersionsQuery.ts
@@ -51,7 +51,7 @@ async function getHarperVersions() {
 	const response = data as HarperVersionsResponse;
 	return {
 		...response,
-		value: dedupeHarperVersionsByTag(response?.value ?? []),
+		value: dedupeHarperVersionsByTag(response.value),
 	} satisfies HarperVersionsResponse;
 }
 
--- a/src/features/clusters/queries/getHarperVersionsQuery.ts
+++ b/src/features/clusters/queries/getHarperVersionsQuery.ts
@@ -7,15 +7,52 @@ export interface HarperVersionsResponse {
 	value: HarperVersion[];
 }
 
-interface HarperVersion {
-	name: 'stable' | 'next' | 'current';
+export interface HarperVersion {
+	/**
+	 * Release tag for this version. `stable`, `next`, `beta`, and `alpha` come from the endpoint (which may
+	 * also return tags we don't know about yet, hence the plain `string`); `current` is synthesized
+	 * client-side for the version a cluster already runs.
+	 */
+	name: string;
 	version: string;
+}
+
+/**
+ * Known release tags, most-preferred first. When the endpoint returns the same version under more than one
+ * tag we keep the most-preferred; any unrecognized tag ranks below all of these.
+ */
+export const HARPER_VERSION_TAG_PREFERENCE = ['stable', 'next', 'beta', 'alpha'] as const;
+
+function tagRank(name: string): number {
+	const index = (HARPER_VERSION_TAG_PREFERENCE as readonly string[]).indexOf(name);
+	return index === -1 ? HARPER_VERSION_TAG_PREFERENCE.length : index;
+}
+
+/**
+ * Collapse versions that share the same version string down to a single entry, keeping the one with the
+ * most-preferred tag (stable > next > beta > alpha > anything else). The endpoint can return, e.g.,
+ * `5.1.21` tagged both `stable` and `next`; the picker should only offer `stable`.
+ */
+export function dedupeHarperVersionsByTag(versions: HarperVersion[]): HarperVersion[] {
+	const bestByVersion = new Map<string, HarperVersion>();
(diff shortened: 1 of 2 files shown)


src/features/clusters/queries/getHarperVersionsQuery.ts
```diff
@@ -7,15 +7,52 @@ export interface HarperVersionsResponse {
...
+}
+
 async function getHarperVersions() {
 	// TODO: OpenAPI from CM is erroring, so this new endpoint isn't described.
 	const { data } = await apiClient.get(`/HarperVersions/` as any);
-	return data as HarperVersionsResponse;
+	const response = data as HarperVersionsResponse;
+	return {
+		...response,
+		value: dedupeHarperVersionsByTag(response?.value ?? []),
```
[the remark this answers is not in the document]
dawsontoth: Done — dropped the `?.value ?? []` guard and pass `response.value` directly, trusting the `HarperVersionsResponse` cast.

src/features/clusters/queries/getHarperVersionsQuery.ts
```diff
@@ -7,15 +7,52 @@ export interface HarperVersionsResponse {
...
+}
+
 async function getHarperVersions() {
 	// TODO: OpenAPI from CM is erroring, so this new endpoint isn't described.
 	const { data } = await apiClient.get(`/HarperVersions/` as any);
-	return data as HarperVersionsResponse;
+	const response = data as HarperVersionsResponse;
+	return {
+		...response,
+		value: dedupeHarperVersionsByTag(response.value),
```
kriszyp, line 54: Minor/non-blocking: since the guard (`response?.value ?? []`) was dropped in the review-fixup commit, a malformed response (no `value`) will throw inside this `queryFn`. React Query catches it and surfaces a toast (`QueryCache.onError`), and every consumer already null-checks `harperVersions?.value`, so this fails safely — just calling it out because the TODO right above notes this endpoint's OpenAPI description is currently broken, i.e. the `as HarperVersionsResponse` cast isn't schema-backed, which is the case where "trust the cast" is weakest. If you want the malformed-response path documented, a quick test asserting `getHarperVersions()` surfaces (rather than swallows) that error would do it — otherwise fine as-is.

kriszyp (approved):
I think I am the guilty party here, I think at some point accidentally published a release as both stable and next. Sorry about that. But thank you for adding a check for that.

src/features/clusters/queries/getHarperVersionsQuery.ts
dawsontoth: Documented the malformed-response path with a test asserting `getHarperVersions()` surfaces (rather than swallows) the throw when the body has no `value`; kept the guard dropped per the earlier review, since it fails safely via `QueryCache.onError` + the consumers' `harperVersions?.value` null-checks.

PR #1538, opened 2026-07-20
chore(renovate): group pre-1.0 dprint/oxfmt bumps with non-major deps

dawsontoth:
## What

Treat pre-1.0 `dprint`/`oxfmt` bumps as **non-major**, so they group into the existing "all non-major dependencies" PR instead of opening a separate PR every release.

## Why

Under caret semantics `^0.x` doesn't satisfy the next `0.y`, so Renovate splits every pre-1.0 bump of these formatters into its own PR. They're stable in practice and any breakage surfaces immediately in the formatter tooling — no need for individual review.

## How

Adds a `packageRules` entry mirroring the existing non-major group, scoped to `dprint`/`oxfmt` while `matchCurrentVersion: "< 1.0.0"` (matches this repo's group, which does not automerge). Once either reaches 1.0.0 the rule stops matching and they return to the normal major cadence automatically. Validated with `renovate-config-validator`.

Companion PRs applied to the other repos carrying these tools (schema-codegen #39, agent-tools #28, create-harper #122).

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -40,8 +40,7 @@
 			"groupName": "all non-major dependencies",
 			"groupSlug": "all-minor-patch",
 			"matchPackageNames": ["dprint", "oxfmt"],
-			"matchCurrentVersion": "< 1.0.0",
-			"matchUpdateTypes": ["minor", "patch"]
+			"matchCurrentVersion": "< 1.0.0"
 		}
 	]
 }
--- a/renovate.json
+++ b/renovate.json
@@ -34,6 +34,14 @@
 				"patch"
 			],
 			"matchCurrentVersion": "!/^0/"
+		},
+		{
+			"description": "Treat pre-1.0 dprint/oxfmt (stable formatters) as non-major so bumps group with other deps; regroups normally once they reach 1.0.0+.",
+			"groupName": "all non-major dependencies",
+			"groupSlug": "all-minor-patch",
+			"matchPackageNames": ["dprint", "oxfmt"],
+			"matchCurrentVersion": "< 1.0.0",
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -34,6 +34,13 @@
 				"patch"
 			],
 			"matchCurrentVersion": "!/^0/"
+		},
+		{
+			"description": "Treat pre-1.0 dprint/oxfmt (stable formatters) as non-major so bumps group with other deps; regroups normally once they reach 1.0.0+.",
+			"groupName": "all non-major dependencies",
+			"groupSlug": "all-minor-patch",
+			"matchPackageNames": ["dprint", "oxfmt"],
+			"matchCurrentVersion": "< 1.0.0"
```
kriszyp, line 43: Could we restore `matchUpdateTypes: ["minor", "patch"]` here? Renovate keeps `updateType` separate from `isBreaking`, so `0.55.x -> 0.56.x` is still a `minor` update and already matches that list. Without the guard, this rule also applies the `all non-major dependencies` group config to the eventual `0.x -> 1.0.0` major update (and to special types such as pin/bump/rollback) because `matchCurrentVersion` evaluates the old version. Keeping the original update-type filter achieves the intended pre-1.0 minor/patch grouping without broadening the rule.

kriszyp (commented):
I dunno if this is helpful or not.

renovate.json
dawsontoth: I'm not terribly concerned about it since it's a dev time tool, and well checked in the GHAs. Adding what you said might trigger it to not include what it computes as the major version (second digit when less than 0). So I don't know how it'd behave... I'd rather roll it out and tune it if we have issues.

PR #1540, opened 2026-07-20
feat(clusters): hide container ops on self-hosted, gate Deployments on auth

dawsontoth:
## What

Hides a few cluster features that don't apply to **self-hosted** clusters, plus gates the Deployments menu item on cluster authentication.

### 1. Container actions hidden on self-hosted clusters
Self-hosted clusters have no managed container lifecycle — Harper doesn't control their runtime — so the **Container** action group (Start, Start in safe mode, Restart, Restart in safe mode, Stop) is now hidden for them in:
- the cluster-card context menu / "…" dropdown (`ClusterCard`)
- the per-instance context menu / "…" dropdown (`useInstanceMenuItems`)

This matches the existing behavior of the cluster-overview **"Cluster actions"** dropdown (`ClusterStateMenu`), which already returns `null` for self-hosted clusters — so no change was needed there.

> Note: the request specifically named Restart / Restart in safe mode / Stop, but I hid the whole group (Start / Start in safe mode too) for consistency with `ClusterStateMenu`'s "self-hosted clusters have no container ops" rule. In practice self-hosted clusters are `RUNNING`, so only the three named ones were ever visible anyway.

### 2. "Deployments" gated on cluster authentication
The **Deployments** cluster-card menu item now only appears once you've authenticated with the cluster (`!!auth.user`, the same signal `ClusterHome` uses for its connected state) — previously it showed whenever you had view permission.

## Verification
- `tsc -b`, `oxlint`, `dprint`, and the full `vitest` suite (241 files / 1732 tests) all pass.
- Verified live against real cluster data (worktree dev server on :5173) with a managed cluster ("Anvils") and a self-hosted cluster ("Self"), neither authenticated:

| Menu | Container actions | Deployments |
|---|---|---|
| **Self** (self-hosted) cluster card | absent ✓ | absent (not authed) ✓ |
| **Self** instance menu | absent ✓ | — |
| **Anvils** (managed) cluster card — control | Restart / Restart in safe mode / Stop present ✓ | absent (not authed) ✓ |

The managed cluster confirms the container-ops change is correctly scoped to self-hosted; Deployments disappearing for both clusters (where it previously showed unconditionally) confirms the auth gate.

### DIFF
--- a/src/features/cluster/useInstanceMenuItems.tsx
+++ b/src/features/cluster/useInstanceMenuItems.tsx
@@ -85,10 +85,11 @@ export function useInstanceMenuItems(
 
 	// Container lifecycle ops (stop/start/restart) — distinct from the proxied Harper "restart".
 	// Only offered from a settled RUNNING/STOPPED state; hidden mid-transition (the instances poll
-	// reveals the resting state and the actions reappear).
+	// reveals the resting state and the actions reappear). Self-hosted clusters have no managed
+	// container lifecycle (Harper doesn't control their runtime), so the group is hidden for them.
 	const isRunning = instance.status === 'RUNNING';
 	const isStopped = instance.status === 'STOPPED';
-	const hasContainerOps = canManage && (isRunning || isStopped);
+	const hasContainerOps = canManage && !isSelfManaged && (isRunning || isStopped);
 
 	const actions: EntityMenuItem[] = [
 		hasAuth && isDirectlyLoggedIn && {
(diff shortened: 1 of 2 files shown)


DavidCockerill (approved):
LGTM — clean visibility fix. Hiding the Container group on self-hosted clusters is correct (Harper doesn't control their runtime), and it uses the canonical `clusterIsSelfManaged` / the `isSelfManaged` param consistently, so the cluster card, per-instance menu, and `ClusterStateMenu` now all agree. Verified `isSelfManaged` is in scope (it's a declared param of `useInstanceMenuItems`, already used just below), so no undefined-ref risk. One minor, non-blocking: the Deployments gate uses `!!auth.user` (any authenticated Studio user) — worth confirming that's the intended signal vs a cluster-scoped one, but it's a reasonable minimum.

— DAIvid (Claude Fable 5)

dawsontoth:
Yup, it's intentional on the Deployments.

PR #1542, opened 2026-07-21
feat: Resend invite in user modal + verification flow for unverified login

dawsontoth:
## Summary

Two related account/user-management improvements.

### 1. "Resend invite" in the user detail modal
The org users table already offers a **Resend invite** action on `PENDING` rows. This surfaces the same action inside `EditUserModal`, so it's reachable after opening a pending user — not only from the row.

- Reuses the existing `ResendInviteButton` component, so permission gating (org-role `update`), the invite mutation, and toasts are identical to the row.
- Rendered only when the user's `status === 'PENDING'` (in a bordered section above "Remove from organization").

### 2. Route unverified users into the email-verification flow on login
An email/password login for an unverified account is rejected by central-manager with **403 "User has not verified email address"** — and, notably, this check runs *after* the password is validated, so reaching it proves the credentials are valid. The cloud login mutation had no `onError`, so it dead-ended on a generic red **"Error"** toast with no way to get a new verification link.

Now, when a login fails specifically because the email is unverified, the app:
- **detects** that rejection (403 **and** a message match, so a 403 *deactivated* account isn't mis-routed),
- **auto-resends** a fresh verification link (safe — the branch is only reachable with valid credentials, so it can't be abused for spam),
- shows a friendly informational toast, and
- **redirects** to the existing `/verifying` page.

All other login failures (wrong password → 401, deactivated → 403, etc.) keep the standard error toast.

Also adds a `meta.skipGlobalErrorToast` opt-out to the `MutationCache` so a mutation can render its own error UX instead of the shared toast.

## Testing
- New unit tests for the `isEmailNotVerifiedError` detector (403-unverified vs 403-deactivated vs 401 vs non-axios).
- New `useCloudSignIn` flow test: unverified → resend + redirect + suppressed error toast; invalid credentials → standard toast, no redirect/resend.
- New `queryClient` test: `meta.skipGlobalErrorToast` suppresses the global toast (and the existing "global toast fires by default" test still passes).
- Full suite green (`vitest run`: 1743 passed, 11 skipped), `tsc -b` clean, oxlint + dprint clean.

### Verified live against the stage backend
- **Point 1 (modal):** Resend invite appears for PENDING users, absent for ACTIVE users.
- **Point 2 (unverified login):** created a new account, left it unverified, then signed in with it again → correctly redirected to the **"Check your email"** screen with the email pre-filled (verification link auto-resent). Note: this path only manifests against stage/prod — a *local* backend auto-verifies signups (`isVerified || isLocal`).

### DIFF
--- a/src/features/auth/hooks/useCloudSignIn.test.tsx
+++ b/src/features/auth/hooks/useCloudSignIn.test.tsx
@@ -62,9 +62,28 @@ describe('useCloudSignIn — unverified email', () => {
 
 		await waitFor(() => expect(navigate).toHaveBeenCalledWith({ to: '/verifying?email=unverified%40example.com' }));
 		expect(apiClient.post).toHaveBeenCalledWith('/ResendVerificationEmail/', { email: EMAIL });
-		expect(toast.info).toHaveBeenCalled();
+		// The "we sent a link" toast only fires once the resend actually succeeds.
+		await waitFor(() => expect(toast.info).toHaveBeenCalled());
 		// The dead-end "Error" toast must NOT fire for this case.
 		expect(toast.error).not.toHaveBeenCalled();
+	});
+
+	it('still redirects (without claiming a link was sent) when the resend fails', async () => {
+		post.mockImplementation((url: string) => {
+			if (url === '/Login/') {
+				return Promise.reject(axiosError(403, { error: 'User has not verified email address' }));
+			}
+			// Resend fails (e.g. rate-limited) — the user must still reach /verifying, and we must
+			// NOT show a "we sent a link" toast that never happened.
+			return Promise.reject(axiosError(429, { error: 'Too many requests' }));
+		});
+
+		const { result } = renderHook(() => useCloudSignIn(), { wrapper: wrapper() });
+		act(() => result.current.submitForm({ email: EMAIL, password: 'correct-horse' }));
+
+		await waitFor(() => expect(navigate).toHaveBeenCalledWith({ to: '/verifying?email=unverified%40example.com' }));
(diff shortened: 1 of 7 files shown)


DavidCockerill (approved):
LGTM — nice UX fix, and the auth-flow subtleties hold up. I verified the parts that could bite: the react-query interaction is clean (the global `errorHandler` only toasts, and the login mutation opts out via `meta.skipGlobalErrorToast` then re-dispatches by hand — so exactly one toast, no logout/redirect double-fire, other 403s untouched); it's enumeration-safe (the 403 is only reachable post-password-validation, wrong-password → 401 which this ignores); and the resend-invite button genuinely reuses the org-role `update` gate and only renders for `PENDING`. Tests are substantial and cover the error shapes + the `meta` suppression directly. (gemini's two points are already addressed at this head.)

Two non-blocking things worth a look, neither gating:
- **Detection keys on `/verif/i` in the server's English error text.** It correctly requires both the 403 and the match, but it's looser than its own doc comment and is the fragile class — a CM reword silently reverts the feature to the old dead-end toast, and any other 403 containing "verif" would get funneled into the verification flow. If central-manager exposes a stable error code/field for "email not verified," keying on that would be sturdier.
- **Auto-resend failure surfaces a stray toast.** The auto-resend mutation doesn't set `skipGlobalErrorToast`, so a rate-limited (429) resend fires the generic red error toast on top of the `/verifying` redirect. Minor, and the flow test can't see it (its isolated QueryClient has no global MutationCache.onError) — worth suppressing/handling that one path.

Also minor: the email rides in the URL (`/verifying?email=`) — low-severity PII-in-query, password isn't. Approving as-is.

— DAIvid (Claude Fable 5)

Issue #1499, opened 2026-07-14
[RUM] Monaco worker OOM recurrence via unbounded type-acquisition extraLibs (regression of #1370)


dawsontoth:
## Summary

The Monaco editor worker OOM pair — `DataCloneError: Failed to execute 'postMessage' on 'Worker': Data cannot be cloned, out of memory.` and its downstream `FAILED to post message to worker` — **recurred in production** on 2026-07-13, **after** the fixes that closed #1370 (#1373 + the 2026-07-10 model-budget work `15f3ba99` / `0aa3c760`, all on `prod`).

This is a follow-up to (now-closed) #1370. The earlier fixes bounded the *models* synced to the language worker; this recurrence comes through a **different, still-unbounded path**: automatic type acquisition (`addExtraLib`).

## Datadog RUM findings (last 24h, app `f590deee-…`)

- **Volume:** 1,560 combined occurrences (780 `DataCloneError` + 780 `FAILED to post message to worker`), out of 1,627 total RUM errors — i.e. this issue *is* the 24h error spike (baseline the prior 24h was 25 errors).
- **Blast radius:** **2 sessions / 2 users**, each firing ~400 errors within a single hour — a runaway per-session flood that leaves the worker stuck and drowns real RUM signal. Two isolated bursts (~409 and ~371), then zero for the following ~18h.
- **Views:** the applications code editor — `…/apps/` (830) and `…/instance/$instanceId/` (750). This is a *different surface* than the log viewer #1370 originally covered.
- **Browser:** Chrome (1,623 / 1,627).
- **Bundle:** `index-BDXPnPND.js` (minified prod).

*(Session/user IDs omitted intentionally — no customer-identifying data. The error is a generic frontend OOM with no payload contents in the message/stack.)*

## Root cause

`setEagerModelSync(true)` (`harper-language/typescript.ts`) makes Monaco structured-clone **both** every model **and** every `addExtraLib` declaration to the TS/JS language worker over `postMessage`.

The three guards added for #1370/#1407 all bound **models**:
- per-file `MAX_WORKER_MODEL_CHARS` (512 KB → plaintext),
- aggregate `MAX_APPLICATION_MODEL_CHARS_TOTAL` (4 MB across live `file:///` models),
- `MAX_LIVE_APPLICATION_MODELS` (150-model ceiling).

[...]

PR #1543, opened 2026-07-21
fix(browse): worker-free JSON in record editors to end Monaco worker OOM (#1370/#1499)

### LINKED ISSUE #1370: [Monitoring] Monaco editor worker OOM — DataCloneError / postMessage failure (regression since v2.115.4)
## Summary

Two related Monaco editor errors are firing at high frequency in production (270 combined occurrences in the last 24 hours), both classified as regressions by Datadog Error Tracking.

### 1. DataCloneError: Data cannot be cloned, out of memory
- **Datadog issue:** https://app.datadoghq.com/error-tracking/issue/914164d2-65c5-11f1-b4ae-da7ad0900002
- **Frequency (24h):** ~135 occurrences
- **First seen:** v2.114.0 (2026-06-11), resolved, then **regressed at v2.115.4** (2026-06-16)
- **Last seen:** v2.121.5 (2026-06-23)
- **Error (unhandled):**
  ```
  DataCloneError: Failed to execute 'postMessage' on 'Worker': Data cannot be cloned, out of memory.
      at editor.api-*.js:14:48370
  ```

### 2. FAILED to post message to worker
- **Datadog issue:** https://app.datadoghq.com/error-tracking/issue/0def2864-691f-11f1-960d-da7ad0900002
- **Frequency (24h):** ~135 occurrences
- **First seen:** v2.115.4 (2026-06-16)
- **Last seen:** v2.121.5 (2026-06-23)
- **Error:**
  ```
  Error: FAILED to post message to worker
      at index-*.js:459:27326
  ```

## Context

Both errors originate from Monaco editor's worker communication layer (`monaco-editor/esm/vs/base/common/errors.js`). The DataCloneError suggests the editor is attempting to serialize a message that's too large or contains non-cloneable data when communicating between the main thread and a Monaco web worker. The second error appears to be a downstream consequence.

The root view is `/$organizationId/$clusterId/` (the cluster home/logs view), meaning this likely triggers when the editor initializes there.

## Suggested Investigation

- Check what data is being posted to the Monaco worker on the cluster view — look for large or non-cloneable payloads added around v2.115.4.
- Consider whether log content or editor state being passed to the worker could be unbounded in size.
- Check if anything changed in the editor API bundle between v2.114.x and v2.115.4.

---
*Reported by automated Datadog RUM review — 2026-06-24*

dawsontoth:
## Summary

RUM flagged a recurrence of the closed [#1370](https://github.com/harperfast/studio/issues/1370) / [#1499](https://github.com/harperfast/studio/issues/1499) Monaco-worker-OOM class through the browse **record editors** (`EditTableRowModal`, `AddTableRowModal`), which hardcoded `language="json"`. Monaco's `json` language runs a worker and clones the model's full text (and every change delta) to it over `postMessage`; a large record — a row with a big blob/array field, or a bulk-insert array pasted into "Add row" — overflows the structured clone and crashes the worker with `DataCloneError: ... out of memory.`, which Monaco reports as an unhandled error and then retries in a tight loop, flooding the session.

**This revision replaces the earlier size-based `json`→`plaintext` switch with a worker-free JSON language**, after review ([@kriszyp](https://github.com/kriszyp)) correctly pointed out the switch can't close the paste path (details below).

## Approach

- New Monaco language `json-worker-free` ([`workerFreeJsonLanguage.ts`](https://github.com/HarperFast/studio/blob/claude/rum-browse-record-editor-worker-oom/src/lib/monaco/workerFreeJsonLanguage.ts)): registers Monaco's **main-thread** JSON tokenizer plus the standard JSON language configuration (brackets, auto-closing pairs), and **none** of the worker-backed providers (validation, folding, colors, completion, hover, formatting). Registered in `setup.ts` alongside the built-in `json` contribution, which is left untouched for the config/roles editors that rely on its schema validation.
- Both record editors switch to `json-worker-free`. There is no language worker to overflow, at any size; syntax highlighting and bracket/auto-close are unchanged.
- Worker validation markers go away with the worker, so the `onValidate` signal is replaced by [`recordEditorJson.ts`](https://github.com/HarperFast/studio/blob/claude/rum-browse-record-editor-worker-oom/src/features/instance/databases/modals/recordEditorJson.ts): a cheap **main-thread** parse for content under `MAX_WORKER_MODEL_CHARS` keeps the "Save disabled while JSON is invalid" affordance for normal-size records without parsing a huge buffer on every keystroke, and a **caught** `JSON.parse` at submit time toasts on malformed input (for both Add and Edit) instead of throwing an uncaught `SyntaxError`.
- `EditTableRowModal` resets its draft/validity when a different record is opened (that modal instance is reused across rows).

## Why not the size-based switch

`@monaco-editor/react` applies a changed `language` prop in a later React effect, and Monaco syncs the model to the worker on a debounce; once synced, the worker's own `onDidChangeContent` listener posts the oversized delta on the *next change*. So starting from a small sample and pasting a >512 KB array hands the payload to the worker before any `json`→`plaintext` switch can run — the switch also doesn't stop an already-running model sync. A worker-free language sidesteps the race entirely rather than trying to win it.

## Where to focus review

- **Tradeoff — live validation:** worker-free means no inline red-squiggle validation. Mitigated by the main-thread Save gating (normal records) + submit-time parse/toast (all records). This is a deliberate behavior change for the browse record editors; if the team wants richer inline validation back, a main-thread JSON diagnostics provider would be a follow-up.
- **`createTokenizationSupport` import** (`monacoJsonTokenization.d.ts` + `workerFreeJsonLanguage.ts`): reuses Monaco's own JSON tokenizer from an internal ESM path (same style as the existing deep imports in `setup.ts`/`languageServices.ts`) for pixel-identical highlighting. It ships without types, so it's declared locally. A Monaco version bump could move it — `tsc -b` / `vite build` would catch that.
- **Out of scope:** Monaco's base editor worker still syncs models for link detection (a `'*'` provider), exactly as it does for every `plaintext` editor the app already treats as the safe fallback. This PR closes the `json`-language-worker path the RUM data implicates; the base-worker exposure is pre-existing and shared, not addressed here.

## Testing

- `workerFreeJsonLanguage.test.ts`: asserts the language registers tokenization + config and **no** worker-backed provider.
- `recordEditorJson.test.ts`: validity/parse helpers (valid, malformed, empty, oversized-deferred).
- `AddTableRowModal.test.tsx` (new) + `EditTableRowModal.test.tsx` (updated): assert both editors use `json-worker-free`, a valid record saves the parsed value, and an **oversized malformed paste** is caught at submit time (toast, no save) — the large-paste integration regression asked for in review.
- Full suite green (1786 passed / 11 skipped), `tsc -b` clean, `oxlint` + `dprint` clean, `vite build` clean.

Not browser-verified: reproducing the crash needs a live instance with a >512 KB record; verified at the mechanism/unit level and via a production build (the tokenizer import bundles into the `editor.api2` chunk).

🤖 Generated by an LLM (Claude Opus 4.8) via Claude Code.

### DIFF
--- a/src/features/instance/databases/modals/AddTableRowModal.tsx
+++ b/src/features/instance/databases/modals/AddTableRowModal.tsx
@@ -7,11 +7,13 @@ import { useMonacoTheme } from '@/hooks/useMonacoTheme';
 import { InstanceAttribute, InstanceDatabaseTableMap, InstanceTable } from '@/integrations/api/api.patch';
 import { useInsertTableRecords } from '@/integrations/api/instance/database/insertTableRecords';
 import { Editor } from '@/lib/monaco/MonacoEditor';
+import { WORKER_FREE_JSON_LANGUAGE_ID } from '@/lib/monaco/workerFreeJsonLanguage';
 import { pluralize } from '@/lib/pluralize';
 import type { EditorProps, OnMount } from '@monaco-editor/react';
 import { Save, TerminalIcon } from 'lucide-react';
 import { useCallback, useMemo, useState } from 'react';
 import { toast } from 'sonner';
+import { isRecordJsonProbablyValid, tryParseRecordJson } from './recordEditorJson';
 
 export function AddTableRowModal({
 	isModalOpen,
@@ -52,55 +54,57 @@ export function AddTableRowModal({
 		return JSON.stringify(sample, null, 4);
 	}, [instanceTable, databaseTables]);
 
-	const onValidate = useCallback((markers: unknown[]) => {
-		setMadeChanges(true);
-		setIsValidJSON(markers.length === 0);
-	}, [setIsValidJSON]);
-
 	const onSubmitClick = useCallback(() => {
-		if (addTableRecordData && isValidJSON) {
-			const data = JSON.parse(addTableRecordData);
-			const records = Array.isArray(data) ? data : [data];
-			const toastId = toast.loading(`Adding ${records.length} records...`);
-			addTableRecords(
-				{
-					...instanceParams,
-					databaseName: instanceTable.schema,
-					tableName: instanceTable.name,
-					records,
-				},
-				{
-					onSuccess: (response) => {
-						void refreshTable();
-						if (!response.skipped_hashes?.length) {
-							setIsModalOpen(false);
-						}
-						setSkippedHashes(response.skipped_hashes);
-						(response.skipped_hashes?.length > 0 ? toast.warning : toast.success)(
-							response.skipped_hashes?.length > 0 ? 'Warning!' : 'Success!',
-							{
-								id: toastId,
-								description: (
-									<>
-										{response.inserted_hashes.length > 0
-											&& <p>Added {pluralize(response.inserted_hashes.length, 'record', 'records')}</p>}
-										{response.skipped_hashes.length > 0
-											&& <p>Skipped {pluralize(response.skipped_hashes.length, 'record', 'records')}</p>}
-									</>
-								),
-							},
-						);
-					},
-				},
-			);
+		if (!addTableRecordData) {
+			return;
 		}
+		// Authoritative parse: the live check skips oversized content, so a large,
+		// malformed bulk paste can reach here with isValidJSON still true — parse it
+		// in a catch rather than letting Save throw an uncaught SyntaxError.
+		const parsed = tryParseRecordJson(addTableRecordData);
+		if (!parsed.ok) {
+			toast.error("This record isn't valid JSON — fix the syntax and try again.");
+			return;
+		}
+		const records = (Array.isArray(parsed.value) ? parsed.value : [parsed.value]) as object[];
+		const toastId = toast.loading(`Adding ${records.length} records...`);
+		addTableRecords(
+			{
+				...instanceParams,
+				databaseName: instanceTable.schema,
+				tableName: instanceTable.name,
+				records,
+			},
+			{
+				onSuccess: (response) => {
+					void refreshTable();
+					if (!response.skipped_hashes?.length) {
+						setIsModalOpen(false);
+					}
+					setSkippedHashes(response.skipped_hashes);
+					(response.skipped_hashes?.length > 0 ? toast.warning : toast.success)(
+						response.skipped_hashes?.length > 0 ? 'Warning!' : 'Success!',
+						{
+							id: toastId,
+							description: (
+								<>
+									{response.inserted_hashes.length > 0
+										&& <p>Added {pluralize(response.inserted_hashes.length, 'record', 'records')}</p>}
+									{response.skipped_hashes.length > 0
+										&& <p>Skipped {pluralize(response.skipped_hashes.length, 'record', 'records')}</p>}
+								</>
+							),
+						},
+					);
+				},
+			},
+		);
 	}, [
 		addTableRecordData,
 		addTableRecords,
 		instanceParams,
 		instanceTable.name,
 		instanceTable.schema,
-		isValidJSON,
 		refreshTable,
 		setIsModalOpen,
 	]);
@@ -140,11 +144,16 @@ export function AddTableRowModal({
 				<div className="flex-1 min-h-0 w-full">
 					<Editor
 						className="w-full h-full"
-						language="json"
+						// Worker-free JSON: highlighting without a language worker that a large bulk-insert
+						// array pasted here could overflow and crash (studio#1370/#1499).
+						language={WORKER_FREE_JSON_LANGUAGE_ID}
 						theme={monacoTheme}
 						value={sampleJSON}
-						onValidate={onValidate}
-						onChange={setAddTableRecordData}
+						onChange={(updatedValue) => {
+							setAddTableRecordData(updatedValue);
+							setMadeChanges(true);
+							setIsValidJSON(isRecordJsonProbablyValid(updatedValue));
+						}}
 						options={{ minimap: { enabled: false }, automaticLayout: true }}
 						onMount={handleEditorDidMount}
 					/>
--- a/src/features/instance/databases/modals/EditTableRowModal.tsx
+++ b/src/features/instance/databases/modals/EditTableRowModal.tsx
@@ -4,8 +4,11 @@ import { Button } from '@/components/ui/button';
 import { Dialog, DialogContent, DialogFooter, DialogHeader, DialogTitle } from '@/components/ui/dialog';
 import { useMonacoTheme } from '@/hooks/useMonacoTheme';
 import { Editor } from '@/lib/monaco/MonacoEditor';
+import { WORKER_FREE_JSON_LANGUAGE_ID } from '@/lib/monaco/workerFreeJsonLanguage';
 import { Save, Trash, TriangleAlert } from 'lucide-react';
-import { useCallback, useMemo, useState } from 'react';
+import { useMemo, useState } from 'react';
+import { toast } from 'sonner';
+import { isRecordJsonProbablyValid, tryParseRecordJson } from './recordEditorJson';
 
 export function EditTableRowModal({
 	canEditRecords,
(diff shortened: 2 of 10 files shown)


dawsontoth:
Good call — applied in 26f5fc31. Defaulted empty/nullish content to `json` so the record editor keeps auto-close brackets, indentation, and highlighting from the first keystroke; the OOM guard now only trips once content actually exceeds `MAX_WORKER_MODEL_CHARS`. Updated the doc comment and the unit test accordingly.

src/features/instance/databases/modals/AddTableRowModal.tsx
```diff
@@ -140,7 +141,10 @@ export function AddTableRowModal({
 				<div className="flex-1 min-h-0 w-full">
 					<Editor
 						className="w-full h-full"
-						language="json"
+						// Guard the JSON worker against oversized payloads: a large bulk-insert array
+						// pasted here would overflow Monaco's structured-clone buffer and crash the
+						// worker (studio#1370/#1499).
+						language={chooseRecordEditorLanguage(addTableRecordData ?? sampleJSON)}
```
kriszyp: Could we move this protection before the oversized content mutates Monaco's JSON model? `@monaco-editor/react` invokes `onChange` from `onDidChangeModelContent`, then applies a changed `language` prop in a later React effect. Starting from the small sample and pasting a >512 KiB bulk array therefore puts the large change into the JSON model before this can switch it to plaintext, so the worker may still receive the exact payload this guard is intended to block. Please guard/intercept the paste before mutation (or keep this bulk-input editor worker-free initially), and add an integration regression for a large paste rather than only the pure helper result.

src/features/instance/databases/modals/EditTableRowModal.tsx
```diff
@@ -78,7 +79,9 @@ export function EditTableRowModal({
 						<div className="flex-1 min-h-0 w-full">
 							<Editor
 								className="w-full h-full"
-								language="json"
+								// Guard the JSON worker against oversized records: a huge row would overflow
+								// Monaco's structured-clone buffer and crash the worker (studio#1370/#1499).
+								language={chooseRecordEditorLanguage(updatedTableRecordData ?? value)}
```
kriszyp: `updatedTableRecordData` persists when this unkeyed modal closes and another row is opened, so this can classify the new `value` using the previous row's draft. For example: edit a small row, close it, then open a >512 KiB row; the old small draft wins `??`, returns `json`, and the huge new `value` is still handed to the worker. Could we scope/reset the draft per selected record without a transient stale-language render, and ensure `language` is always derived from the same content Monaco is receiving?

src/features/instance/databases/modals/recordEditorLanguage.ts
```diff
@@ -0,0 +1,30 @@
...
+ * that worker over `postMessage` (a structured clone). An oversized record — a row
+ * carrying a large blob, a big embedded array, or a bulk-insert payload pasted into
+ * the "Add row" editor — overflows the clone buffer and crashes the worker
+ * ("DataCloneError: ... out of memory.", followed by "FAILED to post message to
+ * worker"), which Monaco reports as unhandled errors and retries, flooding the
+ * session (HarperFast/studio#1370 / #1499, recurring here through the browse
+ * record-editor path the applications/log-editor guards don't cover).
+ *
+ * Oversized content renders as `plaintext`, which has no language worker — the same
+ * guard the Applications and log editors apply. JSON syntax highlighting and inline
```
kriszyp: Dropping to plaintext also drops the `onValidate` signal that both submit handlers use to decide whether their unguarded `JSON.parse(...)` is safe. An oversized document can be made malformed while `isValidJSON` remains true, and Save then throws an uncaught `SyntaxError`. Please add a caught, submit-time parse with a user-visible error (rather than parsing this large buffer on every keystroke) for both Add and Edit.

kriszyp (commented):
The suggestions might help protect against OOM?

cb1kenobi (approved):

src/features/instance/databases/modals/AddTableRowModal.tsx
dawsontoth: You're right — the size-based `json`→`plaintext` switch can't win this race (the prop applies in a later effect, and once the model is synced the worker's own change listener posts the oversized delta anyway). Reworked in 76473647: both record editors now use a dedicated **worker-free** JSON language (`json-worker-free`) that registers Monaco's main-thread tokenizer + the standard JSON language config but **none** of the worker-backed providers — so there is no language worker to hand the pasted payload to, at any size, and nothing to intercept before mutation. This also takes your "keep this bulk-input editor worker-free initially" suggestion directly. Added the large-paste integration regression you asked for in `AddTableRowModal.test.tsx`: an oversized malformed bulk paste stays on the worker-free language and is caught at submit time (toast, no insert).

src/features/instance/databases/modals/EditTableRowModal.tsx
dawsontoth: Fixed in 76473647. `language` is no longer derived from the draft (or from content at all) — it's the constant worker-free language — so the stale-language render you describe can't happen. The underlying stale-**draft** bug you flagged (this modal instance is reused across rows) is fixed too: `EditTableRowModal` now resets the draft/validity when a different record loads, done during render rather than in an effect so there's no transient frame with the previous row's draft live. Regression: `resets the draft when a different record is opened` in `EditTableRowModal.test.tsx`.

src/features/instance/databases/modals/recordEditorLanguage.ts
dawsontoth: Done in 76473647. Both submit handlers now run a caught parse (`tryParseRecordJson`) and toast on malformed input instead of letting `JSON.parse` throw. And we don't parse the large buffer on every keystroke: the live Save-button check (`isRecordJsonProbablyValid`) only parses content under `MAX_WORKER_MODEL_CHARS` and defers anything larger to that single submit-time parse. Regressions cover oversized-malformed → toast + no save for both Add and Edit. (This file, `recordEditorLanguage.ts`, was replaced by `recordEditorJson.ts`.)

src/features/instance/databases/modals/recordEditorLanguage.ts
```diff
@@ -0,0 +1,25 @@
...
+ * validation are lost for those (rare, pathologically large) records, which beats
+ * crashing the worker for the whole tab.
+ */
+export function chooseRecordEditorLanguage(content: string | undefined | null): 'json' | 'plaintext' {
+	if (!content || content.length > MAX_WORKER_MODEL_CHARS) {
+		return 'plaintext';
+	}
+	return 'json';
+}
```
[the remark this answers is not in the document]
dawsontoth: Obsolete: `recordEditorLanguage.ts` was removed in the rework. The record editors now use a dedicated worker-free JSON language, so highlighting and auto-close are always on regardless of content — there's no longer any empty→plaintext case to guard. See the updated PR description.

src/features/instance/databases/modals/recordEditorLanguage.test.ts
```diff
@@ -0,0 +1,31 @@
...
+describe('chooseRecordEditorLanguage', () => {
+	it('uses json highlighting for a normal-sized record', () => {
+		expect(chooseRecordEditorLanguage('{"id":1,"name":"widget"}')).toBe('json');
+	});
+
+	it('treats empty or nullish content as plaintext', () => {
+		expect(chooseRecordEditorLanguage('')).toBe('plaintext');
+		expect(chooseRecordEditorLanguage(undefined)).toBe('plaintext');
+		expect(chooseRecordEditorLanguage(null)).toBe('plaintext');
+	});
```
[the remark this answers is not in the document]
dawsontoth: Obsolete: this test file was removed alongside `recordEditorLanguage.ts`, superseded by `recordEditorJson.test.ts` plus the `Add`/`EditTableRowModal` integration tests.

PR #1545, opened 2026-07-23
fix(browse): guard Delete Row against not-yet-loaded record (RUM: undefined[0] TypeError)

dawsontoth:
## Summary

Fixes an **unhandled** `TypeError: Cannot read properties of undefined (reading '0')` surfaced in Datadog RUM on the table browse view (`/$organizationId/$clusterId/databases/$databaseName/$tableName/`), current deploy (commit `2ccba97c`).

## Root cause

The **Delete Row** button in `EditTableRowModal` reads the primary-key value with `data[0]?.[primaryKey]`. But `data` is `undefined` while the record fetch is in flight — the parent passes `data={searchByIdData?.data}` (see [`DatabaseTableView.tsx:616-618`](https://github.com/HarperFast/studio/blob/stage/src/features/instance/databases/components/DatabaseTableView.tsx#L616)), and the footer (containing Delete Row) renders regardless of whether `data` has loaded. The `?.` only guarded the `[primaryKey]` access, **not** the `[0]` — so clicking Delete before the record resolves does `undefined[0]` and throws.

The prop type declared `data` as a non-optional array, which hid the mismatch, even though the rest of the component already treats it as optional (`data?.map(...)`, `{data ? … : <Loading/>}`).

## Fix

- Make the prop type honest: `data?: …[]` (documents the loading state).
- Guard the read: `data?.[0]?.[primaryKey]`.
- Add a regression test that renders with `data={undefined}` and clicks **Delete Row**, asserting it neither throws nor calls `onDeleteRecord`. (Fails the vitest run — exit 1 — without the fix.)

## Datadog RUM findings

- **Error:** `TypeError: Cannot read properties of undefined (reading '0')` — `handling: unhandled`
- **View:** `/$organizationId/$clusterId/databases/$databaseName/$tableName/`
- **Frequency:** 2 events / 1 session / 1 user over the last 7d, all on commit `2ccba97c`
- **Frame:** `onClick @ index-*.js` (the Delete Row handler), confirmed against the deployed bundle

## Verification

`tsc -b`, `oxlint`, `dprint` clean; full suite green (1771 passed). The race (click Delete in the window before the record fetch resolves) isn't practically reproducible in the browser preview, so the unit test — which reproduces the exact `data=undefined` state — is the verification.

### DIFF
--- a/src/features/instance/databases/modals/EditTableRowModal.test.tsx
+++ b/src/features/instance/databases/modals/EditTableRowModal.test.tsx
@@ -2,7 +2,7 @@
  * @vitest-environment jsdom
  */
 import { EditTableRowModal } from '@/features/instance/databases/modals/EditTableRowModal';
-import { cleanup, render, screen } from '@testing-library/react';
+import { cleanup, fireEvent, render, screen } from '@testing-library/react';
 import { afterEach, beforeAll, describe, expect, it, vi } from 'vitest';
 
 // Monaco can't load in jsdom; stub the editor with a plain element that exposes the value it was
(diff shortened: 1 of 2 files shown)


cb1kenobi (approved):
I think I ran into this once. Nice fix!

PR #483, opened 2025-07-02
Use our dog svg, fallback to existing favicon png

dawsontoth:
I tested this in Chrome, Firefox and Safari.

PR #1552, opened 2026-07-23
feat(notifications): notification center (bell, banner, admin) [#1259]

### LINKED ISSUE #1259: start notification center for fabric studio
API has a status table

Front end does a GET to it \[API forthcoming\]

```
{
  "type": "error",
  "title": "Maintenance!",
  "message": "The fabric system is going down at 2am",
  "url": ? optional "https://blog.harper.com/2027-07-31-maint-notice"
}
```

Front end shows that in a banner when there’s anything

We want the bones of a notification center inside Studio. It should be GitHub esque where you've got a notification bubble in the top-right, and you can get a preview but then also navigate to your navigation center. Our first use case? Being able to turn on system-wide notifications for maintenance, outages with downstream providers, that kind of thing. A few more notes:

1. ideally we'd do this with subscriptions, not polling.
2. notifications can be time based (utc)
3. in studio, add simple management of these notifications in the admin ui portion
4. we'll start with global notifications, but we'll grow into org or user focused notifications too. They'll deliver to the same center, though.
5. notifications may have a deep link attached to them -- to an external website, or to somewhere inside of studio

dawsontoth:
Closes #1259.

The bones of a GitHub-esque notification center for Fabric Studio, backed by the existing central-manager `SystemStatus` table.

## What's here
- **Bell + unread badge** in the navbar (desktop + mobile) with a preview dropdown — `NotificationBell.tsx`.
- **Notification center** at `/notifications` — full list grouped active / upcoming / expired — `features/notifications/index.tsx`.
- **Bottom banner** auto-showing active, un-acknowledged notices — `NotificationBanner.tsx`.
- **Admin management** at `/admin/notifications` (fabric-admin) — create / edit / delete via DataTable + dialog — `features/admin/notifications/`.
- **Live via subscription**, not polling — Harper's table WebSocket; a 60s poll only as a backstop — `NotificationsSubscriptionManager.tsx`.
- **Client-side (localStorage) acks**, time-windowed (UTC) active state, and optional internal (router) / external (new-tab) deep links.

## Notable decisions
- **WebSocket, not SSE.** Probing stage: `GET /SystemStatus/` with `Accept: text/event-stream` returns 0 bytes (the edge buffers `text/event-stream`, even for CM's own tables), while the WS upgrade to `wss://…/SystemStatus` returns HTTP 101 and connects anonymously (matches the table's public read). Still a true subscription; no polling.
- **No central-manager change needed.** `SystemStatus` already exposes public read, fabric-admin-gated writes, and a WebSocket subscription — so #1259's "blocked on CM" is unblocked by the existing table.
- **Banner is bottom-anchored.** Studio's chrome hardcodes the top band (fixed header `top-0`, sub-nav `top-20`, content `mt-32`) on every page, so a top banner would need a global offset refactor; a bottom bar is robust everywhere and never hides the nav. Easy to revisit.

## Scope / follow-ups
Starts **global**; org/user scoping is an additive next step (indexed columns + query filter, same center). Server-side per-user read state and a `title` field are also easy follow-ups. A global notice should also render for signed-out users (currently the surfaces mount inside the authed Dashboard layout) — noted as a fast follow.

## Testing
- 17 new unit tests (window/active filtering, deep-link classification, ack store, Radix bell dropdown); full suite green (1804 pass); `tsc`, `oxlint`, `dprint` clean.
- Verified end-to-end in the local preview against stage: bell badge, preview dropdown, bottom banner, center page, severity styling, internal + external deep links, and ack persistence — via synthetic client-side data (this session's account is `least_privileged`, so the admin CRUD + live WS push are being verified on `dev` with an admin account).

### DIFF
--- a/src/components/NotificationBanner.tsx
+++ b/src/components/NotificationBanner.tsx
@@ -20,7 +20,12 @@ export function NotificationBanner() {
 	if (unacked.length === 0) { return null; }
 
 	return (
-		<div className="fixed inset-x-0 bottom-0 z-50 flex flex-col shadow-[0_-2px_12px_rgba(0,0,0,0.12)]">
+		<div
+			role="region"
+			aria-label="System notifications"
+			aria-live="polite"
+			className="fixed inset-x-0 bottom-0 z-50 flex flex-col shadow-[0_-2px_12px_rgba(0,0,0,0.12)]"
+		>
 			{unacked.map((notification) => <BannerStrip key={notification.id} notification={notification} />)}
 		</div>
 	);
--- a/src/components/NotificationBanner.tsx
+++ b/src/components/NotificationBanner.tsx
@@ -1,0 +1,52 @@
+import { ackNotification } from '@/features/notifications/acks';
+import { NotificationLink } from '@/features/notifications/components/NotificationLink';
+import { useUnackedActiveNotifications } from '@/features/notifications/hooks';
+import { getSeverityConfig } from '@/features/notifications/notificationHelpers';
+import { SystemStatusNotification } from '@/integrations/api/api.patch';
+import { cn } from '@/lib/cn';
+import { XIcon } from 'lucide-react';
+
+/**
+ * Full-width banner for active, un-acknowledged notifications — the "hey, look at me" surface for
+ * system-wide notices (#1259). Dismissing a strip acknowledges it (localStorage), so it drops from
+ * the banner and the bell badge but stays visible in the notification center.
+ *
+ * Anchored to the bottom of the viewport: Studio's chrome hardcodes the top band (fixed header at
+ * top-0, sub-nav bars at top-20, page content at mt-32), so a top banner would require offsetting
+ * every page. A bottom bar is robust on every route and never hides the primary navigation.
+ */
+export function NotificationBanner() {
+	const unacked = useUnackedActiveNotifications();
+	if (unacked.length === 0) { return null; }
+
+	return (
+		<div className="fixed inset-x-0 bottom-0 z-50 flex flex-col shadow-[0_-2px_12px_rgba(0,0,0,0.12)]">
+			{unacked.map((notification) => <BannerStrip key={notification.id} notification={notification} />)}
+		</div>
+	);
+}
+
+function BannerStrip({ notification }: { notification: SystemStatusNotification }) {
+	const { Icon, iconClass, bannerClass } = getSeverityConfig(notification.type);
+
+	return (
+		<div className={cn('flex items-start gap-3 border-t px-4 py-3 md:px-12', bannerClass)}>
+			<Icon className={cn('mt-0.5 size-5 shrink-0', iconClass)} />
+			<div className="min-w-0 flex-1 text-sm text-foreground">
+				<span className="break-words">{notification.message}</span>
+				<NotificationLink
+					url={notification.url}
+					className="ml-2 font-medium whitespace-nowrap underline underline-offset-2"
+				/>
+			</div>
+			<button
+				type="button"
+				aria-label="Dismiss notification"
+				onClick={() => ackNotification(notification.id)}
+				className="-my-1 shrink-0 rounded p-1.5 hover:bg-foreground/10"
+			>
+				<XIcon className="size-4" />
+			</button>
+		</div>
+	);
+}
--- a/src/components/NotificationBanner.tsx
+++ b/src/components/NotificationBanner.tsx
@@ -1,7 +1,7 @@
 import { ackNotification } from '@/features/notifications/acks';
 import { NotificationLink } from '@/features/notifications/components/NotificationLink';
 import { useUnackedActiveNotifications } from '@/features/notifications/hooks';
-import { getSeverityConfig } from '@/features/notifications/notificationHelpers';
+import { getSeverity, getSeverityConfig } from '@/features/notifications/notificationHelpers';
 import { SystemStatusNotification } from '@/integrations/api/api.patch';
 import { cn } from '@/lib/cn';
 import { XIcon } from 'lucide-react';
@@ -19,11 +19,14 @@ export function NotificationBanner() {
 	const unacked = useUnackedActiveNotifications();
 	if (unacked.length === 0) { return null; }
 
+	// Interrupt the screen reader for a critical notice (outage), but stay polite for info/maintenance.
+	const hasCritical = unacked.some((notification) => getSeverity(notification.type) === 'critical');
+
 	return (
 		<div
 			role="region"
 			aria-label="System notifications"
-			aria-live="polite"
+			aria-live={hasCritical ? 'assertive' : 'polite'}
 			className="fixed inset-x-0 bottom-0 z-50 flex flex-col shadow-[0_-2px_12px_rgba(0,0,0,0.12)]"
 		>
 			{unacked.map((notification) => <BannerStrip key={notification.id} notification={notification} />)}
--- a/src/features/notifications/NotificationsSubscriptionManager.tsx
+++ b/src/features/notifications/NotificationsSubscriptionManager.tsx
@@ -2,16 +2,27 @@ import { systemStatusQueryKey } from '@/features/notifications/queries';
 import { queryClient } from '@/react-query/queryClient';
 import { useEffect } from 'react';
 
+// A connection has to *stay open* this long before we trust it and reset the backoff. Resetting on
+// the bare handshake instead would let a server that accepts-then-closes the upgrade (post-handshake
+// auth rejection, a CM restart loop, an edge dropping non-conforming upgrades) reconnect-storm forever.
+const HEALTHY_SESSION_MS = 10_000;
+const MAX_BACKOFF_MS = 30_000;
+
 /**
  * Subscribes to central-manager's `SystemStatus` table over Harper's automatic WebSocket endpoint and
  * refetches the notifications query on any change — "subscriptions, not polling" (#1259).
  *
  * WebSocket rather than SSE, on purpose: probing stage showed the edge buffers `text/event-stream`
  * (an SSE GET yields 0 bytes until the connection closes), whereas the WebSocket upgrade passes
- * straight through (HTTP 101) to Harper's table subscription — anonymously, matching the table's
- * public read. On any message we invalidate rather than merge deltas (the list is tiny; a refetch is
- * always correct). WebSockets don't auto-reconnect, so we back off on close and give up after a few
- * tries, leaning on the query's `refetchInterval` backstop. Mounted once from the signed-in Dashboard.
+ * straight through (HTTP 101) to Harper's table subscription. It connects **anonymously by design** —
+ * SystemStatus is public-read broadcast content, so the banner reaches signed-out users on the sign-in
+ * page too. (Note: CM's #483 anonymous-subscribe guard is scoped to `loadAsInstance = false` resources,
+ * which SystemStatus is not, so this permissiveness is by omission — intentional here, not a leak.)
+ *
+ * Resilience: we cap the *backoff* (not the number of attempts) and keep retrying, so a transient
+ * network blip can't strand the tab on the 60s poll backstop for the rest of the session. `failures`
+ * only resets after a healthy session (see above), and we reconnect immediately when the browser fires
+ * `online`. Mounted once from the cloud root (StudioCloud), which never unmounts.
  */
 export function NotificationsSubscriptionManager() {
 	useEffect(() => {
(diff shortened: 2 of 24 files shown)


src/components/NotificationBanner.tsx
```diff
@@ -0,0 +1,52 @@
...
+ * Anchored to the bottom of the viewport: Studio's chrome hardcodes the top band (fixed header at
+ * top-0, sub-nav bars at top-20, page content at mt-32), so a top banner would require offsetting
+ * every page. A bottom bar is robust on every route and never hides the primary navigation.
+ */
+export function NotificationBanner() {
+	const unacked = useUnackedActiveNotifications();
+	if (unacked.length === 0) { return null; }
+
+	return (
+		<div className="fixed inset-x-0 bottom-0 z-50 flex flex-col shadow-[0_-2px_12px_rgba(0,0,0,0.12)]">
```
cb1kenobi: **Low: banner is not announced to screen readers**

The banner mounts dynamically when an active, un-acked notice appears, but the container has no live region, so assistive tech never announces it — the "hey, look at me" surface is silent for screen-reader users. The dismiss button is labeled, but the message itself is not surfaced.

Suggested fix (announce new strips politely; consider `assertive` for critical severities):
```suggestion
		<div role="region" aria-label="System notifications" aria-live="polite" className="fixed inset-x-0 bottom-0 z-50 flex flex-col shadow-[0_-2px_12px_rgba(0,0,0,0.12)]">
```

—
_Generated by Barber AI_

src/features/notifications/notificationHelpers.ts
```diff
@@ -0,0 +1,137 @@
...
+ * external and opens in a new tab; anything else is treated as an internal router path.
+ */
+export function parseNotificationLink(url: string | null | undefined): NotificationLink | null {
+	if (!url) { return null; }
+	const trimmed = url.trim();
+	if (!trimmed) { return null; }
+	// Protocol-relative (`//host`), or has an explicit scheme (`https:`, `mailto:`, …) → external.
+	if (trimmed.startsWith('//') || /^[a-z][a-z0-9+.-]*:/i.test(trimmed)) {
+		return { kind: 'external', href: trimmed };
+	}
```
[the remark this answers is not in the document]
dawsontoth: Fixed in 9c7698fa — `parseNotificationLink` now allowlists safe external schemes (`http`/`https`/`mailto`) and returns `null` for `javascript:`/`data:`/`vbscript:`/unknown schemes, so an unsafe URL can never reach an `href`. Added unit tests for the rejected schemes.

src/features/notifications/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+	const notifications = data ?? [];
+	const sorted = [...notifications].sort((a, b) => {
+		const stateA = STATE_ORDER[getWindowStatus(a, now).state];
+		const stateB = STATE_ORDER[getWindowStatus(b, now).state];
+		if (stateA !== stateB) { return stateA - stateB; }
+		const severityDelta = SEVERITY_ORDER[getSeverityConfig(a.type).severity]
+			- SEVERITY_ORDER[getSeverityConfig(b.type).severity];
+		if (severityDelta !== 0) { return severityDelta; }
+		// Most recently ending first within a group.
+		return (toMs(b.endAt) ?? Infinity) - (toMs(a.endAt) ?? Infinity);
```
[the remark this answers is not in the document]
dawsontoth: Fixed in 9c7698fa — the comparator now falls back to `Number.MAX_SAFE_INTEGER` instead of `Infinity`, so two open-ended notices compare as `0` rather than `NaN`.

src/features/notifications/notificationHelpers.ts
```diff
@@ -0,0 +1,137 @@
...
+export { BellIcon };
+
+/**
+ * Harper's `Date` scalar can serialise as an ISO string or an epoch-ms number (and numeric strings
+ * have been seen in the wild). Normalise any of those to epoch ms, or `null` when absent/unparseable.
+ */
+export function toMs(value: string | number | null | undefined): number | null {
+	if (value === null || value === undefined) { return null; }
+	if (typeof value === 'number') { return Number.isFinite(value) ? value : null; }
+	const trimmed = value.trim();
```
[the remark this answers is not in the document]
dawsontoth: Fixed in 9c7698fa — `toMs` now returns `null` for non-string/number inputs before calling `.trim()`; added a test covering boolean/object inputs.

src/components/NotificationBanner.tsx
dawsontoth: Fixed in 9c7698fa — the banner container is now `role="region"` with `aria-label="System notifications"` and `aria-live="polite"`, so screen readers announce active notices.

DavidCockerill (commented):
Really nice piece of work — this is a lot of surface area and the shape is right throughout. Non-blocking overall; the subscription pair below is the one thing I'd want fixed before it ships, and the two halves need fixing together.

**Two things I checked hard and found sound**, worth stating since they're the scary parts of a feature like this:

- **The authorization isn't presentational.** `AdminShell.tsx:14`'s `<Navigate to="/" />` is UI-only, but the server enforces independently at `central-manager/src/resources/SystemStatus.js:9-19` (`isFabricAdmin(context)` on create/update/delete) *and* at `roles.yaml:96` (`insert/update/delete: false` for the standard role). `isFabricAdmin` **throws** on a missing user rather than returning falsy, so it fails closed. A normal user POSTing directly gets rejected. Your "no central-manager change needed" claim is accurate.
- **The `datetime-local` ↔ UTC round trip is correct** — this is the bug I most expected to find. `toInputValue` reads with local getters; `fromInputValue` calls `Date.parse` on an offset-less date-time string, which ES2015+ specifies as local, then `.toISOString()`. Same zone both directions, lossless.

## Significant

**1 & 2 — the subscription has two opposite failure modes sharing one root cause** (`NotificationsSubscriptionManager.tsx:30` and `:44`). `failures` is reset by a successful *handshake*, not a *healthy session*.

**Silent permanent give-up.** Backoff sums to 2+4+8+16+32 = 62s, then `if (failures > 5) return;` — and because `StudioCloud` never unmounts, the effect never re-runs. The socket is dead for the rest of the session, with no log line.

*Plainly:* one wifi hiccup and the live feed is dead in that tab until reload. Nothing tells the user, nothing tells us — it silently degrades to the 60s poll. When someone reports "the maintenance banner took a minute to appear," there's no signal anywhere pointing at the cause.

**Unbounded reconnect loop** — the mirror image. If the server accepts the upgrade then closes (post-handshake `notAllowed()`, an edge dropping non-conforming upgrades, a CM restart loop), `onopen` zeroes `failures` before every `onclose`, so it can never exceed 1 and the five-attempt cap is **dead code**. Every open tab — including tabs on the sign-in page belonging to people who never logged in — retries every 2s forever. *(Credit: Codex found this half; I verified it.)*

Please note the interaction: **the trigger for #2 is exactly the kind of server-side hardening someone might add later**, so fix this before anyone tightens the CM side.

Suggested: cap the *backoff*, not the attempts (the `Math.min(…, 30_000)` already does that); only zero `failures` after the socket has stayed open ~10s; re-arm on `window.addEventListener('online', …)`; and `console.warn` on give-up so it's greppable in a RUM session.

**3 — an unparseable `endAt` means "never expires"** (`notificationHelpers.ts:86`). `toMs` returns `null` for *both* "absent" and "present but unparseable", so an unreadable field short-circuits the expiry check — and `getWindowStatus` compounds it by labelling the row a confident `'Active'`. Trigger: a `SystemStatus` row in a shape `toMs` doesn't handle (a `Date` object, a `{$date: …}` wrapper, a non-ISO locale string) — most plausibly created outside Studio, e.g. by the CM integration suite or by hand.

*Plainly:* a "maintenance complete at 3am" banner sits on every user's screen indefinitely, labelled Active, until an admin deletes the row by hand. Users can dismiss it — but only per-device, so it's back on their phone.

Worth making `toMs` distinguish the two cases and having `isActive` fail **closed** on unparseable. The instinct is already in this file — the severity mapping defaults *up* to `critical` — it just isn't applied to the time window. One case in the existing `isActive` describe block at `notificationHelpers.test.ts:41` covers it.

**4 — `NotificationBell.test.tsx` mocks away what it's testing** (`:16`). With both `hooks` and `acks` mocked, the only production code under test is the bell's JSX. `useActiveNotifications` and `useUnackedActiveNotifications` return the *same* array, so "badge shows 2" asserts the mock, not the unread computation. And `useNotificationAcks` returning an empty `Set` means the acked-first branch of the `SEVERITY_ORDER` sort (`NotificationBell.tsx:19`) is never exercised — nothing in the PR sorts a mixed acked/unacked list.

*Plainly:* this test would still pass if the unread-count logic were replaced with "however many notifications exist."

Cheap upgrade: mock only the *query* (or seed the real `queryClient`) and let the hooks run. Worth saying that `notificationHelpers.test.ts` is the opposite and genuinely good — real functions, real-shaped `sta-` fixtures, boundary cases, XSS rejections. That file is the model.

## Non-blocking

- **`SEVERITY_ORDER` is defined twice, identically** (`NotificationBell.tsx:19`, `index.tsx:17`). It belongs next to `SEVERITY_CONFIG` in `notificationHelpers.ts`, exported once — two copies of an ordering rule drift the moment someone adds a fourth severity, and then the bell and the center disagree about which notice is most serious.
- **`readAcks()` trusts the shape, not just the JSON** (`acks.ts:18`). `safeParse` handles malformed JSON correctly — good reuse — but valid JSON of the wrong shape (`{}`) flows through the `string[]` cast, and `[...readAcks(), id]` / `new Set(data)` then throw `TypeError: not iterable`, taking down bell, banner *and* center at once since all three are globally mounted. `Array.isArray(v) ? v.filter(x => typeof x === 'string') : []` closes it. *(Codex flagged this too.)*
- **`ackNotification` can throw on a full localStorage** (`acks.ts:39`). `writeAcks` persists *before* `queryClient.setQueryData`, so a `QuotaExceededError` throws out of an `onClick` and the notice doesn't even disappear in-memory. Swap the order so a quota failure degrades to "ack didn't survive reload" rather than "dismiss is broken." Low likelihood, but Safari private mode has historically thrown on first write. *(Codex too.)*
- **Nothing prunes acks for deleted notifications** (`acks.ts:39`) — slow growth (~12 bytes/id), so hygiene rather than a real quota risk; a two-line intersection on query success makes it self-limiting. **The per-browser scoping I'd leave alone** — you state it as a deliberate v1 trade with a named successor, and being re-nagged once per device is a fair price for shipping without a CM schema change. One line in the PR body confirming that's deliberate would stop it being re-litigated. *(Codex flagged the pruning.)*
- **The admin form accepts a link it will never render** (`index.tsx:248`). `submit()` validates the message and start/end ordering but passes `url` through unchecked — type `javascript:alert(1)` or a typo'd `htp://…` and it saves happily, then `parseNotificationLink` returns `null` and the link silently never appears. Success toast, broken notice. Call `parseNotificationLink(url)` in `submit` so the form can't drift from the renderer.
- **`new WebSocket(url)` is unguarded** (`:30`) — the constructor throws synchronously on a malformed URL, inside a `useEffect` with no `try`. `VITE_CENTRAL_MANAGER_API_URL` is absolute in every checked-in env file so it's fine today, but a relative or `//`-prefixed value in a future deploy config would throw out of the effect on a page anonymous users load.
- Following on from cb1kenobi's banner note: `aria-live="polite"` is right for info/maintenance, but your severity model has a `critical` tier — consider `assertive` when any strip in the stack is critical. That was in the original Barber AI comment; only the `polite` half landed.

## Verified and dropped

Recording these so nobody re-runs the same checks: the admin gate is **not** client-side-only (above); the `datetime-local`↔UTC round trip is correct (above); expiring notices **do** clear within 30s (`useNow(30_000)` re-renders and `useActiveNotifications` re-filters); there's **no** StrictMode double-subscribe (empty dep array, teardown sets `cancelled`, the stale closure's `onclose` early-returns, `close()` during `CONNECTING` is legal); and mutations **do** surface errors via the global `MutationCache.onError`. I also tried several XSS variants through the link and message fields and found no bypass.

One scoping note for the record: **CM#483 does not cover `SystemStatus`** — that guard was scoped to resources declaring `static loadAsInstance = false`, which `SystemStatus` doesn't. So the anonymous WS subscribe here works by *omission* rather than by decision. Not a leak — the table is public-read and carries admin-authored broadcast content — but worth making that permissiveness explicit somewhere rather than incidental, since the next person to read `SystemStatus.js` alongside #483 will wonder.

Performance is a non-blocking note: four independent 30s clocks where one shared tick would do, and an unmemoized sort in the bell that runs on every render of a component present on every page. Both cheap at this scale.

*Cross-model note: ran this by Codex, whose findings are incorporated above and credited inline. The Gemini leg timed out at 240s on the 1539-line diff — not an auth failure, just size.*

— Reviewed by DAIvid (Claude Opus 5), cross-checked with Codex

dawsontoth:
Thanks for the exceptionally thorough pass — the subscription analysis especially. All addressed in `d04ce324` (pushed here and to `dev`).

**Blocking — the subscription pair (`NotificationsSubscriptionManager.tsx`).** Fixed both halves at the shared root you identified (`failures` was keyed to the handshake, not a healthy session):
- `failures` now resets only after the socket stays open ~10s, so an accept-then-close server backs off to the 30s cap instead of retry-storming every 2s.
- We cap the **backoff**, not the attempt count — it keeps retrying at 30s rather than giving up permanently, so one wifi hiccup can't strand the tab on the poll backstop for the session.
- Added a `window` `online` re-arm, a `try/catch` around `new WebSocket`, and a one-time `console.warn` when it goes unstable so it's greppable in RUM.

**Significant:**
- **Fail-closed expiry** — new `parseInstant` distinguishes *absent* from *unparseable*; `isActive` now returns `false` and `getWindowStatus` reports “Schedule unavailable” for an unreadable bound, so a bad `endAt` can't pin a notice on screen forever. Covered in the `isActive` describe block as you noted.
- **Bell test** now mocks only the query and runs the real hooks/ack store/sort — it asserts the expired notice is filtered out, that an ack drops the count, and that un-acked sorts ahead of acked. (`notificationHelpers.test.ts` stays the model.)

**Non-blocking, also done:** `readAcks` validates the array-of-strings shape (wrong-shape JSON can no longer take down all three surfaces); `writeAcks` updates the cache before persisting so a full localStorage degrades to “didn't survive reload”; the admin form validates the link via `parseNotificationLink` before saving; the banner uses `aria-live="assertive"` when any active strip is critical; `SEVERITY_ORDER` is exported once from `notificationHelpers`; one shared 30s clock replaces the four; the center sort is memoized.

**Consciously deferred:**
- **Ack pruning for deleted notifications** — left as a follow-up; it's ~12 bytes/id and the new shape-guard removes the crash path, so it's hygiene rather than a quota risk.
- **Per-browser ack scoping** — deliberate v1 trade (noted in the PR body); the named successor is a server-side per-user read-state table once org/user scoping lands.
- **CM#483 / anonymous subscribe** — documented inline in `NotificationsSubscriptionManager.tsx` (public-read broadcast content, permissive by omission not decision); a matching note in `central-manager`'s `SystemStatus.js` is a good separate-repo follow-up.
- **Perf** — shared clock and memoized center sort landed; the bell sort was already memoized.

src/features/notifications/NotificationsSubscriptionManager.tsx
```diff
@@ -0,0 +1,113 @@
...
+				socket = null;
+				scheduleReconnect();
+			};
+		}
+
+		// Connectivity came back — reconnect now instead of waiting out the backoff.
+		const onOnline = () => {
+			if (cancelled || socket) { return; }
+			if (reconnectTimer) { clearTimeout(reconnectTimer); }
+			failures = 0;
```
DavidCockerill: Non-blocking. The `online` handler bypasses the new backoff unconditionally: `failures = 0` plus an immediate `connect()`. In a fake-timer harness, firing `online` five times with a failure between each produced five connects at the same millisecond — the 30s ceiling this commit just installed doesn't apply on this path.

Much narrower than what it replaced (browsers fire `online` on real interface transitions, not continuously — a handful on captive-portal/VPN churn, not a sustained storm), and I'd merge without it. But it's the one remaining unbounded-connect path in a file whose whole purpose this round was bounding connects.

Cheapest guard: keep the `clearTimeout` + `connect()`, drop the `failures = 0` — or ignore the event if the last attempt was under ~2s ago.

src/features/notifications/NotificationsSubscriptionManager.tsx
```diff
@@ -0,0 +1,113 @@
...
+			const delay = Math.min(1000 * 2 ** failures, MAX_BACKOFF_MS);
+			if (delay === MAX_BACKOFF_MS && (1000 * 2 ** (failures - 1)) < MAX_BACKOFF_MS) {
+				console.warn(
+					'[notifications] SystemStatus subscription unstable; retrying at 30s (data still polls every 60s)',
+				);
+			}
+			reconnectTimer = setTimeout(connect, delay);
+		};
+
+		function connect() {
```
DavidCockerill, line 68: Non-blocking. Coverage here is still 11.86% statements / 9.09% branches — unchanged from last round, while the logic got materially more intricate (healthy-session timer, cap-crossing warn condition, `online` re-arm, `try`/`catch` construct path). This is now the only significant uncovered surface in the PR, and it's the file that produced both round-1 significants.

It's also demonstrably cheap to cover: I verified this commit with a ~45-line jsdom test using `vi.useFakeTimers()` and a fake `WebSocket` class, and it passes against this head — happy to hand it over rather than have you rewrite it. Three assertions lock in everything this commit fixed: after 6 consecutive failures a 7th socket is still created (no give-up), an accept-then-close cycle escalates the delay instead of staying at 2s, and a >10s session resets it.

DavidCockerill (approved):
Approving at `d04ce32`. One commit fixed all four significants **plus every non-blocking item** — nice.

The part I most wanted to check: the subscription pair was fixed **together at the shared root**, not one half at a time. Verified with a fake-timer WebSocket harness — 14 reconnects over 5 minutes in *both* the offline case (was 5-then-permanently-silent) and the accept-then-close case (was ~150). Fixing either half alone would have made the other worse, so that mattered.

Also verified: an unparseable `endAt` now fails closed, and the bell test survives three mutations — it's a real test now rather than an assertion about its own mocks. Suite green, `tsc` clean, CI passing.

Two non-blocking threads. The second comes with a ~45-line test I already wrote and ran against this head — say the word and it's yours rather than something you have to write.

src/features/notifications/NotificationsSubscriptionManager.tsx
dawsontoth: Fixed in 668c9941 — dropped the `failures = 0` (an `online` event now skips the remaining wait but keeps the escalation) and added a 2s floor between online-triggered attempts, so a burst collapses to one connect. Both halves are locked in by tests: one asserts five `online` events with a failure between each produce a single socket, another that the next failure after an online reconnect waits 8s rather than restarting at 2s.

src/features/notifications/NotificationsSubscriptionManager.tsx
dawsontoth: Fixed in 668c9941 — wrote the jsdom suite rather than take the handoff: fake timers + a stub WebSocket, now 93.65% statements / 87.5% branches on the file (from 11.86% / 9.09%). Covers your three plus the newer paths — no-give-up after 6 failures, accept-then-close escalation, healthy-session reset, the online debounce and its non-reset, a throwing constructor, unset base URL, and unmount teardown. Each assertion was mutation-checked: re-adding `failures = 0`, dropping the floor, restoring the give-up cap, or resetting on the bare handshake each fails exactly one test (the first version of my burst test passed under mutation — it was only exercising the already-connected guard, so I rewrote it).

PR #1553, opened 2026-07-24
test(e2e): add Playwright integration-test suite

dawsontoth:
Adds the Playwright end-to-end integration-test suite for studio (`e2e/`), plus a one-line
`vitest.config.ts` exclude so the app's unit tests don't collect these specs.

## What this is

- A **standalone** test project under `e2e/` — its own `package.json` / lockfile / `pnpm-workspace.yaml`,
  deliberately NOT part of the app's pnpm workspace, so its deps + browser downloads stay out of the
  app build. Holds only narrow, disposable test credentials, never `.env.local`.
- Specs drive the real UI to catch regressions: the sign-in page + client-side validation, the
  email-verification screens, an org-users list, and a full **signup → email round-trip (Mailosaur)
  → verify → login** flow.
- Naming picks the project: `*.anon.spec.ts` (no session) vs `*.authed.spec.ts` (logged in, via a
  saved cookie storageState from `auth.setup.ts`).
- Run locally against your dev server (`cd e2e && pnpm test:local:ui`) or headless against deployed
  dev (`pnpm test`). Screenshot baselines come from the Linux container (`docker compose run`).

## Why it's shaped this way

The **autonomous control plane** that runs these tests — daily, on merge-to-`stage`, and on PRs
(triaging failures into issues, sandboxing untrusted PR code) — lives in a **separate repo**,
[studio-e2e-harness](https://github.com/HarperFast/studio-e2e-harness), deliberately kept out of this
repo so a studio PR can't alter what decides whether or how its own code is run. The **specs live
here**, co-located with the app for local dev; both lanes exercise these same files.

Merging this to `stage` is what lets the trusted lane (and, later, the PR lane) run against a base
branch that actually contains the suite.

## Notes

- `e2e/README.md` documents running it + the conventions worth knowing (hash routing, cookie auth,
  Sonner toasts, link/token email verification, the Mailosaur round-trip, screenshot baselines).
- `.env.e2e` (real creds) is gitignored; `.env.e2e.example` shows the shape. Signup is
  domain-allowlisted server-side (`ALLOWLIST_EMAIL_DOMAINS`), so the round-trip spec skips itself
  cleanly if the Mailosaur domain isn't allowlisted on the target env.


e2e/tests/org-users.authed.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+		await page.goto(`/#/${orgId}/users`);
+
+		// The account may belong to the org but lack users-view permission, or the org
+		// route may 403 on this environment. Wait for EITHER the table or the app's error
+		// boundary, and skip cleanly on the latter rather than failing on a missing table.
+		const table = page.getByRole('table');
+		const errorBoundary = page.getByText('Component Error');
+		await expect(table.or(errorBoundary).first()).toBeVisible();
+		test.skip(
+			await errorBoundary.isVisible().catch(() => false),
```
cb1kenobi: **Medium: generic error boundary is treated as a permissions skip, which can mask a real regression**

`ErrorComponent` renders the title `Component Error` for *any* uncaught error on the route (`src/components/ErrorComponent.tsx:27` — the title is optional and defaults generically; it isn't specific to 403/permission failures). This spec treats the presence of that text as proof the account merely lacks `users-view` permission and skips instead of failing. If the org-users page actually breaks (a JS exception, a broken query, an API error) for every account on the target env, this test will silently skip rather than turn red — which defeats the point of an e2e regression suite covering exactly this page.

Suggested fix: distinguish the permission case more precisely, e.g. check the underlying response status (403) via `page.waitForResponse`/`page.on('response')` for the org-users API call, or match specific error copy for the permission case, rather than skipping on the generic `Component Error` boundary text alone.

—
_Generated by Barber AI_

e2e/Dockerfile
```diff
@@ -0,0 +1,24 @@
+# Playwright runner. Node pinned to the repo's .nvmrc (24.18.0); Chromium's Linux
+# build + system deps are installed to match the resolved Playwright version, so
+# screenshot baselines are stable and reproducible (generate baselines HERE, not
+# on macOS). Multi-arch base — builds natively on Apple Silicon (arm64).
+FROM node:24.18.0-bookworm
```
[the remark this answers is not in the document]
dawsontoth: @gemini by not checking the registry itself, you have produced a hallucination by your dated training information.

e2e/tests/signup-verification.anon.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+
+		try {
+			await page.waitForURL((url) => !url.hash.includes('/sign-in'), { timeout: 30_000 });
+			await expect(page).not.toHaveURL(/#\/verifying/);
+			await expect(page.getByLabel('Sign Out')).toBeVisible();
+		} finally {
+			// Delete the throwaway account so the target env doesn't accumulate test users.
+			await deleteThrowawayAccount(page);
+		}
+	});
```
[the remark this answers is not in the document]
dawsontoth: Done — wrapped the whole signup→verify→login flow in `try/finally`, so the throwaway account is cleaned up on any failure after signup (not just the final assertions). Caveat noted in code: self-delete needs a logged-in session, so a failure *before* login has nothing to delete (best-effort no-op). Pushed in e9773851.

e2e/tests/org-users.authed.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+
+		const hrefs = await page.locator('a[href*="#/"]').evaluateAll((els) =>
+			els.map((e) => e.getAttribute('href') ?? '')
+		);
+		for (const href of hrefs) {
+			const m = href.match(/#\/([^/?#]+)/);
+			if (m && !RESERVED.has(m[1])) { return m[1]; }
+		}
+		await page.waitForTimeout(500);
+	}
```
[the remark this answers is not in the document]
dawsontoth: Done — `resolveOrgId` now uses `expect.poll` instead of the manual `waitForTimeout` loop. Note: the suggested snippet returned the assertion result rather than the org id, so I capture the value in the polled closure and return that. Pushed in e9773851.

e2e/tests/org-users.authed.spec.ts
dawsontoth: Good catch — fixed. The spec now watches for a real 403 during the org-users load: it skips only on a genuine permission 403, and fails red if the `Component Error` boundary appears with no 403, so a real break can no longer hide as a permission skip. Pushed in e9773851.

e2e/Dockerfile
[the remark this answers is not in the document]
dawsontoth: Confirmed valid (and thanks for the follow-up retraction) — `node:24.18.0-bookworm` is the repo pinned Node (`.nvmrc`) and a real image tag. No change.

e2e/package.json
```diff
@@ -0,0 +1,30 @@
...
+		"install:browser": "playwright install --with-deps chromium",
+		"update-snapshots": "playwright test --grep @visual --update-snapshots",
+		"typecheck": "tsc --noEmit -p tsconfig.json"
+	},
+	"devDependencies": {
+		"@playwright/test": "^1.55.0",
+		"@types/node": "^24.0.0",
+		"dotenv": "^17.0.0",
+		"mailosaur": "^8.0.0",
+		"typescript": "~7.0.0"
```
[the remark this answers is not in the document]
dawsontoth: These are current, intentional versions — not hallucinated. The studio app is on TypeScript 7.x and `@types/node` 24 (matching Node 24); `pnpm install` resolves them and the suite installs + runs. Downgrading to TS 5.7 / types 20 would diverge from the app, so leaving as-is.

e2e/pnpm-lock.yaml
```diff
@@ -0,0 +1,339 @@
+lockfileVersion: '9.0'
```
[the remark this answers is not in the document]
dawsontoth: The lockfile is valid — generated by `pnpm install` against the versions above and it installs cleanly (verified in the sandbox). `typescript@7.0.2` / `@types/node@24.13.3` exist and match the app, so no regeneration is needed.

e2e/tests/signup-verification.anon.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+ * verification email via Mailosaur → follow its link → confirm the account is
+ * verified by signing in without being bounced back to /#/verifying.
+ *
+ * Runs in the `anon` project (starts logged-out; it creates and then logs into a
+ * brand-new account within the test). Skips cleanly until Mailosaur is configured.
+ *
+ * NOTE: each run creates a real account on the target (dev) environment — see the
+ * account-churn note in README before wiring this into a daily/CI loop.
+ */
+const PASSWORD = 'SuperSecret123!';
```
dawsontoth: Can we randomize this password for each run?

dawsontoth: Done — replaced the static `SuperSecret123!` with a per-run `randomPassword()` (randomUUID-based; always upper + lower + digit + symbol, ~22 chars), generated inside the test alongside the throwaway email. No static credential in the repo, and each run (including retries) gets a fresh one. Pushed in 1db3ea30.

e2e/docker-compose.yml
```diff
@@ -0,0 +1,23 @@
+# Canonical, reproducible test environment. Use this to generate/compare
+# screenshot baselines (they must come from this Linux Chromium, not macOS).
+#
+#   docker compose run --rm e2e                      # run the suite
+#   docker compose run --rm e2e pnpm update-snapshots  # (re)generate baselines
```
dawsontoth, line 5: Let's stick these into the top level package.json so that it's easy to run them locally in the proper way. Can we warn people if they try to run the e2e or update-snapshots outside of a container on their raw machine?

dawsontoth: Done — surfaced these at the repo root in `package.json` (779e8e50):

```
npm run test:e2e:docker      # run the suite in the canonical Linux container
npm run test:e2e:snapshots   # (re)generate the *-linux screenshot baselines
```

So the canonical container runs (and baseline generation) are discoverable without `cd e2e` or remembering the compose invocation. Also noted in `e2e/README`.

e2e/tests/org-users.authed.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+			!orgId,
+			'No accessible organization for this account — add it to an org, or set PLAYWRIGHT_ORG_ID.',
+		);
+
+		// Watch for a 403 during the org-users load. That — NOT the generic error boundary — is the
+		// real "no permission" signal. The app's `Component Error` boundary renders for ANY uncaught
+		// error on the route, so skipping on its text alone would hide a genuine break (JS exception,
+		// broken query, API error) on exactly the page this spec exists to cover.
+		let saw403 = false;
+		page.on('response', (r) => {
```
cb1kenobi, line 52: **Low: the 403 listener isn't scoped to the org-users request, so an unrelated 403 elsewhere on the page can still mask a real crash**

`page.on('response', ...)` flags `saw403` for *any* response on the page during this navigation, not specifically the org-users data fetch (`getOrganizationRolesQueryOptions` → `GET /OrganizationRole/{orgId}`, which also refetches every 10s per its `refetchInterval`). If any other request 403s during the same window — the periodic refetch racing a truly unrelated 403 elsewhere, a feature-flag check, telemetry — `saw403` goes true even though the org-users API itself is fine, masking a genuine render crash on this page. That's a narrower version of exactly the failure mode this fix set out to close. `signup-verification.anon.spec.ts` in this same PR already establishes the more precise pattern (`page.waitForResponse((r) => r.url().includes('/User/') && ...)`); this spec could scope the same way.

Suggested fix:
```suggestion
		let saw403 = false;
		page.on('response', (r) => {
			if (r.status() === 403 && r.url().includes('/OrganizationRole/')) { saw403 = true; }
		});
```

—
_Generated by Barber AI_

dawsontoth: Good catch — fixed in 0ca0744a. Scoped the listener to `r.url().includes('/OrganizationRole/')`, the org-users page's only suspense query (`getOrganizationRolesQueryOptions` → `GET /OrganizationRole/{orgId}`; the user list is derived from its response). So an unrelated 403 — telemetry, a feature-flag check, or that query's 10s refetch racing something else — no longer flips `saw403` and masks a real render crash. Matches the precise pattern the signup spec uses.

cb1kenobi (approved):
LGTM!

PR #503, opened 2025-07-15
[STUDIO-176] Confirm Before Deleting Databases and Tables

dawsontoth:
I extracted the cluster deletion modal, made it generic, and then used it on the browse page.

https://harperdb.atlassian.net/browse/STUDIO-176

PR #1559, opened 2026-07-24
feat(cluster): cluster plan usage card + Usage tab [#1297]

### LINKED ISSUE #1297: Usage remaining display in the UI

dawsontoth:
## Cluster usage-vs-plan display (#1297)

Adds a **Plan usage** card to the managed cluster overview and a **Usage** tab, backed by the central-manager endpoint from **HarperFast/central-manager#503** (`GET /Cluster/:id/usage`, approved and merged).

Closes #1297.

### Verified against live data
The endpoint is deployed to dev + stage, and both surfaces render real cluster data (screenshots below): actual plan name and renewal date, real limits, and real storage sourced from `InstanceMonitoring`.

### Model
Quota is enforced **per region**, never cluster-wide (exhaustion, renewal, billing and the signed license ceiling are all per-region), so the UI meters per region rather than summing into one misleading bar.

### Overview card
- **Single region** → the four headline meters (Reads / Writes / Storage / Compute).
- **Multiple regions** → the single **most-constrained** region×metric, so a hot region can't hide behind an average.

### Usage tab
- One **collapsible card per region**, with status driving the header:
  - **active** → "renews {date}"
  - **exhausted** (burned through a counter cap → re-billing) → amber **"Cycle exhausted"**
  - **lapsed** (expired / not renewed, under cap) → **"No active license"**, collapsed by default
- All **8 metered metrics** as used/limit bars.
- **Rate limits** (per-minute counts *and* bandwidth) + **per-instance resources** in a single shared **"Plan limits & resources"** card when uniform across regions; per-region when they differ.
  These are the **effective per-region** ceilings, not the raw plan row: the endpoint resolves the `-1` sentinel and scales the tier-dependent ones by `Region.purchasedBlockMultiplier` (HarperFast/central-manager#635), so each reads as a number, **Unlimited**, or **"—"**.
- Region ids and plan id shown subtly (monospace, right-aligned) in the card headers.

### Meter states
Finite ceiling → used/limit bar, amber at ≥90%. Plan limit `-1` → hatched **"Unlimited"**. Unresolved plan → **"—"** (never the reassuring "Unlimited" — this is a billing surface). A zero/negative limit can't produce `NaN`/`Infinity`.

### Completeness
Cross-checked field-by-field against the new-cluster plan modal: everything it shows is here, plus used-vs-limit rather than limits alone. Only Cloud Instance Type is out of scope (needs a CM field; dedicated/GPU only).

### Tests
36 cases across the three usage surfaces (`vitest`, jsdom): the meter's three ceiling states and guards, the card's single- vs multi-region branches, its four hide conditions and its subtitle-uniformity rules, and the tab's collapse/expand, status badges, shared-vs-per-region plan info, rate-ceiling states, and empty/error/loading states. Cluster-feature suite **119 passing**, full suite **1890 passing**. `tsc`, `oxlint`, `dprint` clean.

### Depends on
HarperFast/central-manager#635 — the endpoint change that makes `rateLimits` effective per-region values. This branch degrades safely without it (ceilings render **"—"** rather than an understated per-block number), so the two can ship in either order, but real ceilings appear only once #635 is deployed.

### Follow-up
Regenerate the SDK types from #503's OpenAPI to drop the hand-typed response (works fine as-is; a regen is a large unrelated diff, so it belongs on its own).

### DIFF
--- a/src/features/cluster/components/ClusterUsageCard.tsx
+++ b/src/features/cluster/components/ClusterUsageCard.tsx
@@ -1,24 +1,28 @@
 import { fmtBytes, fmtCount, fmtHours, UsageMeter, UsageMetric } from '@/features/cluster/components/UsageMeter';
+import { usageSubtitle, useClusterUsage } from '@/integrations/api/cluster/getClusterUsage';
 import { Link } from '@tanstack/react-router';
 import { ArrowRight } from 'lucide-react';
 
-// ⚠️ MOCKUP — placeholder data, not wired to an API yet (issue #1297).
-// Shape mirrors the planned `GET /Cluster/:id/usage` response so this drops
-// straight onto real data later. Framing per product decision: "used X of Y
-// this cycle" (blocks auto-renew & re-bill on exhaustion for paid tiers; for
-// the free tier the same bar doubles as the hard-limit warning).
+// Plan-usage summary on the managed cluster overview (issue #1297). Framing is "used X of Y this
+// cycle": purchased blocks auto-renew and re-bill on exhaustion for paid tiers, while for the free
+// tier the same bar doubles as the hard-limit warning. Shows only the four headline metrics; the
+// full breakdown lives on the Usage tab.
 
-// Sample data — a mid-usage "Standard" cluster, with Reads pushed near the cap
-// to show the warning state. The overview card shows only the four headline
-// metrics; the full list lives on the Usage tab.
-const SAMPLE: UsageMetric[] = [
-	{ label: 'Reads', used: 9_200_000, limit: 10_000_000, format: fmtCount },
-	{ label: 'Writes', used: 2_300_000, limit: 5_000_000, format: fmtCount },
-	{ label: 'Storage', used: 13_314_398_617, limit: 21_474_836_480, format: fmtBytes },
-	{ label: 'Compute', used: 1.6, limit: 2, format: fmtHours },
-];
+export function ClusterUsageCard({ clusterId, base }: { clusterId: string; base: string }) {
+	const { data } = useClusterUsage(clusterId);
 
-export function ClusterUsageCard({ base }: { base: string }) {
+	// Nothing to show until there's an active plan with recorded usage (also covers loading, errors,
+	// and self-hosted clusters, which report no totals).
+	if (!data || data.selfManaged || !data.totals) { return null; }
+
+	const { totals } = data;
+	const metrics: UsageMetric[] = [
+		{ label: 'Reads', used: totals.reads.used, limit: totals.reads.limit, format: fmtCount },
+		{ label: 'Writes', used: totals.writes.used, limit: totals.writes.limit, format: fmtCount },
+		{ label: 'Storage', used: totals.storageBytes.used, limit: totals.storageBytes.limit, format: fmtBytes },
+		{ label: 'Compute', used: totals.cpuTimeHours.used, limit: totals.cpuTimeHours.limit, format: fmtHours },
+	];
+
 	return (
 		<section className="mt-6 rounded-2xl border border-border bg-card p-5">
 			<div className="flex items-center justify-between gap-3">
@@ -30,12 +34,10 @@ export function ClusterUsageCard({ base }: { base: string }) {
 					View all usage <ArrowRight className="size-3" />
 				</Link>
 			</div>
-			<p className="mt-0.5 text-xs text-muted-foreground">
-				Standard plan · this cycle renews Aug 12 · updated 4 min ago
-			</p>
+			<p className="mt-0.5 text-xs text-muted-foreground">{usageSubtitle(data)}</p>
 
 			<div className="mt-4 grid grid-cols-1 gap-x-8 gap-y-4 sm:grid-cols-2">
-				{SAMPLE.map((m) => <UsageMeter key={m.label} {...m} />)}
+				{metrics.map((m) => <UsageMeter key={m.label} {...m} />)}
 			</div>
 		</section>
 	);
--- a/src/features/cluster/components/ClusterUsageCard.tsx
+++ b/src/features/cluster/components/ClusterUsageCard.tsx
@@ -54,16 +54,14 @@ function MultiRegion({ data }: { data: ClusterUsage }) {
 	const mc = data.mostConstrained;
 	return (
 		<div>
-			<p className="mb-3 text-xs text-muted-foreground">
-				{data.regions.length} regions — usage is metered per region.
-			</p>
+			<p className="mb-1.5 text-xs text-muted-foreground">Most constrained of {data.regions.length} regions</p>
 			{mc
 				? (
 					<UsageMeter
 						{...toMeter(
 							mc.metric,
 							{ used: mc.used, limit: mc.limit, unlimited: false, limitKnown: true },
-							`Most constrained · ${mc.region} ${METRIC_LABEL[mc.metric]}`,
+							`${mc.region} · ${METRIC_LABEL[mc.metric]}`,
 						)}
 					/>
 				)
--- a/src/features/cluster/components/ClusterUsageCard.tsx
+++ b/src/features/cluster/components/ClusterUsageCard.tsx
@@ -61,7 +61,9 @@ function MultiRegion({ data }: { data: ClusterUsage }) {
 						{...toMeter(
 							mc.metric,
 							{ used: mc.used, limit: mc.limit, unlimited: false, limitKnown: true },
-							`${mc.region} · ${METRIC_LABEL[mc.metric]}`,
+							// `region` is nullable on the endpoint — fall back the same way the Usage tab does
+							// rather than interpolating a literal "null" into a customer-facing label.
+							`${mc.region ?? 'Region'} · ${METRIC_LABEL[mc.metric]}`,
 						)}
 					/>
 				)
--- a/src/features/cluster/components/ClusterUsageCard.tsx
+++ b/src/features/cluster/components/ClusterUsageCard.tsx
@@ -1,27 +1,19 @@
-import { fmtBytes, fmtCount, fmtHours, UsageMeter, UsageMetric } from '@/features/cluster/components/UsageMeter';
-import { usageSubtitle, useClusterUsage } from '@/integrations/api/cluster/getClusterUsage';
+import { METRIC_LABEL, toMeter, UsageMeter, UsageMetric } from '@/features/cluster/components/UsageMeter';
+import { ClusterUsage, usageSubtitle, useClusterUsage } from '@/integrations/api/cluster/getClusterUsage';
 import { Link } from '@tanstack/react-router';
 import { ArrowRight } from 'lucide-react';
 
-// Plan-usage summary on the managed cluster overview (issue #1297). Framing is "used X of Y this
-// cycle": purchased blocks auto-renew and re-bill on exhaustion for paid tiers, while for the free
-// tier the same bar doubles as the hard-limit warning. Shows only the four headline metrics; the
-// full breakdown lives on the Usage tab.
+// Plan-usage summary on the managed cluster overview (issue #1297). Quota is per region, so:
+//   • one region  → the four headline meters for that region.
+//   • many regions → the single most-constrained region×metric (an average would hide a hot region),
+//     with the full per-region breakdown a click away on the Usage tab.
 
 export function ClusterUsageCard({ clusterId, base }: { clusterId: string; base: string }) {
 	const { data } = useClusterUsage(clusterId);
 
-	// Nothing to show until there's an active plan with recorded usage (also covers loading, errors,
-	// and self-hosted clusters, which report no totals).
-	if (!data || data.selfManaged || !data.totals) { return null; }
-
-	const { totals } = data;
-	const metrics: UsageMetric[] = [
-		{ label: 'Reads', used: totals.reads.used, limit: totals.reads.limit, format: fmtCount },
-		{ label: 'Writes', used: totals.writes.used, limit: totals.writes.limit, format: fmtCount },
-		{ label: 'Storage', used: totals.storageBytes.used, limit: totals.storageBytes.limit, format: fmtBytes },
-		{ label: 'Compute', used: totals.cpuTimeHours.used, limit: totals.cpuTimeHours.limit, format: fmtHours },
-	];
+	// Nothing to show until there's a managed plan with reportable regions (also covers loading, errors,
+	// and self-hosted clusters).
+	if (!data || data.selfManaged || data.regions.length === 0) { return null; }
 
 	return (
 		<section className="mt-6 rounded-2xl border border-border bg-card p-5">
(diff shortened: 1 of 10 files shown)


dawsontoth:
Backend endpoint that will replace the placeholder data here: HarperFast/central-manager#503

src/features/cluster/components/UsageMeter.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+export const fmtCount = (n: number) =>
+	new Intl.NumberFormat(undefined, { notation: 'compact', maximumFractionDigits: 1 }).format(n);
+export const fmtHours = (n: number) => `${new Intl.NumberFormat(undefined, { maximumFractionDigits: 1 }).format(n)} hr`;
+export const fmtBytes = (n: number) => humanFileSize(n);
+
+export function UsageMeter({ label, used, limit, format }: UsageMetric) {
+	const unlimited = limit === null;
+	const pct = unlimited ? 0 : Math.min(100, Math.round((used / limit) * 100));
+	const warn = pct >= 90;
```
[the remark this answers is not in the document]
dawsontoth: Done — `unlimited` now also covers a negative limit, and the percentage calc is skipped when `limit === 0`, so the bar can't divide by zero. Pushed in 6a2279df.

dawsontoth:
Now verified end-to-end against **live stage data** — HarperFast/central-manager#503 is deployed to dev + stage, so this is no longer running on placeholders.

`GET /Cluster/:id/usage` on stage returns 200 with the expected shape, and both surfaces render it:

**Overview card** — real plan name and renewal (`Fabric Managed Service Block Level 0 plan · renews Dec 4`), real ceilings (10M reads, 2M writes, 2 hr compute), and real storage **4 GB / 5 GB (70%)** sourced from `InstanceMonitoring`.

**Usage tab** — the region card keyed on the real region (`us-1`) and plan (`fabric-block-level-0`), all 8 metered metrics against their real limits, and the shared `Plan limits & resources` card showing the plan's actual rate limits (1,000 reads/min, 1 MB read bandwidth/min, …).

Also in this push:
- **Rebased onto `stage`** (was 38 behind) — linear history, no merge commit, no conflicts. All UI files verified byte-identical to their pre-rebase state.
- **22 component tests added** now that the contract is locked: the meter's three ceiling states + its zero/negative-limit guards, the card's single- vs multi-region branches and all four hide conditions, and the tab's collapse/expand, exhausted-vs-lapsed badges, shared-vs-per-region plan info, and empty/loading states. Cluster-feature suite is **105 passing**; `tsc`/`oxlint`/`dprint` clean.
- Contract re-checked field-by-field against the merged endpoint — no drift.

Out of draft and ready for review.

src/features/cluster/components/ClusterUsageCard.tsx
```diff
@@ -0,0 +1,75 @@
...
+	return (
+		<div>
+			<p className="mb-1.5 text-xs text-muted-foreground">Most constrained of {data.regions.length} regions</p>
+			{mc
+				? (
+					<UsageMeter
+						{...toMeter(
+							mc.metric,
+							{ used: mc.used, limit: mc.limit, unlimited: false, limitKnown: true },
+							`${mc.region} · ${METRIC_LABEL[mc.metric]}`,
```
cb1kenobi: **Medium: `mc.region` renders literally as "null" when the most-constrained region has no name**

`MostConstrained.region` is typed `string | null` (same field as `ClusterUsageRegion.region`), and `Page.tsx`'s `RegionSection` guards the identical field with `region.region ?? 'Region'`. This label doesn't: `` `${mc.region} · ${METRIC_LABEL[mc.metric]}` `` interpolates `null` directly, so a null region here shows the customer-facing string "null · Reads" on the overview card. None of the tests exercise a null `mc.region`, so this wasn't caught.

Suggested fix:
```suggestion
							`${mc.region ?? 'Region'} · ${METRIC_LABEL[mc.metric]}`,
```

—
_Generated by Barber AI_

src/features/cluster/usage/Page.tsx
```diff
@@ -0,0 +1,259 @@
...
+	// Rate limits + per-instance resources are usually identical across a cluster's regions (same plan),
+	// so when they're uniform we show them once in a shared card instead of repeating them per region.
+	const shared = data && !data.selfManaged ? uniformPlanInfo(data.regions) : null;
+
+	return (
+		<ClusterContentWithSubNavMenu className="max-w-4xl pb-20">
+			<h1 className="text-2xl font-light text-foreground">Usage</h1>
+			{data?.selfManaged
+				? <Empty>Usage isn't tracked for self-hosted clusters — they run under their own license.</Empty>
+				: !data || data.regions.length === 0
```
cb1kenobi: **Medium: a fetch error on the Usage tab is shown as "no usage this cycle"**

`useClusterUsage` only destructures `data`/`isLoading`; there's no `isError` check. When the request fails, `data` is `undefined` and `isLoading` is `false`, so this branch renders `<Empty>No usage has been recorded for the current cycle yet.</Empty>` — telling the user they used nothing, when the real problem is the API call failed. This is exactly the kind of thing this surface exists to get right (it's a billing/quota view), and sibling analytics panels in this codebase (e.g. `src/features/instance/status/analytics/tabs/StorageTab.tsx`) do check `isError` and render a distinct error state instead of folding it into empty.

Suggested fix: destructure `isError` from `useClusterUsage(clusterId)` and render a distinct error message (e.g. "Couldn't load usage data — try again.") before falling through to the real empty-state check.

—
_Generated by Barber AI_

src/features/cluster/components/ClusterUsageCard.tsx
dawsontoth: Fixed in afd6fec5 — `mc.region ?? 'Region'`, matching how `RegionSection` already guards the identical field, plus a note on why the fallback is there.

Added a regression test (`falls back to a generic region label when the most-constrained region is unnamed`). Confirmed it reproduces the bug against the pre-fix source — it failed printing exactly `null · Reads`:

```
× falls back to a generic region label when the most-constrained region is unnamed
  Unable to find an element with the text: Region · Reads
     null · Reads
```

It also asserts no literal `null` appears anywhere in the card.

src/features/cluster/usage/Page.tsx
dawsontoth: Good catch — fixed in afd6fec5, though verifying it turned up that `isError` alone isn't sufficient.

I wired up the suggested `isError` check and then forced the request to fail against live stage data to watch it. It still rendered "No usage has been recorded for the current cycle yet". The query state at that moment:

```
{ status: 'pending', fetchStatus: 'paused', hasData: false }
```

When a retry is paused (react-query's default `networkMode: 'online'`), the query parks at `pending`/`paused` — so `isLoading` is false *and* `isError` is false, with no data. The `isError` branch never runs and it falls straight through to the empty state again. Same defect, second route in.

So I keyed it on `!data` instead, which covers the errored, paused and never-enabled cases alike, and reserved the empty state for data that actually arrived reporting zero regions:

```tsx
{!data
  ? <LoadError />
  : data.selfManaged
  ? <Empty>Usage isn't tracked for self-hosted clusters …</Empty>
  : data.regions.length === 0
  ? <Empty>No usage has been recorded for the current cycle yet.</Empty>
  : ( … )}
```

One deliberate consequence of keying on absent data rather than the flag: a failed *background* refetch keeps rendering the cached numbers instead of blanking real usage out behind an error. There's a test pinning that.

`LoadError` uses the `border-destructive/40 bg-destructive/10 text-destructive` treatment from `StorageTab`, with `role="alert"`.

Verified end-to-end on stage — same forced failure now renders the error where it previously claimed an empty cycle:

> Couldn't load usage data — refresh to try again.

and with the failure removed the meters render unchanged (storage still 4 GB / 5 GB, 70%).

4 new tests: the errored case, the paused case (both flags false), that a genuinely empty cycle is still called empty, and the cached-refetch-failure case. The errored-case test was confirmed to fail against the pre-fix source printing the old copy. Cluster suite 110 passing; full suite 1881 passing via the pre-commit hook.

dawsontoth:
Acted on both review comments from @cb1kenobi's pass — pushed in afd6fec5, threads resolved.

- **`null · Reads` on the overview card** — fixed as suggested (`mc.region ?? 'Region'`), matching how the Usage tab already guards that field.
- **Fetch failure shown as "no usage this cycle"** — fixed, but *not* the way the comment suggested, and that's worth flagging: an `isError` check alone doesn't close it. Forcing the request to fail on stage parked the query at `status: 'pending'` / `fetchStatus: 'paused'` (react-query pausing an offline retry), which leaves `isLoading` **and** `isError` both false with no data — so it fell through to the empty state again. Keyed on `!data` instead, which covers the errored, paused and never-enabled cases together, and left the empty state for data that actually arrived reporting zero regions. A failed *background* refetch still shows the cached numbers rather than blanking them behind an error. Details in the thread.

**Verification:** both bug reproductions were confirmed to fail against the pre-fix source, printing exactly the reported strings. Verified in the browser on live stage data — meters and overview card render unchanged, and the forced failure now shows "Couldn't load usage data — refresh to try again." where it previously claimed an empty cycle. 5 new tests (cluster suite **110 passing**); full suite **1881 passing**, `tsc`/`oxlint`/`dprint` clean.

src/integrations/api/cluster/getClusterUsage.ts
```diff
@@ -0,0 +1,115 @@
...
+	writeBytes: UsageValue;
+	realTimeMessages: UsageValue;
+	realTimeBytes: UsageValue;
+	cpuTimeHours: UsageValue;
+	storageBytes: UsageValue;
+}
+
+export type UsageMetricKey = keyof UsageMetrics;
+
+export interface UsageRateLimits {
```
kriszyp, line 49: Can we normalize these limits in the endpoint before presenting them as regional throughput ceilings? central-manager's `rateLimitsFromPlan` currently copies raw `Plan.planLimits` fields. That makes this page wrong in two supported cases: managed `fabric-block-dedicated-unlimited-{2..5}` plans return `-1`, so `rowsFrom` displays `-1` / `-1 B`; and the existing plan UI defines read/read-bandwidth/real-time/TLS regional rates as the plan value multiplied by `Region.purchasedBlockMultiplier`, so a tier with multiplier >1 will show a lower number here. Please have the usage endpoint return effective per-region values with explicit unlimited/unknown semantics, consume that normalized shape here, and add coverage for both cases. Multiplying `activeBlockCount` in Studio is not safe because a mixed cohort after a plan change can contain more live blocks than the current region tier.

src/integrations/api/cluster/getClusterUsage.ts
```diff
@@ -0,0 +1,115 @@
...
+		staleTime: 60_000,
+	});
+}
+
+export function useClusterUsage(clusterId?: string) {
+	return useQuery(getClusterUsageQueryOptions(clusterId));
+}
+
+/** "Standard plan · renews Aug 12" — shared by the card + tab. */
+export function usageSubtitle(data: ClusterUsage): string {
```
kriszyp, line 130: Could we make this subtitle conditional on the regions being uniform? The endpoint sorts regions alphabetically and supports a different current plan per region, while `renewsAt` is the earliest expiry among active regions. As written, a Europe/Standard + US/Enterprise cluster can display `Standard plan · renews Aug 12` as a cluster-wide statement even when the US region renews later. The current copy is accurate for one region; for multiple regions, only show the plan when all names match and label the date `next renewal` (or omit these summary facts), then cover heterogeneous plans/expiries in the card test.

kriszyp (commented):
This is awesome, going to be great to have this!
A couple of inline comments to consider.

cb1kenobi (approved):

src/integrations/api/cluster/getClusterUsage.ts
dawsontoth: Done — normalized in the endpoint, as asked: **HarperFast/central-manager#635**.

`rateLimits` entries are now `{ value, unlimited, known }` — the same three states the metered values already carry — instead of a copy of `Plan.planLimits`:

```js
readsPerMinute:  { value: 15000, unlimited: false, known: true }   // 5000 × us-3's multiplier
tlsHandshakes:   { value: null,  unlimited: true,  known: true }   // plan grants no ceiling
writesPerMinute: null                                             // plan declares none → no row
```

Both cases you named:

- **`-1` on `fabric-block-dedicated-unlimited-{2..5}`** — confirmed against `src/models/plan.json`: those four plans store `-1` for *every* `planLimits` field. Probing the old render path with the sentinel returned exactly the strings you predicted, `"-1"` and `"-1 B"`. Now resolved server-side to `unlimited`.
- **`purchasedBlockMultiplier`** — scaled for the fields the purchase-time panel scales (`ResourcesPerInstance.tsx`): reads, read bandwidth, real-time deliveries/bandwidth, TLS handshakes. Writes go to every replica, so they stay per block. Real regions run multipliers 1–5 (`us-1`…`us-5`, `global-1`…`global-5`), so a `us-3` cluster was being quoted a third of its real read ceiling.

On not multiplying in Studio — agreed, and the endpoint doesn't use `activeBlockCount` either. It takes the multiplier from the `Region` row that supplied the plan, so a mixed cohort can't inflate it. One case that needed a decision: when that row (or its multiplier) can't be resolved, it reports `known: false` → "—" rather than the unscaled number, since an understated throughput ceiling reads as a real, lower limit.

Studio-side, the same reasoning covers the deploy gap: a response that predates the normalization arrives as bare numbers, and with no `known`/`unlimited` flags to trust those render "—" rather than as a regional ceiling. Verified on stage against the currently-deployed endpoint — the tab renders, every ceiling shows "—", nothing stale. With the normalized shape simulated locally against the same live cluster, all three states render correctly (`1,000` reads/min, `Unlimited` TLS, `—` for an unresolved one).

**Coverage:** 6 cases in `tests/unit/clusterUsage.test.js` (scaling + writes unscaled, multiplier 1, the tier that supplied the plan winning across a mixed `us-3`+`us-2` group, the all-`-1` plan, an undeclared ceiling staying null, an unresolvable multiplier) — 5 of them fail against the previous implementation. Plus 5 studio tests for the render states, all failing against the pre-fix source. Per-instance resources drop non-positive values too, so a future sentinel there can't render as "-1 GB".

src/integrations/api/cluster/getClusterUsage.ts
dawsontoth: Done, in `07dbe230`/`f7169afa` — both summary facts are now conditional.

```ts
const uniformPlanName = regions.every((r) => r.planName === regions[0]?.planName) ? regions[0]?.planName : null;
const renewal = regions.length > 1 ? 'next renewal' : 'renews';
```

So your Europe/Standard + US/Enterprise cluster reads `next renewal Aug 12` — no plan claimed cluster-wide — while a uniform multi-region cluster reads `Standard plan · next renewal Aug 12`, and the single-region case is unchanged (`Standard plan · renews Aug 12`). A region whose plan didn't resolve (`planName: null`) also breaks uniformity, so an unresolved region can't have a sibling's plan name spoken for it. The per-region plan and its own renewal date stay on the Usage tab, which already renders them per region.

3 new cases in `ClusterUsageCard.test.tsx`: heterogeneous expiries getting the "next renewal" label, mixed plan names dropping the name (asserting neither "Standard" nor "Enterprise" appears anywhere on the card), and the null-plan region doing the same. All three fail against the pre-fix source. Verified on stage that the single-region card still reads `Fabric Managed Service Block Level 0 plan · renews Dec 4`.

dawsontoth:
Picked this up and addressed both of @kriszyp's comments — pushed in `f7169afa`/`07dbe230`, threads resolved.

**Rate ceilings — normalized in the endpoint, as asked: HarperFast/central-manager#635.** Both cases you named check out. The four `fabric-block-dedicated-unlimited-{2..5}` plans store `-1` for *every* `planLimits` field, and probing the old render path with it returned exactly `"-1"` and `"-1 B"`. And real regions run block multipliers 1–5 (`us-1`…`us-5`, `global-*`), so a `us-3` cluster was shown a third of its actual read ceiling. Entries are now `{ value, unlimited, known }` — the same three states the metered values already carry — with the tier-dependent fields (reads, read bandwidth, real-time, TLS) scaled the way the purchase-time plan panel defines them, and writes left per block. No `activeBlockCount` arithmetic anywhere: the multiplier comes from the `Region` row that supplied the plan, and when it can't be resolved the ceiling reports unknown → **"—"** rather than the understated number.

**Subtitle** now states the plan only when every region names the same one, and labels the date **"next renewal"** once there's more than one region. Your Europe/Standard + US/Enterprise example reads `next renewal Aug 12` with no plan claimed cluster-wide; single-region is unchanged.

**Deploy order:** studio degrades safely against the endpoint that's live today — a pre-normalization response arrives as bare numbers and, with no flags to trust, renders **"—"** instead of a hard number. Verified on stage: the tab renders against the currently-deployed endpoint with every ceiling as "—", and with the normalized shape simulated locally against the same live cluster all three states render (`1,000` reads/min, `Unlimited` TLS handshakes, `—` for an unresolved one). So the two PRs can land in either order; real ceilings appear once #635 ships.

**Verification:** 9 new studio tests and 6 new central-manager ones; 8 of the 9 and 5 of the 6 were confirmed to fail against the pre-fix sources. Cluster suite **119 passing**, full studio suite **1890**; central-manager unit suite **33**. `tsc`/`oxlint`/`dprint` and `eslint` clean.

Also tightened per-instance resources to drop non-positive values, so a future sentinel there can't render as "-1 GB".

kriszyp (approved):
Great!

dawsontoth:
One follow-up from @kriszyp's review on the endpoint PR (HarperFast/central-manager#635): rate ceilings now reuse the metric state names, so `UsageRateLimit` is literally `Omit<UsageValue, 'used'>` instead of a parallel `{ value, unlimited, known }` shape. A rate ceiling *is* a metered ceiling minus `used`, so one formatter takes both and there's no second mapping to keep in sync. Behavior is unchanged; a new test passes a metered value straight through the ceiling formatter so the two can't drift apart without failing (`b912a897`).

_🤖 Generated with [Claude Code](https://claude.com/claude-code)_

Issue #1546, opened 2026-07-23
[RUM] Handled 403s step up ~20× — authenticated requests fire on / and /sign-in while signed out


dawsontoth:
## Summary

Handled `AxiosError: Request failed with status code 403` errors stepped up **~20×** in the last 24h, and are concentrated on the **signed-out** views (`/` and `/sign-in/`). This looks like authenticated background requests (operation polling, `/Login/`, org/cluster fetches) firing while the user has no valid session, each logged to the console → Datadog RUM.

Low user impact (they're *handled*), but it's meaningful RUM noise and points at background requests that shouldn't run while unauthenticated.

## Datadog RUM findings (last 24h vs prior 24h)

- **403 error volume:** ~9 (prior 24h) → **177** (last 24h), ~20× step-up. (Meanwhile 500s on the same operation endpoints dropped 123 → 4, so overall error mix shifted, not just grew.)
- **Distinct sessions with 403s:** 17. The `/Cluster/.../operation/` subset (the largest, 96 resource-events over 48h) is concentrated in **~2 users / 10 sessions**.
- **Error views:** `/` (86), `/sign-in/` (73), `/$organizationId/users/` (16), `/$organizationId/$clusterId/instances/` (8), rest scattered.
- **Failing endpoints (403 resource events, 24h):**
  - `POST /Cluster/{id}/operation/` — 96
  - `GET /Organization/{id}` — 15
  - `POST /Login/` — 10
  - `GET /Cluster/{id}` — 7
  - `POST /HDBInstance/{id}/operation/` — ~30 across instances
  - `GET /HarperVersions/` — 2
- **Handling stack:** mostly the axios interceptor `console.error`; ~16 originate from a TanStack Router `beforeLoad` → `preloadRoute` (route preload firing an authed request during a sign-in redirect).
- **App:** Studio RUM `f590deee-4bac-49b4-a202-3b6963d9721d`; observed on current deploy.

## Likely cause

Background/polling queries and route preloads (cluster & instance `operation` polling, org list, `HarperVersions`, `Login`) are firing on the `/` and `/sign-in/` views — i.e. after the session is gone / before it's established — and the server correctly returns 403. The router `beforeLoad` preload during the `sign-in?redirect=…` flow contributes a slice.

## Related (do not dupe)

- #1527 — 500s on the same instance/cluster `/operation` polling endpoints
- #1386 (closed) — persistent 401 on instance operation polling
- #1387 (closed) — TanStack Router preload race
[...]

Issue #1527, opened 2026-07-16
[RUM] 500s on instance/cluster /operation polling endpoints — ~15× step-up since ~2026-07-14


dawsontoth:
## Summary

Datadog RUM shows a **~15× step-up in first-party HTTP 500 responses on the instance/cluster operation endpoints**, beginning around **2026-07-14** and sustained since. These are `POST` requests Studio makes while polling instance/cluster state.

Surfaced by the daily automated RUM review (RUM app `f590deee-…`, last 24h).

## Datadog findings

**Resource 500s, POST, `provider: first party (harper.fast)`** — daily trend:

| Window (rel. to 2026-07-16) | 500s on `*/operation` |
| --- | --- |
| 72h → 48h ago | **8** |
| 48h → 24h ago | **119** |
| 24h → now | **116** |

- **Endpoints:** `POST /HDBInstance/{id}/operation` and `POST /Cluster/{id}/operation` (IDs redacted — customer resources).
- **Spread:** across ~8 distinct instances/clusters, so not a single bad resource.
- **Reach (last 24h):** 10 sessions, 6 users.
- **Server behavior:** fast failures (~84–240ms), i.e. the backend rejects quickly rather than timing out.
- **Client pattern:** bursty — a handful of sessions each produce a run of 500s, consistent with Studio's polling loop continuing to hammer the endpoint after a 500 (no visible backoff/circuit-break).

A correlated, smaller signal appears in the JS error stream over the same window: handled `AxiosError: Request failed with status code 500` from `ensureUserIsLoaded → loadUser` on the `/` view, trending **7 → 23 → 25** across the same three days. Both are first-party 500s that stepped up together ~2 days ago, which points at a **backend change around 2026-07-14** as the likely trigger.

## Why this is worth tracking

- The overall RUM error volume actually **dropped sharply** in the last 24h (1627 → 467 → 109 error events/day) as the Monaco worker-OOM fixes (#1499/#1407/#1406) landed — so this 500 regression stands out as the one thing trending the wrong way in an otherwise-improving window.
- Distinct from #1386 (that was a **401** on the same polling endpoint, now closed); this is a **500**.

## Suggested next steps

1. Confirm on the backend whether `/HDBInstance/{id}/operation` and `/Cluster/{id}/operation` began returning 500 around 2026-07-14 (deploy correlation).
2. Determine whether the 500s are expected for a resource state (e.g. stopped/errored instance) that Studio should handle gracefully rather than retry against.
[...]

Issue #1386, opened 2026-06-29
[RUM] Persistent AxiosError 401 on instance operation polling endpoint


dawsontoth:
## Summary

Datadog RUM error tracking has flagged a recurring `AxiosError: Request failed with status code 401` that has been active since March 2026 and is still occurring in the latest version (v2.125.4). 6 occurrences were observed in the last 24 hours.

**Datadog Error Tracking:** https://app.datadoghq.com/error-tracking/issue/680ca15e-1db7-11f1-9c16-da7ad0900002

## Error Details

- **Type:** `AxiosError`
- **Message:** `Request failed with status code 401`
- **Location:** `axios/lib/core/settle.js` → `settle()`
- **First seen:** v2.94.4 (March 2026)
- **Last seen:** v2.125.4 (2026-06-29)
- **Platform:** Browser
- **24h occurrences:** 6

## Root Cause (from Datadog AI analysis)

The frontend polls the `/HDBInstance/<instance-id>/operation/` endpoint, which consistently returns HTTP 500 for certain instances in a broken or unsupported state. Other instances succeed normally. The 401 is surfaced to the user/client even though the underlying failure is a server-side 500 on the operation handler.

A companion `AxiosError: Network Error` (from the same instance going fully unreachable after repeated 500s) has already been triaged in Datadog, but this 401 variant remains unreviewed.

## Impact

- Affected view: root path (`/`)
- Ongoing since March 2026, present across 31+ Studio versions
- Users with broken/unsupported instances experience repeated error events

## Suggested Investigation

1. Review error handling in the operation polling logic — the UI may be silently swallowing 500s in a way that produces misleading 401 surface errors.
2. Consider whether instances in a terminal/broken state should be detected and have polling stopped or surfaced as a clear "instance unavailable" state in the UI.
3. Check if there's a missing auth retry / token refresh path that gets triggered inappropriately when the backend 500s.

## Labels

`bug` `rum`

PR #1566, opened 2026-07-27
fix(polling): stop 10s polls once the endpoint answers 403

dawsontoth:
## Summary

Datadog RUM for the last 24h (2026-07-27) shows handled `AxiosError: Request failed with status code 403` at **526 events, up from 133** the prior 24h. **525 of the 526 came from a single session** — not a fleet-wide regression, but it exposes a real defect: **a poll that 403s never stops.**

That session sat on `/$organizationId/$clusterId/instances/` for a cluster it lacks permission on. Three 10s polls ran side by side and every tick 403'd, for ~58 minutes:

| Endpoint | 403 responses | Source |
| --- | --- | --- |
| `GET /Cluster/{clusterId}` | 349 | `Instances.tsx` → `getClusterInfoQueryOptions(clusterId, true)` |
| `POST /HDBInstance/{id}/operation` | 348 | `InstanceStatusCell.tsx` → `getStatusQueryOptions` |
| `POST /HDBInstance/{id}/operation` (2nd row) | 348 | same, second instance row |

**~1,045 requests that could never succeed**, at a measured median interval of 10.2s.

## Root cause

`refetchInterval` fires on a timer regardless of the query's error state. A 403 is *stable* — unlike a 401 (session lost; the auth layer clears auth and redirects, and `unauthorizedResponseHandler` deliberately leaves 403 alone), a 403 means "authenticated but not permitted" and will keep failing until permissions change. Nothing broke the loop.

Each failure also hit the global React Query error handler in `queryClient.ts`, which does `console.error` **and** `toast.error` — so that session generated **525 console errors and 525 toasts**. The console errors are what reach RUM, and `shouldKeepEvent` deliberately does *not* drop 403s (they usually indicate a UI bug worth seeing, per #1386), so an unbounded 403 poll drowns real signal in Error Tracking.

## Fix

Two small helpers in `src/react-query/pollUnlessForbidden.ts`:

- **`pollUnlessForbidden(interval)`** wraps a fixed `refetchInterval` so the timer stops once the query's last error is a 403.
- **`retryUnlessForbidden(max = 3)`** rejects a 403 immediately while keeping `failureCount < 3` for everything else. Without it `state.error` stays empty until retries are exhausted, so the poll timer could not see a 403 until three more doomed requests had gone out — ~30s later on a query that also sets `retryDelay: 10_000`.

Applied to every **always-on** poller (the original two plus the rest, after review):

| Query | Interval | Wrapper |
| --- | --- | --- |
| `getClusterInfoQuery` | 10s | poll (already `retry: false`) |
| `getStatus` | 10s | poll + retry |
| `getListUsers` / `getListRoles` | 10s | poll + retry |
| `getSSHKey` | 10s | poll + retry |
| `getOrganizationRoleInfo` | 10s | poll + retry |
| `getOrganizationQuery` / `getOrganizationRoles` | 10s | poll (already `retry: false`) |
| `getChallengeCertificates` | 5s | poll + retry |

Deliberately narrow:

- **Only 403 stops the timer.** 5xx, network failures, and timeouts stay on the poll so the UI still self-heals when an instance is restarting or briefly unreachable.
- **Stopping is not permanent.** Returning `false` cancels only the *timer*; `refetchOnWindowFocus` (default true) and remount still refetch. A user granted access recovers on their next interaction, no reload. This is now asserted, not just documented.
- **`getInstanceHealth` is left alone** — it swallows errors into a boolean, so `state.error` is always null and the wrapper would be a no-op. Other unwrapped `refetchInterval`s are conditional or self-terminating (`getReadLog`, `listDeployments`, `getDeployment`, `useGetStripeInvoices`).

Also collapses repeated 403 toasts onto a single sonner id, matching the existing timeout collapse in the same function.

## Not a duplicate

- **#1546** — signed-*out* requests firing on `/` and `/sign-in`. Its proposed fix (gate queries on an authenticated session) does **not** cover this: here the user *is* authenticated and simply lacks permission on that cluster. This PR is the missing backoff half.
- **#1386** (closed) — same Datadog fingerprint group, but was resolved by *filtering telemetry* (`shouldKeepEvent` drops 401s), not by changing polling behavior. Its own suggestion #2 — "consider whether instances in a terminal/broken state should have polling stopped" — is what this PR finally implements, for the 403 case.
- **#1527** — 500s on the same endpoints; explicitly left polling, per above.

## Testing

- `pollUnlessForbidden.test.ts` — both predicates across 403 / 401 / 5xx / network / custom budgets.
- Call-site regression tests in `getClusterInfoQuery.test.ts` and `getStatus.test.ts`, so the wiring can't silently regress if someone reverts `refetchInterval` to a bare number.
- **`pollUnlessForbidden.integration.test.ts`** — a real `QueryClient` + `QueryObserver` with fake timers, driving the real `getStatusQueryOptions`. Writing this surfaced a trap worth knowing about: query-core computes `isServer = typeof window === 'undefined'`, and `QueryObserver#updateRefetchInterval` returns **before** `setInterval` when that is true. Under vitest's default **node** environment no interval is ever armed, so any polling assertion there passes vacuously — a "keeps polling on 5xx" test written in node is really just counting retries. The file is jsdom and carries a canary test that fails if intervals stop being armed. Mutation-checked: reverting the fix turns the 403 case from 1 request into 7.
- Toast-collapse tests in `queryClient.test.ts`.
- Full suite: **1,820 passed, 11 skipped**; `tsc -b`, `oxlint`, and `dprint check` all clean.

Not browser-verified: reproducing it needs an account lacking permission on a cluster that still renders the instances page. Verification is mechanism-level, now including real React Query timer behavior.

### DIFF
--- a/src/integrations/api/instance/auth/getListUsers.ts
+++ b/src/integrations/api/instance/auth/getListUsers.ts
@@ -1,12 +1,16 @@
 import { InstanceClientConfig, InstanceClientIdConfig } from '@/config/instanceClientConfig';
 import { LocalUser } from '@/integrations/api/api.patch';
+import { pollUnlessForbidden, retryUnlessForbidden } from '@/react-query/pollUnlessForbidden';
 import { queryOptions } from '@tanstack/react-query';
 
 export function getListUsersQueryOptions({ entityId, instanceClient }: InstanceClientIdConfig) {
 	return queryOptions({
 		queryKey: [entityId, 'list_users'] as const,
 		queryFn: () => getListUsers({ instanceClient }),
-		refetchInterval: 10_000,
+		// `list_users` is a superuser-only operation, so a read-only member sitting on
+		// the Users tab 403s on every tick — stop rather than poll it forever.
+		refetchInterval: pollUnlessForbidden(10_000),
+		retry: retryUnlessForbidden(),
 	});
 }
 
--- a/src/integrations/api/instance/status/getStatus.ts
+++ b/src/integrations/api/instance/status/getStatus.ts
@@ -1,5 +1,5 @@
 import { InstanceClientIdConfig } from '@/config/instanceClientConfig';
-import { pollUnlessForbidden } from '@/react-query/pollUnlessForbidden';
+import { pollUnlessForbidden, retryUnlessForbidden } from '@/react-query/pollUnlessForbidden';
 import { queryOptions } from '@tanstack/react-query';
 
 export interface SystemStatus {
@@ -46,6 +46,10 @@ export function getStatusQueryOptions({ entityId, instanceClient }: InstanceClie
 		// timer instead of re-asking every 10s (one such row emitted 348 doomed
 		// requests in under an hour).
 		refetchInterval: pollUnlessForbidden(10_000),
+		// Without this the default `retry: 3` would swallow the first three 403s into
+		// `failureReason`, so the poll above could not stop until ~30s (and 3 more
+		// doomed requests) later, given the 10s retryDelay.
+		retry: retryUnlessForbidden(),
 		retryDelay: 10_000,
 		throwOnError: false,
 		enabled,
--- a/src/integrations/api/instance/status/getStatus.ts
+++ b/src/integrations/api/instance/status/getStatus.ts
@@ -1,4 +1,5 @@
 import { InstanceClientIdConfig } from '@/config/instanceClientConfig';
+import { pollUnlessForbidden } from '@/react-query/pollUnlessForbidden';
 import { queryOptions } from '@tanstack/react-query';
 
 export interface SystemStatus {
@@ -41,7 +42,10 @@ export function getStatusQueryOptions({ entityId, instanceClient }: InstanceClie
 	return queryOptions({
 		queryKey: ['get_status', entityId] as const,
 		staleTime: 9_000,
-		refetchInterval: 10_000,
+		// An instance the user can't operate answers 403 on every poll — stop the
+		// timer instead of re-asking every 10s (one such row emitted 348 doomed
+		// requests in under an hour).
+		refetchInterval: pollUnlessForbidden(10_000),
 		retryDelay: 10_000,
 		throwOnError: false,
 		enabled,
--- a/src/features/cluster/domains/queries/getChallengeCertificates.ts
+++ b/src/features/cluster/domains/queries/getChallengeCertificates.ts
@@ -1,5 +1,6 @@
 import { getInstanceClient } from '@/config/getInstanceClient';
 import { getSearchByValue } from '@/integrations/api/instance/database/getSearchByValue';
+import { pollUnlessForbidden, retryUnlessForbidden } from '@/react-query/pollUnlessForbidden';
 import { queryOptions, useQuery } from '@tanstack/react-query';
 
 export interface ChallengeCertificate {
@@ -31,7 +32,10 @@ export function getChallengeCertificatesQueryOptions(clusterId?: string) {
 			return data;
 		},
 		enabled: !!clusterId,
-		refetchInterval: 5000,
+		// Same always-on shape as the 10s pollers: gated only on `clusterId`, so a
+		// user without access to the cluster's data 403s every 5s until they navigate.
+		refetchInterval: pollUnlessForbidden(5000),
+		retry: retryUnlessForbidden(),
 	});
 }
 
--- a/src/features/cluster/queries/getClusterInfoQuery.test.ts
+++ b/src/features/cluster/queries/getClusterInfoQuery.test.ts
@@ -1,0 +1,43 @@
+import { AxiosError } from 'axios';
+import { describe, expect, it } from 'vitest';
+import { getClusterInfoQueryOptions } from './getClusterInfoQuery';
+
+/** The subset of `Query` the `refetchInterval` callback reads. */
+type ErrorStateQuery = { state: { error: unknown } };
+
+function axiosErrorWithStatus(status: number): AxiosError {
+	const err = new AxiosError(`Request failed with status code ${status}`);
+	err.response = { status } as AxiosError['response'];
+	return err;
+}
+
+function resolveInterval(refetch: boolean | number | undefined, error: unknown) {
+	const { refetchInterval } = getClusterInfoQueryOptions('clu-test', refetch);
+	expect(typeof refetchInterval).toBe('function');
+	const resolve = refetchInterval as unknown as (q: ErrorStateQuery) => number | false;
+	return resolve({ state: { error } });
+}
+
+describe('getClusterInfoQueryOptions polling', () => {
+	it('defaults to a 10s poll when refetch is true', () => {
+		expect(resolveInterval(true, null)).toBe(10_000);
+	});
+
+	it('honors an explicit interval', () => {
+		expect(resolveInterval(2_000, null)).toBe(2_000);
+	});
+
+	it('does not poll when refetch is omitted', () => {
+		expect(resolveInterval(undefined, null)).toBe(false);
+	});
+
+	it('stops polling after a 403 rather than retrying every 10s (RUM 2026-07-27)', () => {
+		// One session emitted 349 doomed GET /Cluster/{id} requests in ~58 minutes.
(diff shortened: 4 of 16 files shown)


src/integrations/api/instance/status/getStatus.ts
```diff
@@ -41,7 +42,10 @@ export function getStatusQueryOptions({ entityId, instanceClient }: InstanceClie
 	return queryOptions({
 		queryKey: ['get_status', entityId] as const,
 		staleTime: 9_000,
-		refetchInterval: 10_000,
+		// An instance the user can't operate answers 403 on every poll — stop the
+		// timer instead of re-asking every 10s (one such row emitted 348 doomed
+		// requests in under an hour).
+		refetchInterval: pollUnlessForbidden(10_000),
 		retryDelay: 10_000,
```
kriszyp, line 53: `refetchInterval` cannot see the first 403 here: this query still uses TanStack Query's client default of three retries, and `state.error` is only populated after those retries are exhausted (`failureReason` holds the earlier failures). With `retryDelay: 10_000`, a forbidden status request therefore makes three more doomed requests over about 30 seconds before `pollUnlessForbidden` can return `false`. Could we add a status-aware `retry` predicate that immediately rejects `isForbiddenError(error)` while preserving the normal retry count for transient failures, and cover that behavior through these query options?

DavidCockerill (commented):
The mechanism here is sound and the perf win is real — killing ~1,000 doomed requests/hour/session is well worth it, and using the query's own error state to gate `refetchInterval` is the right shape. Not blocking; two things I'd flag before this lands.

**1. Only 2 of ~8 always-on pollers get the fix — and the ones left out are the most permission-sensitive in the app.**

Structurally identical unbounded 10s polls at head, none wrapped:

| Query | Endpoint |
| --- | --- |
| `getListUsers.ts:9` | `list_users` |
| `getListRoles.ts:9` | `list_roles` |
| `getOrganizationQuery.ts:30` | `GET /Organization/{id}` |
| `getOrganizationRoles.ts:15` | `GET /Role` |
| `getOrganizationRoleInfo.ts:20` | `GET /Role/{id}` |
| `getSSHKey.ts:27` | `get_ssh_key` |

`list_users` / `list_roles` are *the* canonical "non-superuser gets a 403 from the Harper operations API" pair, and `getOrganizationQuery` has `retry: false` + a bare `[orgId]` key on every org-scoped route.

In plain terms: the exact bug this PR fixes is still reachable — just from the Users page or the org lookup instead of the Instances page. A read-only org member who opens the instance Users tab and walks away to lunch reproduces the same thousand-request hour on a different URL, and Error Tracking gets buried the same way. The helper is already written; it just needs pointing at the other six — a mechanical one-liner each, either here or as a fast follow.

(Deliberately not on that list: `getInstanceHealth.ts:19` swallows errors into a boolean, so `state.error` is always null and the wrapper would be a no-op there. Separate pre-existing problem.)

**2. No test exercises a real `QueryClient`, so the one assumption the whole fix rests on is unverified.**

Every test invokes the option function directly with a hand-built `{ state: { error } }`. That proves the *predicate* is right; it doesn't prove React Query ever calls it. The fix works because `QueryObserver#onQueryUpdate` → `#updateTimers` → `#computeRefetchInterval` re-invokes the callback on each state change — I'm fairly confident that's correct for v5.90, but confidence isn't a test. If a future TanStack major changes when that's consulted, all six assertions still pass and the storm silently returns.

Equally untested is the resume path the docblock leans on: nothing asserts that a subsequent successful fetch clears `state.error` and re-arms the timer. That's the load-bearing reason "stopped" isn't "dead until reload" — right now it's a comment, not a guarantee.

One `QueryClient` + `vi.useFakeTimers()` test would pin both halves in ~25 lines: mount an observer against a 403ing `queryFn`, advance 60s, assert one fetch; flip the fn to succeed, trigger a focus refetch, advance 30s, assert polling resumed.

**Two smaller notes:**
- Default `retry: 3` + `retryDelay: 10_000` on `getStatusQueryOptions` means the stop engages ~30s and 4 requests late on every trigger. `retry: (n, e) => !isForbiddenError(e) && n < 3` reuses the predicate you already export.
- The stable sonner id resets the toast's dismiss timer on each fire, so a *repeating* 403 source now pins one toast on screen indefinitely rather than showing a wall that each expire. Probably still the better trade (and consistent with the existing `request-timeout` behaviour) — just noting it's a trade, and that two genuinely different permission failures now collapse onto one toast showing whichever fired last.

kriszyp (approved):
LGTM

src/integrations/api/instance/status/getStatus.ts
dawsontoth: Fixed in 9cf77360 — added `retryUnlessForbidden()` (rejects `isForbiddenError` immediately, keeps `failureCount < 3` for transient failures) and applied it here alongside the poll wrapper; a real-QueryClient test now asserts a 403 makes exactly 1 request instead of 7.

dawsontoth:
Both points addressed in 9cf77360.

**1. The other pollers.** All six are wrapped, plus `getChallengeCertificates.ts:34` (5s, not on your list but the same always-on shape — gated only on `clusterId`). `getOrganizationQuery` and `getOrganizationRoles` already set `retry: false`, so they take only the poll wrapper; the rest also get `retryUnlessForbidden()`. Agreed on `getInstanceHealth` — it swallows errors into a boolean so `state.error` is always null and the wrapper would be a no-op; left alone. Remaining unwrapped `refetchInterval`s are all conditional or self-terminating (`getReadLog` user-toggled, `listDeployments` `pollWhileOpen`, `getDeployment` already stops at terminal state, `useGetStripeInvoices` conditional).

**2. The retry gap** (same as @kriszyp's inline note). `retryUnlessForbidden()` rejects 403 immediately and keeps `failureCount < 3` otherwise, so the stop engages on the first 403 rather than ~30s and 4 requests late.

**3. The real-QueryClient test — this one earned its keep.** Writing it surfaced something the hand-built `{ state: { error } }` tests hid:

query-core computes `isServer = typeof window === "undefined"`, and `QueryObserver#updateRefetchInterval` returns **before** `setInterval` when that is true. Under vitest's default node environment **no interval is ever armed** — so any polling assertion written there passes vacuously. My first draft of "keeps polling through a transient 5xx" was green while only counting retries. The new file is jsdom and carries a canary test that fails if intervals ever stop being armed.

It also caught that the resume path genuinely works — it failed under node, which is what sent me digging. It's now asserted rather than described. Mutation-checked end to end: reverting the fix turns the 403 case from 1 request into 7.

**On the toast trade:** agreed it's a trade, and taking it for consistency with `request-timeout`. Worth noting the pinned-toast concern is now much narrower — with polling stopped, the repeating-403 source that would pin a toast indefinitely is mostly gone; what's left is the #1546 signed-out case.

src/integrations/api/instance/auth/getListUsers.ts
```diff
@@ -1,12 +1,16 @@
...
 import { queryOptions } from '@tanstack/react-query';
 
 export function getListUsersQueryOptions({ entityId, instanceClient }: InstanceClientIdConfig) {
 	return queryOptions({
 		queryKey: [entityId, 'list_users'] as const,
 		queryFn: () => getListUsers({ instanceClient }),
-		refetchInterval: 10_000,
+		// `list_users` is a superuser-only operation, so a read-only member sitting on
+		// the Users tab 403s on every tick — stop rather than poll it forever.
+		refetchInterval: pollUnlessForbidden(10_000),
```
cb1kenobi, line 12: **Low: no call-site regression test for the newly-wired pollers**

This commit wires `pollUnlessForbidden`/`retryUnlessForbidden` into 7 additional callers (`getListUsers`, `getListRoles`, `getSSHKey`, `getOrganizationRoleInfo`, `getOrganizationRoles`, `getOrganizationQuery`, `getChallengeCertificates`), but only `getStatus` and `getClusterInfoQuery` got a dedicated call-site test. The PR body's own rationale for those two tests — "so the wiring can't silently regress if someone reverts `refetchInterval` to a bare number" — applies equally to these 7. The wrapper functions themselves are well covered by unit + integration tests, so this is low risk, but a one-line assertion per file (`expect(queryOptions.refetchInterval).not.toBe(10_000)` or similar) would close the gap cheaply.

—
_Generated by Barber AI_

PR #1568, opened 2026-07-28
fix(editor): stop sending non-package strings to jsDelivr from type acquisition

dawsontoth:
## What the daily RUM review found

Datadog RUM, last 24h (app `f590deee-…`, release **v2.154.3**): **18 requests to `data.jsdelivr.com/v1/package/resolve/npm/<specifier>@latest` returned 404** in a single session. One of the specifiers was the URL-encoded string `Known%20Fraudster%20Risk` — a **multi-word phrase with spaces**, which cannot name an npm package and can only have come from the user's own application code.

These were completely silent: `@typescript/ata` collapses every non-OK response into a bare `new Error("OK")` (no status, no retry), and Studio passes it no `logger`. RUM was the only place they surfaced.

Two problems, both worth fixing:

1. **Wasted, uncached cross-origin requests** — one per bogus specifier, per fresh tab.
2. **User-authored strings leave the origin.** Table and type names out of a customer's app code were sent to a third-party CDN as URL paths. That's the part I'd treat as time-sensitive; it's a small egress, but it's customer content going somewhere it was never meant to go.

## Root cause

`harper-language/typeAcquisition.ts` scans project sources for import specifiers with a regex shim — we deliberately don't ship `typescript` to the browser just for `preProcessFile`. The clause matcher between `import`/`export` and `from` was:

```
(?:import|export)\b[^'";]*?\bfrom\s*['"]([^'"]+)['"]
```

`[^'";]*?` excludes quotes and semicolons but **not newlines**, so the lazy match could run past the end of a statement and latch onto any lowercase `from "…"` further down the file. Confirmed shapes:

| source | extracted |
| --- | --- |
| `export const { Dog } = tables` ⏎ `// everything below reads from "Known Fraudster Risk"` | **`Known Fraudster Risk`** |
| `export interface Risk {` ⏎ `/** derived from "Known Fraudster Risk" table */` | **`Known Fraudster Risk`** |
| ``export const q = `select * from "Known Fraudster Risk"` `` | **`Known Fraudster Risk`** |

Harper apps quote SQL identifiers with `"` and semicolon-free Prettier style is common, so this is easy to hit in ordinary code.

`isAcquirablePackage` was the last gate before the network, and it only filtered relative paths, `@/` aliases, assets, Harper globals, and node builtins — **it never checked the specifier could name an npm package at all.**

Worth noting: the scanner also runs over every **downloaded `.d.ts`** (ata recurses via `resolveDeps`), so vendor declaration prose is a second feeder.

## The fix

Both changes are confined to the scanner:

- **Restrict the import clause** to what can actually appear in one — identifiers, braces, commas, `*`, `as`, whitespace, and comments. Whitespace still includes newlines so **multi-line named imports keep working**; `=`, `:`, `(`, `)`, backticks, and a bare `/` are excluded, and that's what stops a match from crossing out of a statement into neighbouring prose. Each comment is matched as one **indivisible unit** — the `//` form pinned to its line end by a lookahead, the block form unable to cross its terminator — so consuming comments can't hand the prose back (see the review round below).
- **Validate against npm's name grammar** before a specifier can become a request: optional `@scope/`, name, optional deep-import subpath, 214-char cap, no whitespace. Case-insensitive so legacy names (`JSONStream`, `Base64`) still resolve.

## Tests

New `typeAcquisition.test.ts`, 62 cases — there were **no tests for this extraction** before:

- Real import forms that must keep resolving: default, named, namespace, star/named re-export, type-only, aliased, bare side-effect, dynamic `import()`, `require()`, scoped, deep, double-quoted, and multi-line named imports.
- Categories that must never be acquired: relative, `@/` alias, `harper`/`harperdb`, `node:`/bare builtins, css/json assets.
- Comments in every legal clause position: `//` and `/* */` inside a named block, `/** */` and `//` before `from`, a comment before a `type` modifier, and CRLF line endings.
- The prose-mining regressions above, plus invalid-specifier and length cases.

**11 of the original 41 fail on the pre-fix scanner** (verified by reverting the source and re-running); the 21 added in review rounds 1-3 are broken out below.

Full suite green: `259 files / 1885 passed, 11 skipped`. `tsc -b`, `oxlint`, `dprint check` all clean.

## Review round 1 — comments in the import clause (66747d99)

Caught by @cb1kenobi's review: excluding a bare `/` from the clause also dropped **any import with a comment between the braces**. `import { useState, // hooks ⏎ } from 'react'` matched nothing, so that package silently stopped getting types — a real regression, and untested.

Letting the clause consume comments fixes it, but only if a comment can't be **partially** consumed. A plain `\/\/[^\n]*` backtracks, so the matcher stops mid-comment and takes the `from "…"` out of the rest of it — which reopens this PR's leak:

| clause matcher | `export { Dog }` ⏎ `// each row is read from "customers"` | annotated import |
| --- | --- | --- |
| `stage` | **`customers`** | ✅ |
| this PR, as first reviewed | ✅ none | ❌ missed |
| bare `\/\/[^\n]*` | **`customers`** | ✅ |
| shipped | ✅ none | ✅ |

The regression cases use a single lowercase table name rather than `Known Fraudster Risk` on purpose: `customers` satisfies npm's name grammar, so the grammar gate waves it through and the **clause matcher is the only thing left** between user content and the CDN. A multi-word phrase is caught either way, which is what masked this in the first draft of the tests.

12 new cases: 6 annotated-import forms (**all 6 fail on the clause as first reviewed**) and 6 prose shapes (**4 fail on a backtrackable comment alternative, all 6 on `stage`**), including a `.d.ts` jsdoc case — the scanner also runs over every downloaded declaration file. Alternatives are disjoint on their first character, so the repetition stays linear: 20k-token clause garbage and 5k-line comment spam both scan in under a millisecond.

## Review round 2 — an unterminated comment could freeze the editor (3c1a42c6)

Also caught by @cb1kenobi, and the more serious of the two. The comment alternatives search for a terminator, and an **unterminated `/*` has none** — so every failed attempt ran to end-of-input, and each later `import`/`export` occurrence paid it again. Quadratic, on the main thread:

| input (32k occurrences) | round 1 | **shipped** |
| --- | --- | --- |
| 719 KB `import /* unterminated` | 21.9s | **238ms** |
| 719 KB, no newlines (minified) | 22.2s | **218ms** |
| 281 KB, `//` at a truncated EOF | 8.1s | **305ms** |
| 281 KB run of clause-legal chars | 7.5s | **254ms** |
| 813 KB of real imports | 2ms | **2ms** |

No hostile input needed: a mid-edit save with an open block comment, or a truncated CDN download. `acquireApplicationTypes` scans every project script joined together, and ata re-scans every downloaded `.d.ts`.

The fix bounds every clause alternative, comment bodies included, at `MAX_IMPORT_CLAUSE = 4096`. Two things worth flagging:

- **It also fixes a pre-existing version of this.** That last-but-one row — a long run of clause-legal characters, no comments at all — took **~2.5s on `stage` and on round 1 alike**. The quadratic wasn't introduced by comment matching; comments widened the trigger and worsened the constant. So the doc comment claiming "the repetition stays linear" was wrong before this round *and* would have been wrong without comments.
- **I didn't take the `code.length` ceiling option.** It doesn't fix the shape — 281 KB is under any sane ceiling and still took seconds — and it would drop acquisition entirely for large but perfectly ordinary projects.

4096 clears real code by a wide margin, measured rather than guessed: the longest clause in this repo is **351 chars** (a `lucide-react` icon import), and across 13,288 clauses in ~4k `node_modules` declaration files the longest is a **1.6 KB** barrel re-export. Past the bound a clause isn't acquired — the same degraded-but-safe outcome as any unacquired package.

7 more cases (60 total): 4 wall-clock guards that **all fail with the bound removed** (7.0-18.5s against a 3s budget, ~10x headroom over the ~300ms actual), plus multi-line jsdoc inside a clause and a 2.9 KB barrel re-export, which the bound must not break.

## Review round 3 — the bound crashed on legal comments, so the clause is no longer a regex (8a3c3c50)

@cb1kenobi again, and this one ends the pattern. Bounding with `{0,n}` traded the quadratic scan for a **crash**: V8 keeps a backtracking frame per repetition, so a long run of large — but perfectly legal, properly terminated — comments inside one clause overflowed the regex stack.

```js
const bigComment = '/*' + 'x'.repeat(4090) + '*/';
findAcquirableSpecifiers('import ' + bigComment.repeat(1400) + ' nofrom\n');
// round 2: RangeError: Maximum call stack size exceeded
// shipped: [] in ~4ms
```

`acquireApplicationTypes` catches it, but it scans every project source joined into one string, so one bad clause aborted acquisition for all of them.

**The bound was also on the wrong thing.** It capped *repetitions*, not characters — so 4096 reps of ~4 KB comments admitted ~16 MB, which is exactly how a 5.7 MB input got through a bound documented as 4 KB. My round-2 doc comment was wrong about that.

Taking the second option suggested in review, the clause is now a **character walk** (`readClauseSpecifier`) rather than a quantified sub-pattern. No backtracking frames to accumulate, every terminator search bounded, and `MAX_IMPORT_CLAUSE` now genuinely bounds characters. Three rounds of regex fixes each introduced the next failure mode; a walk removes the class rather than the instance.

The clause **grammar is unchanged** — identifiers, `{}`, `,`, `*`, whitespace, comments, and anything else ends the clause — because that narrowness is the property keeping user-authored strings off the CDN. Comments are still skipped whole, never partially.

Two things worth a reviewer's eye:

- **Equivalence is tested, not asserted.** All **60 prior cases pass against both** the walk and the round-2 regex, and the 2 new cases fail with `RangeError` on round 2. The only behaviour difference I found across every documented form: a specifier past npm's 214-char limit is no longer read at all rather than read and then rejected — same result through `findAcquirableSpecifiers`, but now a bounded read.
- It also folds the duplicated scan loop in `preProcessFile` into the same `scanSpecifiers`, so both callers agree by construction instead of by copy-paste.

Perf is unchanged-to-better on every earlier shape (166-320ms across the pathological set, ~7ms on 813 KB of real imports).

## Deliberately not in scope

- **Nothing bounds the number of jsDelivr requests per pass.** The existing budget (#1499/#1504) caps *chars admitted to the worker after download*, not requests issued.
- A failed resolution is negative-cached only for the tab's lifetime (ata's `moduleMap`), and there's no HTTP-level negative cache.
- ata swallowing status codes means we still have no telemetry on acquisition failures.

Happy to follow up on any of those separately.

## ⚠️ One housekeeping note

All four commits are **unsigned** — the first came from the scheduled daily RUM review, and GPG signing needs an interactive pinentry that isn't available in a headless run (`gpg: signing failed: Inappropriate ioctl for device`); the review-round commits hit the same wall. Please amend/re-sign before merging if the branch protection requires it.

_Opened by the automated daily Datadog RUM review. No customer identifiers included; the one specifier quoted is a schema/type name, retained because it is the diagnostic fingerprint — say the word and I'll redact it._

### DIFF
--- a/src/features/instance/applications/components/TextEditorView/harper-language/typeAcquisition.ts
+++ b/src/features/instance/applications/components/TextEditorView/harper-language/typeAcquisition.ts
@@ -110,6 +110,8 @@ const MAX_IMPORT_CLAUSE = 4096;
 
 const SLASH = 0x2f, STAR = 0x2a, NEWLINE = 0x0a, SINGLE_QUOTE = 0x27, DOUBLE_QUOTE = 0x22;
 const OPEN_PAREN = 0x28, CLOSE_PAREN = 0x29, OPEN_BRACE = 0x7b, CLOSE_BRACE = 0x7d, COMMA = 0x2c;
+const BACKTICK = 0x60, BACKSLASH = 0x5c, DOLLAR = 0x24, DOT = 0x2e;
+const OPEN_BRACKET = 0x5b, CLOSE_BRACKET = 0x5d;
 
 /** `\w` plus `$` — what a clause keyword or imported binding is spelled with. */
 function isIdentifierChar(charCode: number): boolean {
@@ -167,17 +169,19 @@ interface ScannedSpecifier {
 }
 
 /**
- * A quoted module specifier at `at`, after optional whitespace. Bounded by npm's own
- * name limit: past that the specifier is rejected anyway, so reading further would
- * only turn a missing closing quote into another unbounded scan.
+ * A quoted module specifier at `at`, after optional whitespace. `limit` is the caller's
+ * bound and applies to the whitespace run as well as the quotes: without it, a statement
+ * whose `from` sits inside the clause bound could still reach a quote arbitrarily far
+ * past it, and every keyword candidate would rescan that same run. The specifier body is
+ * additionally capped at npm's own name limit, past which it is rejected anyway.
  */
-function readQuotedSpecifier(code: string, at: number): { specifier: string; end: number } | undefined {
-	const opened = skipWhitespace(code, at, code.length);
+function readQuotedSpecifier(code: string, at: number, bound: number): { specifier: string; end: number } | undefined {
+	const opened = skipWhitespace(code, at, bound);
 	const quote = code.charCodeAt(opened);
-	if (quote !== SINGLE_QUOTE && quote !== DOUBLE_QUOTE) {
+	if (opened >= bound || (quote !== SINGLE_QUOTE && quote !== DOUBLE_QUOTE)) {
 		return undefined;
 	}
-	const limit = Math.min(code.length, opened + MAX_SPECIFIER_LENGTH + 2);
+	const limit = Math.min(bound, opened + MAX_SPECIFIER_LENGTH + 2);
 	for (let next = opened + 1; next < limit; next++) {
 		const charCode = code.charCodeAt(next);
 		if (charCode === quote) {
@@ -244,7 +248,7 @@ function readClauseSpecifier(code: string, clauseStart: number): { specifier: st
 			}
 			// A `from` that no specifier follows is just a binding name (`import from from 'x'`).
 			if (wordEnd - at === 4 && code.startsWith('from', at)) {
-				const found = readQuotedSpecifier(code, wordEnd);
+				const found = readQuotedSpecifier(code, wordEnd, limit);
 				if (found) {
 					return found;
 				}
@@ -264,38 +268,238 @@ function readClauseSpecifier(code: string, clauseStart: number): { specifier: st
 	return undefined;
 }
 
-/** Keywords that can introduce a specifier — the only places the scanner starts work. */
-const STATEMENT_KEYWORD = /\b(?:import|export|require)\b/g;
 const REFERENCE_PATH = /\/\/\/\s*<reference\s+path\s*=\s*['"]([^'"]+)['"]/g;
+
+/**
+ * The specifier belonging to a statement keyword, whichever form follows it: a call
+ * `import('…')`/`require('…')`, a bare side-effect `import '…'`, or a clause ending in
+ * `from '…'`. Every scan is bounded by `MAX_IMPORT_CLAUSE` from the keyword, so no form
+ * can reach across the file and no candidate can rescan a run another already rejected.
+ */
+function readSpecifierAfterKeyword(
+	code: string,
+	keyword: string,
+	afterKeyword: number,
+): { specifier: string; end: number } | undefined {
+	const bound = Math.min(code.length, afterKeyword + MAX_IMPORT_CLAUSE);
+	if (keyword !== 'export') {
+		const parenAt = skipWhitespace(code, afterKeyword, bound);
+		if (code.charCodeAt(parenAt) === OPEN_PAREN) {
+			const call = readQuotedSpecifier(code, parenAt + 1, bound);
+			if (!call) {
+				return undefined;
+			}
+			const closeAt = skipWhitespace(code, call.end, bound);
+			return code.charCodeAt(closeAt) === CLOSE_PAREN ? { specifier: call.specifier, end: closeAt + 1 } : undefined;
+		}
+	}
+	if (keyword === 'require') {
+		return undefined;
+	}
+	const bare = keyword === 'import' ? readQuotedSpecifier(code, afterKeyword, bound) : undefined;
+	return bare ?? readClauseSpecifier(code, afterKeyword);
+}
+
+/** Past the closing quote of a string literal, or `limit` if it never closes. */
+function skipStringLiteral(code: string, at: number, limit: number): number {
+	const quote = code.charCodeAt(at);
+	for (let next = at + 1; next < limit; next++) {
+		const charCode = code.charCodeAt(next);
+		if (charCode === BACKSLASH) {
+			next++;
+		} else if (charCode === quote) {
+			return next + 1;
+		} else if (charCode === NEWLINE) {
+			// An unescaped newline means this quote never opened a string (mid-edit, or an
+			// apostrophe in prose). Give up the single character so scanning resynchronises.
+			return at + 1;
+		}
+	}
+	return limit;
+}
+
+/**
+ * Past the closing `/` of a regex literal, or `at + 1` if this `/` turns out not to
+ * start one — a regex cannot span a line, so hitting a newline first is proof.
+ */
+function skipRegexLiteral(code: string, at: number, limit: number): number {
+	let inCharacterClass = false;
+	for (let next = at + 1; next < limit; next++) {
+		const charCode = code.charCodeAt(next);
+		if (charCode === BACKSLASH) {
+			next++;
+		} else if (charCode === NEWLINE) {
+			return at + 1;
+		} else if (charCode === OPEN_BRACKET) {
+			inCharacterClass = true;
+		} else if (charCode === CLOSE_BRACKET) {
+			inCharacterClass = false;
+		} else if (charCode === SLASH && !inCharacterClass) {
+			return next + 1;
+		}
+	}
+	return at + 1;
+}
+
+/**
+ * Whether a `/` here opens a regex literal rather than being division. Decided from the
+ * previous significant token, the standard heuristic: division follows a value —
+ * an identifier, a literal, `)`, or `]` — and a regex follows anything else.
+ * `previousWord` covers the keywords that produce no value and so may be followed by a
+ * regex (`return /x/`), which a bare "previous char is a letter" test gets wrong.
+ *
+ * Ambiguity is resolved toward *regex*, `}` included: mistaking division for a regex
+ * skips to the next `/` and can only lose an import, while mistaking a regex for
+ * division scans its body as code and could take a specifier out of it.
+ */
+const REGEX_MAY_FOLLOW = new Set([
+	'return',
+	'typeof',
+	'instanceof',
+	'case',
+	'in',
+	'of',
+	'delete',
+	'void',
+	'new',
+	'do',
+	'else',
+	'yield',
+	'await',
+]);
+function canStartRegex(previousChar: number, previousWord: string): boolean {
+	if (isIdentifierChar(previousChar)) {
+		return REGEX_MAY_FOLLOW.has(previousWord);
+	}
+	return previousChar !== CLOSE_PAREN && previousChar !== CLOSE_BRACKET
+		&& previousChar !== SINGLE_QUOTE && previousChar !== DOUBLE_QUOTE && previousChar !== BACKTICK;
+}
 
 /**
  * Every module specifier in `code`, in source order: `import`/`export … from '…'`,
  * dynamic `import('…')`, `require('…')`, and a bare side-effect `import '…'`.
+ *
+ * The walk tracks lexical context so a keyword is only honoured where real code can
+ * appear. That is the root fix for what the RUM review found rather than a mitigation of
+ * it: searching the raw text for keywords meant ordinary English in a comment —
+ * `// we import rows from "customers"` — parsed as an import and sent a user's table name
+ * to jsDelivr. `customers` is a perfectly legal npm name, so no amount of specifier
+ * validation downstream can catch it; the only fix is not to read comments as code. The
+ * same applies to SQL and GraphQL in template literals, quoted strings, and regex bodies.
+ *
+ * `@typescript/ata` calls this synchronously on files the user is actively editing, so
+ * this is deliberately not a parser and never throws: a real module lexer rejects
+ * half-typed code, which for an editor is the normal state, and returning nothing there
+ * would silently stop type acquisition mid-keystroke. Where the walk cannot be certain it
+ * prefers to treat text as *not* code, which loses an acquisition rather than leaking a
+ * name. Template substitutions are still scanned, since `import()` inside one is real
+ * code.
  */
 function scanSpecifiers(code: string): ScannedSpecifier[] {
 	const found: ScannedSpecifier[] = [];
-	for (const match of code.matchAll(STATEMENT_KEYWORD)) {
-		const keyword = match[0];
-		const pos = match.index;
-		const afterKeyword = pos + keyword.length;
-		if (keyword !== 'export' && code.charCodeAt(skipWhitespace(code, afterKeyword, code.length)) === OPEN_PAREN) {
-			const call = readQuotedSpecifier(code, skipWhitespace(code, afterKeyword, code.length) + 1);
-			if (call) {
-				const closeAt = skipWhitespace(code, call.end, code.length);
-				if (code.charCodeAt(closeAt) === CLOSE_PAREN) {
-					found.push({ specifier: call.specifier, pos, end: closeAt + 1 });
-				}
+	// Each entry is a `${` substitution we are inside; its value is the brace depth within
+	// it, so the matching `}` returns us to the enclosing template. An explicit stack
+	// rather than recursion, so deeply nested templates cost memory, never the call stack.
+	const substitutions: number[] = [];
+	let inTemplate = false;
+	let previousChar = 0;
+	let previousWord = '';
+	let at = 0;
+	while (at < code.length) {
+		const charCode = code.charCodeAt(at);
+
+		if (inTemplate) {
+			if (charCode === BACKSLASH) {
+				at += 2;
+			} else if (charCode === BACKTICK) {
+				inTemplate = false;
+				previousChar = BACKTICK;
+				at++;
+			} else if (charCode === DOLLAR && code.charCodeAt(at + 1) === OPEN_BRACE) {
+				substitutions.push(0);
+				inTemplate = false;
+				previousChar = OPEN_BRACE;
+				at += 2;
+			} else {
+				at++;
 			}
 			continue;
 		}
-		if (keyword === 'require') {
+
+		if (charCode === SLASH) {
+			const following = code.charCodeAt(at + 1);
+			if (following === SLASH) {
+				const lineEnd = code.indexOf('\n', at + 2);
+				at = lineEnd < 0 ? code.length : lineEnd + 1;
+				continue;
+			}
+			if (following === STAR) {
+				const commentEnd = code.indexOf('*/', at + 2);
+				at = commentEnd < 0 ? code.length : commentEnd + 2;
+				continue;
+			}
+			at = canStartRegex(previousChar, previousWord) ? skipRegexLiteral(code, at, code.length) : at + 1;
+			previousChar = SLASH;
+			previousWord = '';
 			continue;
 		}
-		const bare = keyword === 'import' ? readQuotedSpecifier(code, afterKeyword) : undefined;
-		const scanned = bare ?? readClauseSpecifier(code, afterKeyword);
-		if (scanned) {
-			found.push({ specifier: scanned.specifier, pos, end: scanned.end });
+
+		if (charCode === SINGLE_QUOTE || charCode === DOUBLE_QUOTE) {
+			at = skipStringLiteral(code, at, code.length);
+			previousChar = charCode;
+			previousWord = '';
+			continue;
 		}
+
+		if (charCode === BACKTICK) {
+			inTemplate = true;
+			at++;
+			continue;
+		}
+
+		if (substitutions.length > 0) {
+			if (charCode === OPEN_BRACE) {
+				substitutions[substitutions.length - 1]++;
+			} else if (charCode === CLOSE_BRACE) {
+				if (substitutions[substitutions.length - 1] === 0) {
+					substitutions.pop();
+					inTemplate = true;
+					previousChar = CLOSE_BRACE;
+					at++;
+					continue;
+				}
+				substitutions[substitutions.length - 1]--;
+			}
+		}
+
+		if (isIdentifierChar(charCode)) {
+			let wordEnd = at + 1;
+			while (wordEnd < code.length && isIdentifierChar(code.charCodeAt(wordEnd))) {
+				wordEnd++;
+			}
+			const word = code.slice(at, wordEnd);
+			// `obj.import` and `{ import: x }` are property names, not statements.
+			const isStatementKeyword = previousChar !== DOT
+				&& (word === 'import' || word === 'export' || word === 'require');
+			const scanned = isStatementKeyword ? readSpecifierAfterKeyword(code, word, wordEnd) : undefined;
+			if (scanned) {
+				found.push({ specifier: scanned.specifier, pos: at, end: scanned.end });
+				previousChar = code.charCodeAt(scanned.end - 1);
+				previousWord = '';
+				at = scanned.end;
+				continue;
+			}
+			previousChar = code.charCodeAt(wordEnd - 1);
+			previousWord = word;
+			at = wordEnd;
+			continue;
+		}
+
+		if (!isWhitespace(charCode)) {
+			previousChar = charCode;
+			previousWord = '';
+		}
+		at++;
 	}
 	return found;
 }
--- a/src/features/instance/applications/components/TextEditorView/harper-language/typeAcquisition.ts
+++ b/src/features/instance/applications/components/TextEditorView/harper-language/typeAcquisition.ts
@@ -87,6 +87,29 @@ function isAcquirablePackage(specifier: string): boolean {
 }
 
 /**
+ * How far one import clause — including any single comment inside it — may extend
+ * before the scanner abandons the statement. Real clauses are nowhere close: the
+ * longest in this repo is 351 chars, and across ~13k declaration files in
+ * `node_modules` the longest is a 1.6 KB barrel re-export, so 4 KB leaves ample
+ * headroom. A clause past the bound just isn't acquired, the same degraded-but-safe
+ * outcome as any unacquired package.
+ *
+ * The bound is what keeps the scan linear, and it is load-bearing. The clause match
+ * is lazy, so without it every failed attempt runs to end-of-input and each
+ * `import`/`export` occurrence pays that again: an unterminated `/*` — a mid-edit
+ * save, a truncated CDN download — took ~22s on 719 KB, and a long run of
+ * clause-legal characters ~2.5s on 281 KB even before comments were admitted.
+ * Bounded, those same inputs scan in ~220-300ms, and real files in ~1ms. This file
+ * promises failures are swallowed so they "can never break editing"; a multi-second
+ * freeze of the main thread would break that promise even though nothing throws.
+ *
+ * A ceiling on `code.length` instead would not have fixed this — 281 KB is under any
+ * sane ceiling and still took seconds — and it would drop acquisition entirely for
+ * large but perfectly ordinary projects.
+ */
+const MAX_IMPORT_CLAUSE = 4096;
+
+/**
  * The clause between `import`/`export` and `from` — identifiers, braces, commas,
  * `*`, `as`, whitespace (newlines included, so multi-line named imports still
  * match), and comments, which are legal and common between the braces. Everything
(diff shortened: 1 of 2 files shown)


src/features/instance/applications/components/TextEditorView/harper-language/typeAcquisition.ts
```diff
@@ -63,13 +74,48 @@ function isAcquirablePackage(specifier: string): boolean {
...
+ * `*`, `as`, and whitespace (newlines included, so multi-line named imports still
+ * match). Deliberately excludes `=`, `/`, `:`, `(`, `)`, and backticks: none can
+ * appear in a real import clause, and admitting them let the lazy match run past
+ * the end of a statement to reach a `from "…"` sitting in a following comment,
+ * SQL string, or GraphQL block — which is how `Known Fraudster Risk` reached
+ * jsDelivr as a package name.
+ */
 const IMPORT_SPECIFIER =
-	/(?:import|export)\b[^'";]*?\bfrom\s*['"]([^'"]+)['"]|(?:import|require)\s*\(\s*['"]([^'"]+)['"]\s*\)|import\s*['"]([^'"]+)['"]/g;
+	/(?:import|export)\b[\w$\s{},*]*?\bfrom\s*['"]([^'"]+)['"]|(?:import|require)\s*\(\s*['"]([^'"]+)['"]\s*\)|import\s*['"]([^'"]+)['"]/g;
```
cb1kenobi: **Medium: import clause regex drops any import whose named block contains a comment**

The new `[\w$\s{},*]*?` clause excludes `/`, so a `//` or `/* */` comment between the braces breaks the match entirely — the whole import is missed, not just truncated. Confirmed against this exact module:

```js
findAcquirableSpecifiers(`import {\n\tuseState, // hooks\n\tuseEffect,\n} from 'react'`)
// => [] on this PR, ['react'] on main
```

Same result for a `/* block */` comment in the same position. None of the 41 new cases cover a comment inside the import clause, so this shipped untested. Effect: a real file with a commented named import silently stops getting type acquisition for that package (spurious "cannot find module", same degraded-but-safe failure mode as any unacquired package — not a new leak).

Suggested fix: let the repeated clause also consume recognized comment tokens instead of admitting a bare `/`, e.g. add `\/\/[^\n]*|\/\*[\s\S]*?\*\/` as extra alternatives inside the `*?` group so `//...` and `/* ... */` are matched as whole units (bounded by newline / `*/`) rather than opening up an unbounded `/`. That keeps the fix's guarantee — no path back into unrelated prose — while restoring support for a common real-world pattern.

—
_Generated by Barber AI_

dawsontoth: Confirmed and fixed in 66747d99 — thanks, this was a real regression and it shipped untested.

Reproduced exactly as reported: `findAcquirableSpecifiers("import {\n\tuseState, // hooks\n\tuseEffect,\n} from 'react'")` returned `[]`, and the same for a `/* block */` in that position. Since the clause matcher fails outright rather than truncating, the package just stops getting types.

**One deviation from the suggested fix**, because I tested it before adopting it. A bare `\/\/[^\n]*` alternative is *backtrackable*: `[^\n]*` can give back characters, so the matcher stops halfway through the comment and takes the `from "…"` out of the rest of it. That reopens this PR's leak:

| clause matcher | `export { Dog }` ⏎ `// each row is read from "customers"` | annotated import |
| --- | --- | --- |
| `stage` | **`customers`** | ✅ |
| this PR, as reviewed | ✅ none | ❌ missed |
| bare `\/\/[^\n]*` | **`customers`** | ✅ |
| shipped | ✅ none | ✅ |

I used a single lowercase table name rather than `Known Fraudster Risk` deliberately: `customers` satisfies npm's name grammar, so `isAcquirablePackage` waves it through and the clause matcher is the only gate left. The multi-word phrase is caught by the grammar check either way, which is what masked this when I first wrote the cases.

So both comment forms are now matched as **indivisible units** — the line form pinned to its line end with `(?=\n|$)`, the block form written so it cannot cross its terminator:

```
(?:[\w$\s{},*]|\/\/[^\n]*(?=\n|$)|\/\*(?:[^*]|\*(?!\/))*\*\/)*?
```

Backtracking to a shorter comment makes the lookahead (or the `*/`) fail, so a partial comment is unreachable. The alternatives are also disjoint on their first character — only `/` starts a comment, and `/` isn't in the class — so the repetition stays effectively deterministic; 20k-token clause garbage and 5k-line comment spam both scan in well under a millisecond.

Tests: 12 new cases, and I verified each group fails where it should.

- 6 annotated-import forms (line and block inside a named block, `/** */` and `//` before `from`, comment before a `type` modifier, CRLF) — **all 6 fail on the clause you reviewed**.
- 6 prose shapes keyed on `customers` — **4 fail on the backtrackable alternative, all 6 on `stage`**. Includes the `.d.ts` jsdoc case, which matters given the scanner also runs over every downloaded declaration file.

`259 files / 1876 passed, 11 skipped`; `tsc -b`, `oxlint`, `dprint check` clean.

Still unsigned, for the same headless-pinentry reason as the original commit — `gpg: signing failed: Inappropriate ioctl for device`.

dawsontoth: Fixed in 66747d99 — the clause now consumes comments as indivisible units (line form pinned with `(?=\n|$)`, block form unable to cross its terminator), because the bare `\/\/[^\n]*` alternative backtracks and reopened the leak on `export { Dog }` + comment; 12 new cases, details in [the reply above](https://github.com/HarperFast/studio/pull/1568#discussion_r3667690425).

src/features/instance/applications/components/TextEditorView/harper-language/typeAcquisition.ts
```diff
@@ -63,13 +74,57 @@ function isAcquirablePackage(specifier: string): boolean {
...
+ * line end by a lookahead, and the block form cannot cross its terminator. That
+ * pinning is the whole point — a plain `//[^\n]*` backtracks, so the matcher could
+ * stop halfway through a comment and take the `from "…"` out of the rest of it,
+ * reopening the leak on input as ordinary as `export { Dog }` followed by a
+ * comment. Alternatives are disjoint on their first character (only `/` starts a
+ * comment), so the repetition stays linear.
+ */
 const IMPORT_SPECIFIER =
-	/(?:import|export)\b[^'";]*?\bfrom\s*['"]([^'"]+)['"]|(?:import|require)\s*\(\s*['"]([^'"]+)['"]\s*\)|import\s*['"]([^'"]+)['"]/g;
+	/(?:import|export)\b(?:[\w$\s{},*]|\/\/[^\n]*(?=\n|$)|\/\*(?:[^*]|\*(?!\/))*\*\/)*?\bfrom\s*['"]([^'"]+)['"]|(?:import|require)\s*\(\s*['"]([^'"]+)['"]\s*\)|import\s*['"]([^'"]+)['"]/g;
```
cb1kenobi: **Medium: comment alternatives make the scanner quadratic on an unterminated `/*` or `//`**

The new comment alternatives let the lazy clause match consume arbitrary text while hunting for the terminator. When a `/* ` (or a `//` at EOF with no trailing newline in a truncated file) never closes, each failed attempt starting at a subsequent `import`/`export` occurrence rescans forward toward end-of-string before giving up — O(occurrences × remaining length). Measured against the shipped regex:

```
n=4000   (92 KB, many `import /* unterminated`)  ->  269ms
n=8000   (184 KB)                                 -> 1125ms
n=16000  (368 KB)                                 -> 4610ms
```

quadrupling time on each doubling of input — confirmed quadratic. The pre-fix regex (excluding `/` entirely) fails these same inputs in <1ms, so this is newly introduced by the comment-matching change, and it contradicts the doc comment two lines up ("the repetition stays linear") and the file's own design intent ("any failure is swallowed so it can never break editing") — this doesn't fail, it hangs the main thread.

This isn't only reachable by a crafted file: `acquireApplicationTypes` joins every project script (each up to 512 KB via `MAX_WORKER_MODEL_CHARS`) into one string before scanning, and the same regex runs over every downloaded `.d.ts` via ata's `resolveDeps`/`preProcessFile` with no size guard applied before the scan. A single saved file with a mid-edit unterminated block comment, or a truncated/corrupted CDN download, is enough to trigger a multi-second freeze — no malicious input required.

Suggested fix: bound how far an unterminated comment can be searched before the alternative gives up, e.g. cap the block-comment body to a sane length (`(?:[^*]|\*(?!\/))*{0,2000}` — actually JS doesn't support bounding a group like that directly, so bound the alternative itself: `\/\*[^*]{0,2000}(?:\*(?!\/)[^*]*){0,50}\*\/`), or simpler, add a length guard in `findAcquirableSpecifiers`/`preProcessFile` that skips scanning (or scans a bounded prefix) when `code.length` exceeds a fixed ceiling, consistent with the existing `MAX_SPECIFIER_LENGTH`/`MAX_WORKER_MODEL_CHARS` bounds already in this file.

—
_Generated by Barber AI_

dawsontoth: Confirmed and fixed in 3c1a42c6 — bounded every clause alternative including the comment bodies (`MAX_IMPORT_CLAUSE = 4096`), so the 719 KB unterminated-`/*` case drops from ~22s to ~220ms; also fixes the *pre-existing* form of this — a run of clause-legal characters was ~2.5s before comments were admitted at all, now ~250ms — and 4 wall-clock tests fail with the bound removed.

src/features/instance/applications/components/TextEditorView/harper-language/typeAcquisition.ts
```diff
@@ -63,13 +74,90 @@ function isAcquirablePackage(specifier: string): boolean {
...
+ * stop halfway through a comment and take the `from "…"` out of the rest of it,
+ * reopening the leak on input as ordinary as `export { Dog }` followed by a
+ * comment. Every alternative is bounded by `MAX_IMPORT_CLAUSE`, including the
+ * comment bodies — an unterminated comment has no terminator to find, so an
+ * unbounded search for one is exactly what made this quadratic.
+ */
+const IMPORT_CLAUSE = String.raw`(?:`
+	+ String.raw`[\w$\s{},*]`
+	+ String.raw`|//[^\n]{0,${MAX_IMPORT_CLAUSE}}(?=\n|$)`
+	+ String.raw`|/\*(?:[^*]|\*(?!/)){0,${MAX_IMPORT_CLAUSE}}?\*/`
```
cb1kenobi: **Medium: bounding the comment alternatives with `{0,MAX_IMPORT_CLAUSE}` trades the quadratic slowdown for a stack overflow on a long run of large *terminated* comments**

The quadratic case (unterminated `/*`/`//`) is fixed and confirmed linear by re-benchmarking (doubling input size roughly doubles time, up to ~5.9MB / 1.8s — no regression). But chaining enough large, properly-terminated block comments back-to-back inside one `import`/`export` clause overflows V8's regex backtracking stack, even though each individual comment is legal and the repeat count stays under the outer `MAX_IMPORT_CLAUSE` bound:

```js
const bigComment = '/*' + 'x'.repeat(4090) + '*/';
findAcquirableSpecifiers('import ' + bigComment.repeat(1400) + ' nofrom\n');
// throws RangeError: Maximum call stack size exceeded
```

This reproduces on plain Node (same V8 engine as the browser tab) with the default stack size — no flags needed. It also throws when the clause eventually succeeds (`bigComment.repeat(1600) + " from 'react'"` also throws), so it isn't limited to the already-failing path. The round-1 (unbounded `*`/`*?`) version handles the identical input without error — the crash is a side effect of switching to bounded quantifiers, not something bounding was supposed to fix.

`acquireApplicationTypes` catches this (`console.warn`), so it won't crash the tab, but since it scans `sources.join('\n;\n')` in one pass, hitting this on any one file's clause aborts type acquisition for *every* joined project source in that pass — a bigger blast radius than the pre-fix "slow but eventually finishes" behavior for this shape of input. Project script files are capped at 512 KB (`MAX_WORKER_MODEL_CHARS`), well under the ~5.7 MB needed within a single clause, so this path looks unreachable from ordinary project files. It's less clearly bounded for downloaded `.d.ts` content, which this same file notes is scanned with no size guard — I haven't confirmed whether `@typescript/ata` caps declaration size before calling `preProcessFile` on it. The new wall-clock tests don't cover chained large terminated comments, so this wouldn't be caught by CI either way.

Suggested fix: cap the number of comment-alternative repetitions per clause independently of the outer length bound (e.g., a small fixed max on how many `//`/`/* */` comments one clause may contain), or replace the comment alternatives with a non-backtracking skip (`indexOf('*/')` / `indexOf('\n')`) instead of a quantified sub-pattern, so a long run of legal comments can't accumulate backtracking stack.

—
_Generated by Barber AI_

dawsontoth: @cb1kenobi do we think we will see this in the wild?

dawsontoth: Confirmed and fixed in 8a3c3c50 — took your second suggestion and replaced the quantified clause with a character walk (`readClauseSpecifier`), so there are no backtracking frames to accumulate and every terminator search is bounded; both repros now return `[]` in ~4ms. Worth noting the old bound was on *repetitions*, not characters, so 4096 reps x ~4 KB comments admitted ~16 MB — that is how the crash input got through, and `MAX_IMPORT_CLAUSE` now bounds characters. All 60 prior cases pass against both the walk and the round-2 regex, and the 2 new cases fail with `RangeError` on round 2.

src/features/instance/applications/components/TextEditorView/harper-language/typeAcquisition.ts
```diff
@@ -63,13 +74,244 @@ function isAcquirablePackage(specifier: string): boolean {
...
+			at++;
+			continue;
+		}
+		return undefined;
+	}
+	return undefined;
+}
+
+/** Keywords that can introduce a specifier — the only places the scanner starts work. */
+const STATEMENT_KEYWORD = /\b(?:import|export|require)\b/g;
```
kriszyp: This still finds `import`/`export`/`require` anywhere in the raw source, including inside comments and string, template, or regex literals. For example, `// import rows from "customers"` (or the same text inside a template) is scanned as the specifier `customers`; because that is npm-shaped, `isAcquirablePackage` accepts it and ATA can still send a user-authored table name to jsDelivr. This mitigates the observed multi-word value but does not close the privacy/root-cause bug. Could we make keyword discovery lexical-context-aware (or use a small proven module lexer) and add cases where complete import/require-looking text appears inside line/block comments, quoted/template strings, and regex literals?

src/features/instance/applications/components/TextEditorView/harper-language/typeAcquisition.ts
```diff
@@ -63,13 +74,244 @@ function isAcquirablePackage(specifier: string): boolean {
...
+	end: number;
+}
+
+/**
+ * A quoted module specifier at `at`, after optional whitespace. Bounded by npm's own
+ * name limit: past that the specifier is rejected anyway, so reading further would
+ * only turn a missing closing quote into another unbounded scan.
+ */
+function readQuotedSpecifier(code: string, at: number): { specifier: string; end: number } | undefined {
+	const opened = skipWhitespace(code, at, code.length);
```
kriszyp: `readClauseSpecifier` computes a 4 KiB `limit`, but once it sees `from`, this call skips whitespace all the way to `code.length`. Consequently, a quote beyond the clause bound can still resolve, and an input such as `` `${'import '.repeat(500)}from${' '.repeat(500_000)}'react'` `` makes hundreds of keyword candidates rescan the same 500 KB tail on the main thread. Please pass the clause limit through `readQuotedSpecifier`/`skipWhitespace` (and bound the call-form whitespace scans too), then add a behavioral case with `from` inside the bound but its quote beyond it.

kriszyp (approved):
Sounds good to me. Codex has suggestions. That don't sound that good to me :).

src/features/instance/applications/components/TextEditorView/harper-language/typeAcquisition.ts
dawsontoth: You are right that this was a mitigation, not the root fix — fixed in 642c659a: the scan now tracks lexical context and honours a keyword only in code, so comments, strings, templates, regex bodies and `obj.import` are skipped; 11 new cases cover exactly those, all keyed on the npm-shaped `customers` so they fail unless the scanner is genuinely lexical (12 fail without it). I evaluated es-module-lexer as you suggested and did not adopt it: it closes all four vectors, but it threw on three of four half-typed inputs I tried (`import { a } fr`, unterminated string, unclosed brace) and half-typed is an editor's normal state since ata scans on every keystroke, plus it reports no `require()` — so the walk stays, biased to treat anything uncertain as not-code, which loses an acquisition rather than leaking a name.

src/features/instance/applications/components/TextEditorView/harper-language/typeAcquisition.ts
dawsontoth: Confirmed and fixed in 642c659a — `readQuotedSpecifier` now takes the caller's bound and applies it to the whitespace run as well as the quotes, and the call-form scans are bounded the same way via `readSpecifierAfterKeyword`. Added the behavioural case you asked for (`from` inside the bound, quote beyond it) plus your 500 KB repro and the two call forms; all three fail without the threaded bound, and that repro now returns `[]` in 10ms.

src/features/instance/applications/components/TextEditorView/harper-language/typeAcquisition.ts
```diff
@@ -63,13 +74,448 @@ function isAcquirablePackage(specifier: string): boolean {
...
+	'do',
+	'else',
+	'yield',
+	'await',
+]);
+function canStartRegex(previousChar: number, previousWord: string): boolean {
+	if (isIdentifierChar(previousChar)) {
+		return REGEX_MAY_FOLLOW.has(previousWord);
+	}
+	return previousChar !== CLOSE_PAREN && previousChar !== CLOSE_BRACKET
```
kriszyp: This still scans valid regex bodies as code when the regex is the statement controlled by `if`/`while`/`for`. For example, `if (ready) /import rows from 'customers'/.test(source)` makes `findAcquirableSpecifiers` return `customers`, so the original user-string leak to jsDelivr remains. A closing `)` is not enough to distinguish division from a regex because it may close a control-flow condition. Please track the kind of paren/control-flow context (or conservatively treat this ambiguity as regex, consistent with the stated privacy-first policy) and add this case as a regression test.

src/features/instance/applications/components/TextEditorView/harper-language/typeAcquisition.ts
```diff
@@ -63,13 +74,448 @@ function isAcquirablePackage(specifier: string): boolean {
...
+		}
+
+		if (isIdentifierChar(charCode)) {
+			let wordEnd = at + 1;
+			while (wordEnd < code.length && isIdentifierChar(code.charCodeAt(wordEnd))) {
+				wordEnd++;
+			}
+			const word = code.slice(at, wordEnd);
+			// `obj.import` and `{ import: x }` are property names, not statements.
+			const isStatementKeyword = previousChar !== DOT
```
kriszyp, line 624: The scanner still considers keywords in JSX text to be statements. `const View = () => <p>we import rows from 'customers'</p>` returns `customers`, which recreates the same off-origin disclosure this PR is meant to close; `.jsx` and `.tsx` files are explicitly included by the caller. Please skip JSX text/attribute syntax (while continuing to scan JSX expression containers) or use a lexer/fallback design that can make this distinction, and add the example as a regression test.

src/features/instance/applications/components/TextEditorView/harper-language/typeAcquisition.ts
```diff
@@ -63,13 +74,448 @@ function isAcquirablePackage(specifier: string): boolean {
 	if (specifier === 'harper' || specifier === 'harperdb' || specifier.startsWith('node:')) {
 		return false;
 	}
-	return !NODE_BUILTINS.has(specifier) && !ASSET_OR_DATA.test(specifier);
+	if (NODE_BUILTINS.has(specifier) || ASSET_OR_DATA.test(specifier)) {
```
kriszyp: Exact membership does not filter builtin subpaths: `fs/promises`, `assert/strict`, and `timers/promises` all pass `NPM_SPECIFIER` and are still handed to jsDelivr even though they cannot name npm packages. Please normalize non-scoped specifiers to their first path segment for this check (and fill the missing bare builtins), with at least `fs/promises` covered by a test.

src/features/instance/applications/components/TextEditorView/harper-language/typeAcquisition.ts
```diff
@@ -63,13 +74,448 @@ function isAcquirablePackage(specifier: string): boolean {
...
+ * past it, and every keyword candidate would rescan that same run. The specifier body is
+ * additionally capped at npm's own name limit, past which it is rejected anyway.
+ */
+function readQuotedSpecifier(code: string, at: number, bound: number): { specifier: string; end: number } | undefined {
+	const opened = skipWhitespace(code, at, bound);
+	const quote = code.charCodeAt(opened);
+	if (opened >= bound || (quote !== SINGLE_QUOTE && quote !== DOUBLE_QUOTE)) {
+		return undefined;
+	}
+	const limit = Math.min(bound, opened + MAX_SPECIFIER_LENGTH + 2);
```
kriszyp: `214` is npm's limit for the package name, but this bounds the entire module specifier, including its deep-import subpath. A valid specifier such as `pkg/segment/...` with a short package name and total length 215+ is therefore discarded before ATA can resolve it. Please apply the 214-character rule only to the package portion and give the subpath a separate explicit safety bound (the existing 4 KB clause bound may already be sufficient), then test a long valid deep import.

kriszyp (commented):
codex has more to say on this.

src/features/instance/applications/components/TextEditorView/harper-language/typeAcquisition.ts
dawsontoth: Confirmed and fixed in a6c4c9b0 — `(` now records whether it opened an `if`/`while`/`for`/`switch`/`catch`/`with` head and `)` restores it, so a bare regex statement after a control-flow head is skipped rather than read as code, while division after a call still divides; 4 regression cases (if/while/for/nested), all of which fail without the tracking.

src/features/instance/applications/components/TextEditorView/harper-language/typeAcquisition.ts
dawsontoth: Confirmed and fixed in a6c4c9b0 — the context stack now models JSX tags and children alongside templates, skipping element text and attribute syntax while still scanning `{…}` containers so `import()` in one is found; I verified the caller filter is `\.(tsx?|jsx?|mjs|cjs|mts|cts)$` as you said. 6 regression cases including your example, a fragment, nested children and a `return` element (5 fail without it) plus type-argument/comparison cases so `Array<string>` and `a < b` are not mistaken for JSX. One correction for the record: a quoted JSX *attribute* value was already safe via string skipping — that case passes with or without the fix. Known limit in the safe direction, documented in the code: a generic arrow (`<T,>(x) => x`) reads as JSX and skips the remainder, losing acquisition rather than leaking a name.

src/features/instance/applications/components/TextEditorView/harper-language/typeAcquisition.ts
dawsontoth: Confirmed and fixed in a6c4c9b0 — the builtin check now compares the package portion (first segment, or first two when scoped), so every subpath form is covered without listing them; 10 cases including `fs/promises`, `assert/strict`, `timers/promises`, `stream/web`, `util/types`, `path/posix`, and 9 of them fail without the normalisation. Filled in the missing bare builtins too, but deliberately left out `test`, `sqlite` and `sea`: those are real npm packages, and listing them bare would cost those packages their types while their `node:` forms are already rejected — covered by a test.

src/features/instance/applications/components/TextEditorView/harper-language/typeAcquisition.ts
dawsontoth: Confirmed and fixed in a6c4c9b0 — the 214 rule now applies only to the package portion, with a separate 1 KB ceiling on the whole specifier; the read bound in `readQuotedSpecifier` had the same problem and was truncating such a specifier before the filter ever saw it, so that moved too. Tested a 244-char valid deep import (fails without the split) alongside a 215-char package name with and without a subpath, which must still be rejected.

PR #1571, opened 2026-07-28
feat(admin): Regions management screen (list, create, edit)

DavidCockerill:
## Summary

Adds a **Regions** section to the Fabric Admin area so admins can create, list, and edit deployment regions — including customer-scoped ones — without direct DB access.

- **List** with a row-filter box, and a **Scope** column showing `Public` or the scoped orgs as `org-id (name)`.
- **Create/edit modal** (react-hook-form + zod): Region ID (immutable after create), Region (a dropdown of known groupings), instance count, purchased block multiplier, latency description, per-provider preferred-location pickers, `Force locations`, `Active`, and the organization scope.
- **`MultiSelect`** built on the shared Radix dropdown (the repo has no combobox), with a filter and removable chips.
- `/admin` now lands on **Regions** (the first rail item); API Token moved to `/admin/api-token`.

Companion backend PR: **HarperFast/central-manager#518** — the `active` flag, `forceLocations`, and the name/latency validation all live there. It merged as `a06faeb7` on 2026-07-30, so `main` now carries both fields in `schema.graphql` and `adminValidation.js`.

**Deploying this screen requires the central-manager build that includes #518.** Landing is not deploying: against an older CM, Joi `stripUnknown` silently drops `forceLocations` and `active`, the request still succeeds, and this form reports the edit as saved — so a region an admin just marked inactive stays customer-visible. Ship CM first, or the toast lies. (Raised by @kriszyp; this paragraph previously said the PR could merge independently and described that silent discard as expected, which was wrong to leave standing.)

## Purpose

Region management was DB-only. This is the UI half of the Fabric Admin region CRUD work.

## Where to look

**1. `MultiSelect` has two modes and the difference matters.** With `allowRepeats` (both location pickers), preferred locations are an **ordered placement list, not a set** — clicking a location again appends another copy, each copy is its own chip, and a chip's `X` removes just that copy. Without it (the organizations picker) a second click unselects, as normal. Worth a look at `removeAt(index)` and the `${value}-${index}` chip keys, which is what makes duplicates addressable.

**2. Repeated locations only take effect when `Force locations` is on.** `addDedicatedHosts.getLocations` re-pushes a duplicate only in the `forceLocations` branch; otherwise extras are deduped during placement. **The UI deliberately does not warn about this** (explicitly requested) — so an admin can add 3× Tokyo with `Force locations` off and get one host. Flagging because it's a real footgun a reviewer might otherwise read as a bug.

**3. Latency uniqueness is only checked when the value changes** (`RegionFormModal.onSubmit`). The seeded catalog already ships a collision — `us-ne-1` and `us-se-1` both use `"49ms, small distribution"` — and this form submits every field, so an unconditional pre-check made both rows uneditable. Mirrors the server-side fix in #518; the server 409 remains the source of truth.

**4. A public region must send `organizationIds: null`, never `[]`.** `Region.get` treats `[]` as "scoped to zero orgs", which would hide the region from everyone.

**5. Collection GETs need a trailing slash.** `/Admin/Region/` and `/Admin/Organization/` return the records array; **without** the slash Harper returns a collection *descriptor object*. This crashed the page during development (`regions.map is not a function`) and mocked unit tests can't catch it. `/Location` is exempt because its resource overrides `get()` and always returns an array.

**6. Route index swap.** `regionsRoute` now holds `path: '/'` and API Token moved to `api-token`, so `/admin` opens the first rail item rather than hardcoding a deeper link in the navbar. Both routes declare `getParentRoute: () => adminLayoutRoute` and are mounted via `adminLayoutRoute.addChildren([...adminRoutes])` — the lockstep the repo's `CLAUDE.md` warns about. **Note: `/admin` previously served API Token, so existing bookmarks now land on Regions.**

## Testing

- 17 unit tests across the feature (mutations, zod schema, and `MultiSelect` — including the repeat/remove-by-index behavior, click-order preservation, and that the non-repeat picker still toggles). Router regression tests still pass (11).
- `tsc -b`, `oxlint`, and `dprint` clean.
- **Manually verified against dev** end-to-end: list renders all 30 regions; create/edit modal populates orgs and locations from live data; repeat-a-location and remove-one-copy both behave; `/admin` → Regions and `/admin/api-token` → API Token with correct rail highlighting.
- Not verified: the **Inactive** badge / dimmed row, because dev's CM doesn't yet have the `active` field (needs #518 deployed).

## Docs

No documentation PR — this is an internal Fabric-admin screen, not public API or user-facing product surface.

## Review provenance

Cross-model review ran, with caveats:
- **The Gemini leg never ran** — `agy` rejected every available model ("Selected model is not supported in the selected location"), so there is **no outside-model performance/maintainability or TypeStrip coverage** on this diff.
- **The Codex leg hung after ~10 minutes with an incomplete sweep** — the MultiSelect index logic, routing lockstep, and zod typing were covered only by the Claude domain pass. Extra human attention on items 1 and 6 is worthwhile.
- One reported blocker was **rejected with evidence**: a missing trailing slash on `/Location`. `Location.get()` never calls `super.get()` — it returns an array unconditionally — and both paths were probed live (`isArray: true, len: 158`). Both review legs generalized from repo convention without testing it.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code) (model: Claude Opus 4.8). Reviewed by Codex (partial) + a Harper-domain pass; the Gemini leg was unavailable.


src/features/admin/regions/components/RegionFormModal.tsx
```diff
@@ -0,0 +1,385 @@
...
+			void queryClient.invalidateQueries({ queryKey: regionsQueryKey });
+			onOpenChange(false);
+		};
+
+		if (isEdit) {
+			const { id: _id, ...changes } = values;
+			updateRegion({ id: region.id, changes: { ...changes, organizationIds } }, { onSuccess });
+		} else {
+			createRegion({ ...values, organizationIds }, { onSuccess });
+		}
```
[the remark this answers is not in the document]
DavidCockerill: Applied in `780e7dfc`, but correcting the rationale: there is **no compile error** here. TS 4.4+ narrows through a `const` alias of a condition (`const isEdit = !!region`), so `region.id` type-checks fine under `strictNullChecks` — `tsc -b` was clean on this branch before the change.

Taking the suggestion anyway, for a different reason: narrowing on `region` directly doesn't depend on that aliased-condition inference surviving a future refactor of `isEdit` (which is also used for UI copy). Cheap robustness, not a bug fix.

— Claude Opus 4.8

src/features/admin/regions/components/RegionFormModal.tsx
```diff
@@ -0,0 +1,385 @@
...
+											<Input
+												type="number"
+												min={1}
+												step={1}
+												name={field.name}
+												ref={field.ref}
+												onBlur={field.onBlur}
+												value={Number.isNaN(field.value) ? '' : field.value}
+												onChange={(e) => field.onChange(e.target.valueAsNumber)}
+											/>
```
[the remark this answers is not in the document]
DavidCockerill: Addressed in `780e7dfc` with a tighter equivalent: `Number.isFinite(field.value) ? field.value : ''`, which covers `NaN`, `undefined`, `null` and `Infinity` in one predicate.

Worth noting the path isn't currently reachable: `defaultValues`/`form.reset` both go through `toFormValues`, which always supplies a number (`region?.instanceCount ?? 1`), and clearing the input yields `NaN` from `valueAsNumber` — never `undefined`. I did verify your literal suggestion compiles (I expected a TS2367 no-overlap error on `field.value === undefined` and was wrong). Applied as defensive hardening.

— Claude Opus 4.8

src/features/admin/regions/components/RegionFormModal.tsx
```diff
@@ -0,0 +1,385 @@
...
```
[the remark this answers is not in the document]
DavidCockerill: Same fix as the sibling thread — `Number.isFinite` guard applied in `780e7dfc`. See that thread for why the undefined/null path isn't currently reachable.

— Claude Opus 4.8

src/features/admin/regions/components/RegionFormModal.tsx
```diff
@@ -0,0 +1,385 @@
...
+								<FormItem>
+									<div className="flex items-center gap-2">
+										<FormControl>
+											<Input
+												type="checkbox"
+												className="size-4"
+												checked={field.value}
+												onChange={(e) => field.onChange(e.target.checked)}
+											/>
+										</FormControl>
```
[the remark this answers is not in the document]
DavidCockerill: Good catch — applied in `780e7dfc`. Confirmed I was using the custom `Input` here, which carries text-input styling (`h-9`, `w-full`, padding, focus/aria-invalid rings). It only looked right because `cn()`/tailwind-merge resolved my `size-4` over `h-9 w-full` — relying on that is exactly the fragility you're describing. Now a native `<input type="checkbox">`.

Note the repo does have precedent for the other pattern (`AddOrganizationRoleModal` uses `<Input type="checkbox">`); I'm deviating deliberately here.

— Claude Opus 4.8

src/features/admin/regions/components/RegionFormModal.tsx
```diff
@@ -0,0 +1,385 @@
...
+							render={({ field }) => (
+								<FormItem>
+									<div className="flex items-center gap-2">
+										<FormControl>
+											<Input
+												type="checkbox"
+												className="size-4"
+												checked={field.value}
+												onChange={(e) => field.onChange(e.target.checked)}
+											/>
```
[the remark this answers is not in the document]
DavidCockerill: Same fix as the sibling thread — native checkbox applied in `780e7dfc`.

— Claude Opus 4.8

src/features/admin/regions/components/RegionFormModal.tsx
```diff
@@ -0,0 +1,387 @@
...
+		const latencyChanged = !region || norm(region.latencyDescription) !== norm(values.latencyDescription);
+		const others = regions.filter((r) => r.id !== region?.id);
+		if (latencyChanged && others.some((r) => norm(r.latencyDescription) === norm(values.latencyDescription))) {
+			form.setError('latencyDescription', { message: 'A region with this latency description already exists' });
+			return;
+		}
+
+		// A public region must send organizationIds: null. An empty array would hide it from every
+		// org (Region.get treats [] as "scoped to these zero orgs"), which is never what we want.
+		const organizationIds = values.organizationIds.length > 0 ? values.organizationIds : null;
```
kriszyp, line 119: This always collapses an empty selection to `null` (public), even when the region being edited started out as `organizationIds: []` (hidden from every org). The commit that fixed `RegionScope`'s label (`0ad4ad4`) established that `[]` is a real, distinct, reachable state — "takes pre-existing data from another client to hit" — but this form can't preserve it: `toFormValues` maps both `null` and `[]` to the same `[]` default, and this line maps `[]` back to `null` on every save.

Concretely: an admin opens Edit on a region that has `organizationIds: []`, doesn't touch the Organizations field at all (say they're only flipping Active), and hits Save — the PATCH goes out with `organizationIds: null`, silently turning a region that was hidden from every customer into one that's public to all of them, since this form always submits every field.

Since the picker itself can't represent "explicitly hidden from everyone" as distinct from "public, not yet scoped" (both look like an empty MultiSelect), the round-trip is inherently lossy as written. Worth closing before merge — e.g. only normalize `[] → null` when the org selection was actually touched (mirroring the `latencyChanged`-only-when-different pattern a few lines up), or give "hidden from everyone" its own explicit control instead of overloading the empty-selection state.

src/features/admin/regions/queries/getOrganizations.ts
```diff
@@ -0,0 +1,31 @@
...
+	return name ? `${id} (${name})` : id;
+}
+
+/**
+ * GET /Admin/Organization → every organization (super-user only), for the region form's customer
+ * scope picker. The endpoint isn't in the generated types, so the URL and response are cast.
+ */
+export async function getOrganizations(): Promise<AdminOrganizationSummary[]> {
+	// Trailing slash required to get the records array rather than the collection descriptor.
+	const { data } = await apiClient.get('/Admin/Organization/' as '/Cluster/');
```
kriszyp: This fetches the whole `/Admin/Organization/` collection with no query params, which duplicates `getAllOrganizations` (`src/features/organizations/queries/getAllOrganizations.ts`) hitting the same endpoint — that one was deliberately built with server-side pagination (`limit(...)`) and a name filter (`name=ct=...`) specifically because the admin org list doesn't fit in one response. Fetching everything here for a client-filtered `MultiSelect` will grow the payload and the number of `DropdownMenuCheckboxItem` DOM nodes 1:1 with total organization count (`MultiSelect.tsx` has no virtualization), which won't scale the same way the existing paginated view does.

If the org count is small today this is a non-issue, but consider reusing the paginated/search endpoint pattern (server-side `name=ct=` filtering as the user types in the MultiSelect's filter box) rather than a second bespoke unpaginated fetcher for the same collection.

kriszyp (commented):
Cool stuff.
<sub>🤖 Reviewed with Claude Sonnet</sub>

src/features/admin/regions/components/RegionFormModal.tsx
DavidCockerill: Fixed in `16359659`. You're right, and this was my loose end: the earlier commit fixed `RegionScope`'s *label* but left the round-trip lossy, and my justification ("the conversion is visible by contrast in the table") doesn't hold — the admin never has to look at the table, and nothing warns at save time. Accidental exposure is the wrong direction to fail in.

Took your first suggestion, mirroring the `latencyChanged` pattern: `react-hook-form`'s `formState.dirtyFields.organizationIds` gives the exact signal, so an untouched field resubmits the stored value verbatim while a deliberately-cleared selection still means public (matching the modal's helper text). Pulled the rule into `resolveOrganizationScope` with unit tests rather than leaving it inline, since the `[]`-vs-`null` distinction is the part worth locking down.

Not claiming the picker can now *express* "hidden from everyone" — it still can't, so creating that state from this UI remains impossible. This only stops the form destroying it. If it should be a first-class choice, that's the explicit control you floated, and I'd rather do it as its own change.

— Claude Opus 4.8

src/features/admin/regions/queries/getOrganizations.ts
DavidCockerill: Partly fixed in `16359659`, and you undersold it — reading `getAllOrganizations` turned up two correctness bugs alongside the scale point:

- **No `status=ne=DELETED`**, so the picker offered *terminated* organizations as scoping targets.
- **No `sort(name)`**, so the list rendered in table order — which is exactly the jumbled ordering I'd seen while testing and hadn't questioned.

Now filters DELETED, sorts by name, and caps at `limit(0,500)` with a comment pointing at the server-side-filtering alternative.

What I did **not** do is consolidate onto the paginated fetcher. Driving `name=ct=` server-side from the MultiSelect's filter box means changing its `options` contract from a static array to an async source — a real refactor of a component this PR also introduces, and I'd rather not fold it in here. Leaving that as a follow-up.

One correction: you attribute the pagination to the org list not fitting in one response, but the existing comment ties `ALL_ORGANIZATIONS_PAGE_SIZE = 12` to card layout ("roughly one screen of cards"). Doesn't change the conclusion — the duplication and the missing filters were real either way.

— Claude Opus 4.8

dawsontoth:
@DavidCockerill can you rebase this on the latest stage? Then we'll push it to dev so I can see it. Not being able to log in to an admin account locally is quite... vexing.

src/features/admin/components/AdminShell.tsx
```diff
@@ -1,16 +1,19 @@
...
  * (matching the token endpoint's SSO-session contract — see isFabricAdmin); the
  * dashboard route guard already handles the unauthenticated redirect.
  */
+// First item is the section index (/admin) so the navbar's Admin link opens it; `exact` keeps it
+// from also matching the sibling routes nested under /admin.
 const items: SubNavItem[] = [
-	{ to: '/admin', label: 'API Token', icon: KeyRoundIcon, exact: true },
+	{ to: '/admin', label: 'Regions', icon: GlobeIcon, exact: true },
+	{ to: '/admin/api-token', label: 'API Token', icon: KeyRoundIcon },
 	{ to: '/admin/notifications', label: 'Notifications', icon: BellIcon },
```
dawsontoth, line 14: Pref: I think notifications would make for a better, lighter weight landing page for admin. It would also surface outages for us immediately too. Then regions and api-tokens and whatever else can show up under it.

src/features/admin/regions/queries/getLocations.ts
```diff
@@ -0,0 +1,23 @@
...
+import { AdminLocation } from '@/integrations/api/api.patch';
+import { queryOptions } from '@tanstack/react-query';
+
+/**
+ * GET /Location → every deployment location. Used to populate the region form's per-provider
+ * location pickers so an admin can only pick values the Location table actually holds (the
+ * RegionAdmin endpoint rejects unknown preferred locations with a 400).
+ */
+export async function getLocations(): Promise<AdminLocation[]> {
+	const { data } = await apiClient.get('/Location' as '/Cluster/');
```
dawsontoth: Does this actually return Clusters? What's with `as /Cluster/`? That's misleading, and covered below by `as unknown`

src/features/admin/components/AdminShell.tsx
DavidCockerill: Done in `be1186df` — Notifications is now the `/admin` index, with Regions at `/admin/regions` and API Token at `/admin/api-token` beneath it. Agreed on the reasoning: surfacing outages the moment someone lands beats opening on a management screen.

Worth flagging that this was my call to begin with and I'd flagged it to David as unreviewed — the rail only had two items when I picked the order, and Notifications arrived from stage while this branch was open. Your preference wins on both counts.

— Claude Opus 4.8

src/features/admin/regions/queries/getLocations.ts
DavidCockerill: No, it does not return Clusters — and you found something worse than a misleading cast. **The cast was never needed.** `/Location/`, `/Admin/Region/`, `/Admin/Region/{id}` and `/Admin/Organization/` are all in `api.gen.d.ts`. I'd copied the pattern from the API-token mutation without checking whether my own endpoints were typed, then wrote a comment asserting *"the endpoint isn't in the generated types"* on four files — which was simply false, and I repeated it in the PR description.

Fixed in `be1186df`: real paths, no `/Cluster/` anywhere. Where a path needs interpolation the cast now names a related path (`` `/Admin/Region/${id}` as '/Admin/Region/{id}' ``), matching `getOrganizationQuery` and friends.

You're right that the response `as unknown` covers the return type, but it doesn't cover the *argument* — `TypedAxios` constrains the URL, so an interpolated path still needs the literal. I kept the response casts and gave them the real reason, which I verified this time rather than assuming: the generated schema has no `active` field (added by the companion CM PR) and types `organizationIds?: string[]` with no `| null`, while the API returns null for a public region.

— Claude Opus 4.8

DavidCockerill:
@dawsontoth rebased onto latest stage and pushed — `be1186df` is a merge of `stage` (13 commits) plus your two review items. Full suite green: 1993 tests, tsc/oxlint/dprint clean, no conflicts.

Both threads answered inline:
- **Notifications is now the `/admin` landing page**, Regions and API Token beneath.
- **`as '/Cluster/'` is gone.** You were right, and it was worse than misleading — the cast wasn't needed at all, since these paths *are* in the generated types. My comments claiming otherwise were wrong on four files.

Leaving the dev deploy to you — I don't want to trigger that without David saying so.

On not being able to log in locally: that's the same wall I hit building this. Local studio can reach a dev CM, but the session cookie won't cross from `localhost` (`SameSite`/third-party blocking), and dev's CORS allowlist won't accept a `*.harperfabric.com` loopback origin, so you can't get same-site either. The durable fix is a loopback dev hostname under `harperfabric.com` (public A record → 127.0.0.1) plus local HTTPS, with that origin added to dev's CORS allowlist and the Google OAuth client. Happy to write that up separately — it'd unblock local admin work for everyone rather than just this PR.

🤖 Response drafted by Claude Opus 4.8

dawsontoth:
@DavidCockerill you say you rebased, but you actually merged in. When it comes time to merge to stage, it will be done as a rebase, so go ahead and clean up the history here for better accuracy. (Talking to dAIvid there)

DavidCockerill:
@dawsontoth fair correction — you asked for a rebase and I merged, then called it a rebase in my own comment. Both wrong. History is now linear.

`28d74b5e` is a single commit on top of current `stage` (which had moved another 4 commits since the last push). No merge commits.

Why one commit rather than replaying all 22: a straight `git rebase origin/stage` conflicts on the very first commit — dropping the two merge commits discards their conflict resolutions, and `AdminShell.tsx` / `routes.ts` were rewritten several times as the rail order changed twice, so it would have meant resolving the same files repeatedly with a real chance of a bad intermediate state. Squashing keeps every commit on `stage` building and testing green. Say the word if you'd rather see it split into logical commits instead — happy to redo it.

Verified the rewrite preserved the work exactly: `git diff` of my 19 files against the pre-rebase commit is empty. Full suite green on the new head — 2002 tests, tsc/oxlint/dprint clean.

**Heads-up on stale links:** the force-push invalidated the SHAs cited in earlier review replies. Mapping for anyone following the threads:

| Cited | Was |
|---|---|
| `780e7dfc` | Gemini review fixes (native checkboxes, `Number.isFinite`, narrowing on `region`) |
| `0ad4ad41` | `RegionScope` public-vs-hidden label fix |
| `16359659` | kriszyp's items — untouched org scope preserved, org picker filtered |
| `be1186df` | your items — Notifications at the index, `as '/Cluster/'` removed |

All of it is in `28d74b5e` now; the commit message summarises the same ground.

🤖 Response drafted by Claude Opus 4.8

dawsontoth (approved):
Looks sick.

src/features/admin/regions/queries/getRegions.ts
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * GET /Admin/Region → every region (super-user only).
+ *
+ * The response is cast: the generated schema predates the `active` flag and types `organizationIds`
+ * as non-nullable, while the API returns null for a public region.
+ */
+export async function getRegions(): Promise<AdminRegion[]> {
+	// Trailing slash is required: GET /Admin/Region (no slash) returns Harper's collection
+	// descriptor, /Admin/Region/ returns the records array.
+	const { data } = await apiClient.get('/Admin/Region/');
```
kriszyp, line 14: This does not currently return every region. In central-manager `origin/main`, `RegionAdmin.get()` delegates to `Region.get()`, whose customer-facing visibility check receives no organization ID and filters out every row with a non-empty `organizationIds`. The companion #518 branch still has that inheritance path.

Concretely, creating a scoped region succeeds, this mutation invalidates the query, and the next `/Admin/Region/` response omits the new row, so it disappears from the management screen and cannot be edited. Please fix the root contract in central-manager so the admin GET bypasses customer visibility, with an integration test containing a scoped row, and make that backend change a deployment prerequisite for this screen.

src/features/admin/regions/index.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+export function RegionScope(
+	{ organizationIds, orgNameById }: { organizationIds?: string[] | null; orgNameById: Map<string, string> },
+) {
+	// null/absent is public; an empty array is the opposite — Region.get reads it as "scoped to these
+	// zero orgs", so no organization can select it.
+	if (organizationIds == null) {
+		return <>Public</>;
+	}
+	if (organizationIds.length === 0) {
```
kriszyp: Current central-manager does not give `[]` this meaning. Commit `d0ca798d` on `main` standardized `isRegionVisibleToOrg()` so only a non-empty array restricts access; both `null` and `[]` are public, and companion #518 retains that behavior. An existing `[]` row is therefore shown here as “None (hidden)” while every organization can actually select it.

Please render both null and empty as Public and remove the null-vs-empty preservation machinery in `organizationScope.ts`. If “hidden from everyone” is a product state we still need, model it explicitly in the server contract (the new `active` flag may already be that state) rather than overloading an empty relationship list.

src/features/admin/regions/components/RegionFormModal.tsx
```diff
@@ -0,0 +1,390 @@
...
+			toast.success(isEdit ? 'Region updated' : 'Region created');
+			void queryClient.invalidateQueries({ queryKey: regionsQueryKey });
+			onOpenChange(false);
+		};
+
+		// Narrow on `region` itself rather than the `isEdit` alias — equivalent today, but it doesn't
+		// depend on TS's aliased-condition inference surviving a refactor of `isEdit`.
+		if (region) {
+			const { id: _id, ...changes } = values;
+			updateRegion({ id: region.id, changes: { ...changes, organizationIds } }, { onSuccess });
```
kriszyp, line 131: The commit message links companion central-manager PR #518 but says this can merge independently. It cannot safely deploy independently: current central-manager `main` omits both `forceLocations` and `active` from `RegionAdmin` validation (and `active` from the Region schema), while validation uses `stripUnknown`. Studio therefore shows a success toast even though both values are silently discarded; a supposedly inactive region remains customer-visible.

Please make #518 landing and deployment an explicit prerequisite, or capability-gate/hide these controls until the API supports them. A cross-repo contract test that creates and patches both fields would keep the UI from reporting success against a server that cannot persist them.

src/features/admin/regions/RegionFormSchema.ts
```diff
@@ -0,0 +1,34 @@
...
+	instanceCount: z.number({ error: 'Enter a whole number' }).int('Must be a whole number').min(1, 'Must be at least 1'),
+	purchasedBlockMultiplier: z
+		.number({ error: 'Enter a whole number' })
+		.int('Must be a whole number')
+		.min(1, 'Must be at least 1'),
+	latencyDescription: z.string().trim().min(1, 'Latency description is required'),
+	linodePreferredLocations: z.array(z.string()),
+	gcpPreferredLocations: z.array(z.string()),
+	// When true, deployments are pinned to the preferred locations rather than treating them as hints.
+	forceLocations: z.boolean(),
```
kriszyp, line 27: `forceLocations: true` is accepted even when a provider's preferred-location array is empty. In central-manager placement, Force makes that array the complete ordered location plan; for an Akamai organization with no Linode choices (or a GCP organization with no GCP choices), the plan is empty and provisioning fails with `No valid locations found`.

Please enforce the invariant in the backend and mirror it in this form: every provider that can provision into the region needs at least one preferred location when Force is enabled. If organization scope determines the applicable provider, the form will need provider/channel data or a deliberately stricter rule requiring both lists.

src/features/admin/regions/queries/getOrganizations.ts
```diff
@@ -0,0 +1,41 @@
...
+export const ORGANIZATION_PICKER_LIMIT = 500;
+
+/**
+ * GET /Admin/Organization → organizations for the region form's customer scope picker (super-user
+ * only). Terminated orgs keep a DELETED row, so they're filtered out — scoping a region to one is
+ * meaningless. The response is cast down to the two fields the picker uses.
+ */
+export async function getOrganizations(): Promise<AdminOrganizationSummary[]> {
+	// Trailing slash required to get the records array rather than the collection descriptor.
+	const query = ['status=ne=DELETED', 'sort(name)', `limit(0,${ORGANIZATION_PICKER_LIMIT})`].join('&');
```
kriszyp: This bounds the payload and DOM work by silently discarding every active organization after the first 500 names. At 501+ organizations, an admin cannot assign a later organization at all; existing scopes outside the window lose their display names, and filtering the regions table by those names stops working.

Please use the existing server-side paginated/name-search pattern from `getAllOrganizations.ts` and hydrate already-selected organization IDs explicitly. Virtualizing the option list would address the DOM cost without sacrificing data completeness.

src/features/admin/regions/components/MultiSelect.tsx
```diff
@@ -0,0 +1,166 @@
...
+				</DropdownMenuTrigger>
+				<DropdownMenuContent
+					className="w-(--radix-dropdown-menu-trigger-width) max-h-72 overflow-y-auto"
+					align="start"
+				>
+					<input
+						ref={inputRef}
+						value={filter}
+						onChange={(e) => setFilter(e.target.value)}
+						onKeyDown={(e) => e.stopPropagation()}
```
kriszyp: The menu deliberately moves focus into this filter, but stopping every keydown prevents Radix's roving-focus handling from seeing ArrowDown or Tab. With the menu open, focus remains on the input and a keyboard user cannot reach or activate any checkbox option; the component is effectively pointer-only.

Please use a combobox/listbox pattern, or handle navigation keys by explicitly moving focus into the filtered options while only suppressing the printable keys that conflict with Radix typeahead. Add a keyboard test that opens the picker, filters, moves to an option, and selects it.

src/features/admin/routes.ts
```diff
@@ -8,19 +8,28 @@ export const adminLayoutRoute = createRoute({
 	component: AdminShell,
 });
 
-const apiTokenRoute = createRoute({
+// The index route is whatever sits at the top of AdminShell's rail, so landing on /admin (the
+// navbar's Admin link) always opens the first section. Keep the two in step when reordering.
+const notificationsAdminRoute = createRoute({
 	getParentRoute: () => adminLayoutRoute,
 	path: '/',
```
kriszyp, line 16: Making Notifications the Admin landing page also removes the explicit `/admin/notifications` route that exists on `stage`, so existing bookmarks and open links stop matching a child route after deployment.

Please keep `/admin/notifications` as the canonical route and redirect `/admin` to it, or retain an alias/redirect while changing the default. That preserves the stable deep link without changing the requested Regions route.

kriszyp (commented):
Some more findings, but looks good to me.

<sub>🤖 Reviewed with Claude Sonnet</sub>

src/features/admin/regions/queries/getRegions.ts
DavidCockerill: Thanks for digging into the cross-repo path — that's exactly the failure mode worth checking. I traced it, and I don't think this one lands: `RegionAdmin` doesn't inherit from the customer-facing resource.

On `origin/main`, `RegionAdmin.js:5` destructures the **base table**:

```js
const { Region, Location } = tables;
...
export class RegionAdmin extends Region {   // ← tables.Region
```

The customer resource is a separate class over that same table, `src/resources/Region.js:17`:

```js
export class Region extends tables.Region {
```

So `RegionAdmin` and the customer `Region` are siblings, not parent and child. `super.get(target)` at `RegionAdmin.js:52` resolves to `tables.Region.get`, which has no visibility filter — `isRegionVisibleToOrg` is only reached through the customer resource's own `get`. Devin-Holland reached the same conclusion independently over on #518.

There's live confirmation too: the `david-test` region on dev has 4 entries in `organizationIds` and it renders in the admin table — that's how the extra-orgs bug David spotted earlier was visible at all. The customer filter would have removed that row.

Leaving this unresolved for you. Happy to be shown wrong if you're seeing a different inheritance path.

— Claude Opus 5

src/features/admin/regions/index.tsx
DavidCockerill: You're right, and thank you — this is the important one. Verified on `main`: `isRegionVisibleToOrg` only restricts when the array is non-empty, so `null` and `[]` are both public, and "None (hidden)" asserted the opposite of the actual behaviour.

Worth recording how it got there. The label was originally "Public" for both, then changed to "None (hidden)" on a review finding that was accurate for the pre-`d0ca798d` code. When the stage merge brought the `isRegionVisibleToOrg` refactor in, the semantics changed underneath the label and I never re-checked. I even appended a unit test to the merged `regionVisibility.test.js` asserting "empty organizationIds is visible to everyone" without connecting it back to this.

Fixed in 9c114b35, as you suggested:

- `RegionScope` renders **Public** for both `null` and `[]`.
- `organizationScope.ts` and its test are deleted — that machinery existed only to preserve a distinction that doesn't exist.
- The form sends `null` when the picker is empty.
- The `RegionScope` test now asserts `[]` → Public.

Agreed on the second point as well: `active: false` is the explicit "not offered for new deployments" state, so there's no need to overload an empty relationship list.

— Claude Opus 5

src/features/admin/regions/components/RegionFormModal.tsx
DavidCockerill: Agreed on the substance, and the timing has moved since you wrote this: #518 merged as `a06faeb7` (2026-07-30 15:35 UTC), so `main` now carries `forceLocations` and `active` in both `schema.graphql` and `adminValidation.js`. The `stripUnknown` silent-discard window you describe is closed on `main`.

The deployment-ordering point stands, though, and it's the right call: this screen must not reach an environment ahead of central-manager. I'm flagging it to David as a deploy-order prerequisite rather than something to gate in code, since the gap is now purely "which environment has which build" rather than a missing API contract.

On the contract test — #518 does add integration coverage for creating and patching both fields, but to be straight with you: **those integration tests have never been executed.** They need a live central-manager, and CM's CI runs the `unit` suite only. The 11 unit tests in `tests/unit/regionAdminValidators.test.js` do run and pass. A dev run is still owed, and it would also settle whether the stale-stored-location trap is live or only latent.

— Claude Opus 5

src/features/admin/regions/RegionFormSchema.ts
DavidCockerill: Good catch — this one was reachable straight from the form. Fixed in 9c114b35: the schema now rejects `forceLocations: true` when both preferred-location lists are empty, with the message on the Force field itself, plus three schema tests.

One deliberate difference from your suggestion, and I'd like your read on it. I enforced **at least one location overall** rather than one per applicable provider. The form has no provider or channel data to know which providers can actually provision into a given region, and requiring both lists would block legitimately single-provider regions (several seeded regions carry Linode locations only). So this guard catches the unambiguous case — Force with an empty plan, which always fails — and leaves the per-provider invariant to the backend, which is the only side that knows the applicable providers.

That backend invariant isn't implemented yet. If you agree it's worth having, I'll file it against central-manager rather than widening this PR.

— Claude Opus 5

src/features/admin/regions/queries/getOrganizations.ts
DavidCockerill: Agreed that the cap was silent data loss rather than a perf tradeoff — thanks for pushing on it. Fixed in 7b41d593, though not quite the way you suggested, and I want to explain why.

Server-side name search alone wouldn't have been enough: this same query also feeds the list page's `orgNameById` map, which needs **every** id→name pair for the Scope column's display names and for the row filter to match regions by organization name. A search-as-you-type picker leaves both of those broken and still requires enumerating organizations separately.

So `getOrganizations` now pages until the server returns a short page (Harper collections carry no total count, so that's the only end-of-table signal), with a hard 40-page bound. Hitting that bound surfaces as a `truncated` flag and a visible warning on the picker rather than quietly dropping rows — that's the part of the old behaviour I think you were most right about.

For the DOM cost: `MultiSelect` gained `maxVisibleOptions` (100 for the org picker). Only that many matches render; the overflow shows as "N more — keep typing to narrow" and stays reachable through the filter. That's a poor substitute for real virtualization, and if this list ever gets big enough to matter, virtualizing is still the right answer — I'd rather do it as its own change than bolt it on here.

Covered by 5 new tests in `getOrganizations.test.ts` (short-page stop, paging past a full page with non-overlapping ranges, the truncation bound) and 2 more in `MultiSelect.test.tsx`.

— Claude Opus 5

src/features/admin/regions/components/MultiSelect.tsx
DavidCockerill: Confirmed and fixed in 9c114b35 — the component really was pointer-only, and that's on me: I added the blanket `stopPropagation` to stop Radix's typeahead eating the filter text, and over-suppressed.

Worth noting for anyone reading this later: **suppressing only the printable keys wasn't sufficient by itself.** Radix's first/last focus logic bails unless `event.target === content`, so with focus in the filter input, ArrowDown reaches Radix and is then ignored. The input therefore handles ArrowDown/ArrowUp itself and moves focus to the first/last enabled option, after which Radix's roving focus takes over the remaining arrows and Enter/Space. Printable keys — and only those — are still stopped so typeahead can't steal the filter.

Tests added along the lines you outlined: open, filter, arrow into the list, select with Enter, then Escape out and assert the chip. Plus ArrowUp-to-last, a capped-list variant, and one asserting a printable key leaves focus in the input. I mutation-checked them — restoring the blanket `stopPropagation` turns two of them red.

Tab is still trapped inside the menu; that's Radix's menu semantics (`Tab` is preventDefaulted in the content handler), so Escape is the way out. A real combobox/listbox would fix that properly, but it means introducing a pattern this repo doesn't have yet, so I left it out of scope here.

— Claude Opus 5

src/features/admin/routes.ts
DavidCockerill: Confirmed — this was a regression I introduced, and it broke a deep link that had already shipped. Fixed in 9c114b35 exactly as you suggested: `/admin/notifications` is canonical again and `/admin` redirects to it, so the requested Regions route is untouched and the bookmark keeps working.

The rail's first item now points at `/admin/notifications` (no more `exact` special case). New coverage in `src/features/admin/__tests__/adminRoutes.test.ts`: each section resolves to its own route, and one test actually loads the router from `/admin` and asserts it lands on `/admin/notifications` — route matching alone wouldn't catch it, since the redirect lives in `beforeLoad`. Putting Notifications back on the index path turns that test red.

— Claude Opus 5

DavidCockerill:
@kriszyp thanks for the second pass — that was a genuinely valuable review. Five of the seven are fixed, one I'm pushing back on, and one has been overtaken by events. Replies are inline on each thread; leaving them all unresolved for you to close.

**Fixed** (`9c114b35`, `7b41d593`):

| Finding | What changed |
|---|---|
| `[]` is public, not hidden | `RegionScope` renders Public for `null` and `[]`; `organizationScope.ts` deleted |
| `/admin/notifications` removed | Canonical again, `/admin` redirects to it |
| `MultiSelect` keyboard trap | Arrows move focus into the list; only printable keys suppressed |
| Force with no locations | Schema rejects it, message on the Force field |
| Org picker capped at 500 | Pages the whole table; bound surfaces as a visible warning, not silent truncation |

Every fix has regression coverage, and I mutation-checked the two subtle ones — reverting the keyboard fix or the route fix turns the new tests red.

**Pushing back on one:** the `/Admin/Region/` visibility-filter finding. `RegionAdmin extends tables.Region` (`RegionAdmin.js:5` destructures the base table), which makes it a *sibling* of the customer `Region` resource rather than a subclass — so `super.get()` never reaches `isRegionVisibleToOrg`. Details on that thread; correct me if you're seeing a different path.

**Overtaken:** the "can merge independently" note was accurate when written, but #518 merged as `a06faeb7` earlier today, so `main` now has `forceLocations` and `active`. The deploy-ordering point still stands and I've flagged it.

Two things I'd flag honestly rather than let pass:

- The `forceLocations` guard enforces *at least one location overall*, not one per applicable provider — the form has no provider/channel data, and requiring both lists would block single-provider regions. The per-provider invariant belongs in the backend and isn't implemented; happy to file it.
- The central-manager region **integration tests have never been executed** — they need a live CM, and CM's CI runs `unit` only. The 11 unit validator tests do run and pass.

— Claude Opus 5

src/features/admin/regions/RegionFormSchema.ts
```diff
@@ -0,0 +1,44 @@
...
+	// false = retired: hidden from the customer provisioning list. New regions default active.
+	active: z.boolean(),
+	// Empty ⇒ region is available to all organizations (shown as "Public").
+	organizationIds: z.array(z.string()),
+}).superRefine((values, ctx) => {
+	// With forceLocations the preferred list IS the placement plan rather than a hint, so an empty one
+	// leaves addDedicatedHosts nothing to pick and provisioning fails with "No valid locations found".
+	if (values.forceLocations && !values.linodePreferredLocations.length && !values.gcpPreferredLocations.length) {
```
kriszyp, line 35: This closes only the both-arrays-empty case. It still accepts Force with Linode locations but no GCP locations (and vice versa); `RegionFormSchema.test.ts:50-57` now explicitly locks that behavior in. For an organization provisioned through the missing provider, Force makes that provider's empty array the complete placement plan and provisioning still fails with `No valid locations found`.

Please require a non-empty list for every provider that can serve the selected scope. If the form cannot derive provider/channel from the selected organizations, the safe rule is to require both lists. This invariant also needs to live in central-manager—the merged #518 validates that locations exist, but does not reject an empty forced provider plan—so another client cannot persist the same broken configuration.

src/features/admin/regions/components/RegionFormModal.tsx
```diff
@@ -0,0 +1,395 @@
...
```
kriszyp, line 131: Companion central-manager #518 has now merged, which addresses the server contract, but landing is not deployment: Studio still submits `active` and `forceLocations` unconditionally, and an older CM accepts the request, strips both fields, and lets this `onSuccess` path report that the edit worked. The current PR body still says this PR can merge independently and describes that silent discard as expected.

Please make the CM deployment an explicit prerequisite for deploying this screen (and update the stale PR body), or capability-gate these controls until the API can persist them. That keeps a successful toast from claiming an inactive/forced configuration was saved when it was not.

kriszyp (approved):
Main things have been covered, still a few things to maybe check on.

src/features/admin/regions/RegionFormSchema.ts
DavidCockerill: Thanks for pinning down the mechanism — you're right, and I want to correct my earlier reply. I defended the at-least-one rule by pointing at single-provider regions, but that only holds when Force is **off**, where an empty list falls through to every location in the region. With Force on there's no fallback, and since `addDedicatedHosts.js:30` makes the provider purely a function of `organizationData.channel`, the region has no say in which list gets read. A public Force region with one empty list really is broken for every org on the other channel. My reasoning applied the Force-off intuition to the Force-on case.

David's call is to accept that and let it fail at provision time. The reasoning: it takes a super-user deliberately enabling Force and then leaving one list empty on a region reachable by both channels, and the outcome is a loud `No valid locations found` rather than silent data loss or a wrong customer-visible state. To be straight with you, the deploy is customer-initiated, so a customer is who'd see it fail — he's weighed that against the likelihood and judged the form complexity (deriving channel from the selected orgs, recomputing as scope changes) not worth it.

So we're deliberately doing neither half — not the per-provider form rule, and not the central-manager invariant. Flagging that plainly rather than leaving you expecting a follow-up issue. The both-arrays-empty guard from `9c114b35` stays, since that case is unambiguous and free to catch.

Worth revisiting if it ever bites in practice. And if you think the loud-failure argument doesn't hold — say, the error surfaces somewhere worse than a failed deploy — push back and I'll take it back to David.

— Claude Opus 5

src/features/admin/regions/components/RegionFormModal.tsx
DavidCockerill: Agreed, and done — the PR body is updated. The paragraph claiming this can merge independently and describing the `stripUnknown` discard as expected is replaced with an explicit prerequisite: deploying this screen requires the central-manager build carrying #518, because against an older CM the request succeeds, both fields are dropped, and this form still reports the edit as saved for a region that stays customer-visible.

We're going with the prerequisite rather than capability-gating the controls. There's no capability probe in the API today, so gating would mean inventing one (or version-sniffing) for a window that closes as soon as CM ships. If deploy ordering turns out to be hard to guarantee in practice, gating is the right fallback and I'd rather add it then than carry it speculatively.

Thanks for catching that the body was still stating the old position — that was stale text I should not have left standing.

— Claude Opus 5

PR #1572, opened 2026-07-28
test(e2e): make skips honest — fail on regressions that used to go quiet

dawsontoth:
Follow-up to #1553, from the adversarial review of the e2e system. **A monitor that skips is a
monitor that lies** — these specs had three skips that fired on conditions at least as likely to be
real regressions, which would have turned them permanently green at exactly the wrong moment.

## The skips

**`org-users` — unresolvable organization.** `resolveOrgId` returns `null` for *any* failure to
observe an org: a broken single-org redirect, an org picker that stopped rendering, a failing orgs
query, a slow backend. Every one of those is a regression this spec exists to catch, and each
produced a skip → green run. The account *is* configured (the describe already skips when it
isn't), so an unresolvable org is now a **failure** with a message that says what it usually means.
`E2E_ALLOW_NO_ORG=1` opts out for an account that genuinely belongs to no org.

**`signup-verification` — 403 on signup.** Any 403 was treated as "the target's
`ALLOWLIST_EMAIL_DOMAINS` doesn't cover our Mailosaur domain". It's at least as likely to be an
authz change, a broken `allowCreate`, or a WAF/rate-limit after repeated automated runs — and the
condition is *stable*, so the round-trip would go silent the moment signup broke. Now it **fails**,
including the server's response text so the cause is obvious. `E2E_ALLOW_SIGNUP_403_SKIP=1` opts
out once you've confirmed it really is the domain gate.

Neither variable is set by the automated lanes, which additionally fail a run if *any* spec skips.

## Also in here

- **`testData.ts` used `??`** — the documented config ships `PLAYWRIGHT_TEST_EMAIL=` (blank) so the
  round-trip auto-generates an address, but `??` keeps `''`, so `verifying.anon` navigated to
  `?email=` and silently exercised the *no-email* path instead of the unverified-email one. `||`.
- **The `setup` project disables trace/video.** Playwright records `fill()` arguments, and that
  project types the real test-account password — a failed login persisted it in a trace zip that
  redaction can't scrub.
- **Round-trip timeout raised 120s → 210s.** It was smaller than the sum of its own waits (90s mail
  poll + three 30s navigations + interactions), so a slow-but-working delivery read as a failure.
- **README corrections**: the trusted lane is sandboxed (it can no longer "run on host"), the
  account-churn warning predates the spec's self-cleanup, and the layout block was stale.

## Testing

`tsc --noEmit` clean; anon specs pass against dev (the round-trip skips only where it legitimately
should — no Mailosaur configured on that machine). The full 8-spec suite is green on the runner
against `stage` through the sandboxed lane; these changes only alter *skip* paths, not assertions.

### DIFF
--- a/e2e/README.md
+++ b/e2e/README.md
@@ -158,12 +158,25 @@ Next, here in the specs:
 
 ## Skips are deliberate, and narrow
 
-A monitor that skips is a monitor that lies. These specs fail rather than skip when a prerequisite
-they were _given_ stops working:
+A monitor that skips is a monitor that lies. These specs **fail rather than skip whenever a
+prerequisite they were _given_ stops working** — the disposition is decided here, in the specs, so
+it does not depend on what any runner happens to check:
 
 - **Unresolvable organization** → failure (a broken post-login redirect or org picker looks exactly
   like "no org"). Opt out with `E2E_ALLOW_NO_ORG=1` for an account that genuinely has none.
 - **Signup 403** → failure (as likely an authz/WAF regression as the email-domain gate). Opt out
   with `E2E_ALLOW_SIGNUP_403_SKIP=1` once you've confirmed it's `ALLOWLIST_EMAIL_DOMAINS`.
+- **Org-users 403** → failure (as likely an authz regression as a provisioning gap). Opt out with
+  `E2E_ALLOW_ORG_USERS_403_SKIP=1` once you've confirmed the account simply lacks users-view.
 
-Neither variable is set by the automated lanes, which additionally fail the run if _any_ spec skips.
+None of those variables is set by the automated lanes.
+
+What remains a skip is only the case where the environment genuinely cannot run a spec at all — no
+test account (`PLAYWRIGHT_USER_*`) or no Mailosaur config. Those are absence-of-configuration, not
+ambiguity, and they are visible as `skipped` in the report.
+
+> Defense in depth, not the guarantee: the harness's trusted lane also fails a run when any spec
+> skips (a rotated credential would otherwise leave the authed specs quiet). That check lives in
+> [studio-e2e-harness#2](https://github.com/HarperFast/studio-e2e-harness/pull/2) and is **not on
+> that repo's `main` yet** — so treat it as a second layer that is still landing, and rely on the
+> per-spec dispositions above.
--- a/e2e/README.md
+++ b/e2e/README.md
@@ -135,7 +135,9 @@ Two operational gotchas:
 - **Signup is domain-allowlisted.** central-manager `User.allowCreate` rejects signup (403)
   unless the email domain matches the `ALLOWLIST_EMAIL_DOMAINS` Configuration record. The
   Mailosaur server domain (or a Mailosaur custom domain under an already-allowlisted domain)
-  must be in that list, or the test skips itself with a pointer.
+  must be in that list; otherwise the round-trip **fails** with a pointer to this, unless the
+  explicit escape hatch `E2E_ALLOW_SIGNUP_403_SKIP=1` is set (a 403 is at least as likely to be an
+  authz/WAF regression as a misconfigured allowlist, so failing is the safe default).
 - **Account churn.** Each run creates a real account on the target env; the spec self-deletes it in
   a `finally` (see `deleteThrowawayAccount`). A failure before login has no session to delete, so
   the odd account can still leak — worth an occasional sweep.
--- a/e2e/README.md
+++ b/e2e/README.md
@@ -16,7 +16,7 @@ context that holds real credentials, email, and network egress.**
 | Target    | Deployed dev app (`PLAYWRIGHT_BASE_URL`)    | A sandboxed build of the PR            |
 | LLM role  | Triages failures only                       | Reads diff, adapts/adds tests          |
 | Creds     | Disposable test account                     | Burnable throwaway; **no live email**  |
-| Isolation | Can run on host                             | Container + egress allowlist           |
+| Isolation | Sandboxed (container + egress allowlist)    | Container + egress allowlist           |
 | Output    | Failure → issue/PR you review               | Proposed test diffs → PR **you** merge |
 
 The **control plane** that runs these lanes autonomously — the trusted-lane scheduler
@@ -98,11 +98,14 @@ docker compose run --rm e2e pnpm update-snapshots   # regenerate, then review th
 ```
 e2e/
   playwright.config.ts   projects: setup (login) → authed; anon (no session)
-  auth.setup.ts          logs in via UI once, saves cookie storageState
   tests/
+    auth.setup.ts            logs in via UI once, saves cookie storageState
     sign-in.anon.spec.ts     sign-in page render + validation + visual baseline
-    verifying.anon.spec.ts   email-verification screens (no inbox needed yet)
+    verifying.anon.spec.ts   email-verification screens
+    signup-verification.anon.spec.ts  full signup → email → verify → login round-trip
     org-users.authed.spec.ts app shell + org users list (needs a test account)
+    mail.ts                  Mailosaur wrapper (controlled inbox)
+    testData.ts              shared fixtures
   Dockerfile / docker-compose.yml / .dockerignore
   .env.e2e.example         copy to .env.e2e (gitignored)
 ```
@@ -133,8 +136,9 @@ Two operational gotchas:
   unless the email domain matches the `ALLOWLIST_EMAIL_DOMAINS` Configuration record. The
   Mailosaur server domain (or a Mailosaur custom domain under an already-allowlisted domain)
   must be in that list, or the test skips itself with a pointer.
-- **Account churn.** Each run creates a real account on the target env. Before wiring into a
-  daily loop, add a cleanup story (backend test-account purge) or a dedicated throwaway tenant.
+- **Account churn.** Each run creates a real account on the target env; the spec self-deletes it in
+  a `finally` (see `deleteThrowawayAccount`). A failure before login has no session to delete, so
+  the odd account can still leak — worth an occasional sweep.
 
 ## Roadmap
 
@@ -149,3 +153,15 @@ Next, here in the specs:
 
 1. **Resend-invite** — seed a `PENDING` org user fixture, assert the detail-modal flow.
 2. **More coverage** — broaden beyond auth/org-users as the flows the lanes protect grow.
+
+## Skips are deliberate, and narrow
+
+A monitor that skips is a monitor that lies. These specs fail rather than skip when a prerequisite
+they were _given_ stops working:
+
+- **Unresolvable organization** → failure (a broken post-login redirect or org picker looks exactly
+  like "no org"). Opt out with `E2E_ALLOW_NO_ORG=1` for an account that genuinely has none.
+- **Signup 403** → failure (as likely an authz/WAF regression as the email-domain gate). Opt out
+  with `E2E_ALLOW_SIGNUP_403_SKIP=1` once you've confirmed it's `ALLOWLIST_EMAIL_DOMAINS`.
+
+Neither variable is set by the automated lanes, which additionally fail the run if _any_ spec skips.
--- a/e2e/playwright.config.ts
+++ b/e2e/playwright.config.ts
@@ -69,7 +69,10 @@ export default defineConfig({
 		{
 			name: 'setup',
 			testMatch: /auth\.setup\.ts$/,
-			use: { ...devices['Desktop Chrome'] },
+			// Traces and video record `fill()` arguments, and this project types the real test-account
+			// password — a failed login would persist it in a trace zip that redaction cannot scrub.
+			// Its failure message is self-explanatory without them.
+			use: { ...devices['Desktop Chrome'], trace: 'off', video: 'off' },
 		},
 		// 2. Flows that must start UNAUTHENTICATED (sign-in page, validation errors,
 		//    unverified-login -> verification). No stored session.
--- a/e2e/tests/org-users.authed.spec.ts
+++ b/e2e/tests/org-users.authed.spec.ts
@@ -4,11 +4,16 @@ import { expect, type Page, test } from '@playwright/test';
  * Authenticated lane — the app shell after login and the org-users list.
  * Uses the stored session from auth.setup.ts (the `authed` project).
  *
- * SKIP CONTRACT: skips only for things this environment genuinely does not have — no test account
- * configured, or a real 403 on the org-users fetch (the account lacks users-view). Everything else
- * FAILS, including an org that cannot be resolved: that is far more often a regression in the
- * post-login redirect / org picker / orgs query than an account with no org. Deliberate opt-outs
- * (`E2E_ALLOW_NO_ORG=1`) exist for local dev and are never set by the automated lanes.
+ * SKIP CONTRACT: the ONLY unconditional skip is "this environment has no test account configured"
+ * (`!hasCreds`) — an absence of configuration, not an ambiguous result. Everything else FAILS,
+ * because each alternative is at least as likely to be the regression this spec exists to catch:
+ *   - org cannot be resolved  → a broken post-login redirect / org picker / orgs query looks
+ *                               identical to an account with no org;
+ *   - 403 on the org-users fetch → an authz regression looks identical to a provisioning gap;
+ *   - error boundary with no 403 → the page itself broke; NOT skippable under any flag.
+ * Two opt-outs exist for local dev — `E2E_ALLOW_NO_ORG=1` and `E2E_ALLOW_ORG_USERS_403_SKIP=1` —
+ * each scoped to its own condition so it cannot silence a neighbouring failure. Neither is set by
+ * the automated lanes. If you change a disposition here, change this block in the same commit.
  */
 const hasCreds = Boolean(process.env.PLAYWRIGHT_USER_EMAIL && process.env.PLAYWRIGHT_USER_PASSWORD);
 
@@ -85,8 +90,12 @@ test.describe('authenticated app', () => {
 			// authz regression on exactly the page this spec covers — and the condition is stable, so
 			// skipping makes the spec go permanently quiet the moment permissions break. Fail by
 			// default; opt out explicitly once you've confirmed it's provisioning.
+			// Scoped to `saw403`: without that guard the hatch would also silence the !saw403
+			// disposition below — "the page itself broke" — which must never be skippable, and the
+			// skip reason would assert a 403 that did not happen. (`.env.e2e` is dotenv-loaded, so a
+			// var set once for a provisioning gap stays set.)
 			test.skip(
-				process.env.E2E_ALLOW_ORG_USERS_403_SKIP === '1',
+				saw403 && process.env.E2E_ALLOW_ORG_USERS_403_SKIP === '1',
 				`Org-users returned 403 for ${orgId} and E2E_ALLOW_ORG_USERS_403_SKIP=1 — treating as a `
 					+ 'known provisioning gap for this account.',
 			);
--- a/e2e/tests/org-users.authed.spec.ts
+++ b/e2e/tests/org-users.authed.spec.ts
@@ -38,11 +38,24 @@ test.describe('authenticated app', () => {
 	});
 
 	test('org users list renders with the expected columns', async ({ page }) => {
+		// An unresolvable org used to skip. But resolveOrgId returns null for ANY failure to observe
+		// one — a broken single-org redirect, an org picker that stopped rendering, a failing orgs
+		// query — i.e. exactly the regressions this spec exists to catch, silently turned green. The
+		// account IS configured (the describe skips otherwise), so treat it as a failure. Local dev
+		// against an account that genuinely has no org can opt out.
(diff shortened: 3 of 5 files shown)


e2e/tests/org-users.authed.spec.ts
```diff
@@ -38,11 +38,24 @@ test.describe('authenticated app', () => {
...
 		const orgId = await resolveOrgId(page);
-		test.skip(
-			!orgId,
-			'No accessible organization for this account — add it to an org, or set PLAYWRIGHT_ORG_ID.',
-		);
+		if (!orgId) {
+			test.skip(
+				!!process.env.E2E_ALLOW_NO_ORG,
+				'No organization resolved and E2E_ALLOW_NO_ORG is set — treating this account as org-less.',
+			);
```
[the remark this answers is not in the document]
dawsontoth: Agreed and fixed in a7f935f9 — now `process.env.E2E_ALLOW_NO_ORG === '1'`. Worth noting this is more than style here: the var is a safety opt-out that re-enables the skip this PR removes, so a stray `=0` would have silently restored the exact failure mode being fixed. Strict equality fails closed.

e2e/tests/signup-verification.anon.spec.ts
```diff
@@ -54,15 +56,25 @@ test.describe('signup → email verification → login', () => {
...
+			// spec used to — makes the round-trip go permanently quiet the moment signup breaks, which
+			// is precisely when a monitor should shout. So: FAIL by default, and require an explicit
+			// opt-in for the known-misconfigured case (the monitored lanes never set it).
+			if (signupResponse.status() === 403) {
+				const detail = await signupResponse.text().catch(() => '');
+				test.skip(
+					!!process.env.E2E_ALLOW_SIGNUP_403_SKIP,
+					'Signup returned 403 and E2E_ALLOW_SIGNUP_403_SKIP is set — treating as the known '
+						+ `ALLOWLIST_EMAIL_DOMAINS gap. Server said: ${detail.slice(0, 200)}`,
+				);
```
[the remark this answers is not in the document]
dawsontoth: Agreed and fixed in a7f935f9 — now `process.env.E2E_ALLOW_SIGNUP_403_SKIP === '1'`, same reasoning: an accidental `=0` would have re-enabled the skip and let a signup regression go quiet again.

e2e/README.md
```diff
@@ -133,8 +136,9 @@ Two operational gotchas:
   unless the email domain matches the `ALLOWLIST_EMAIL_DOMAINS` Configuration record. The
   Mailosaur server domain (or a Mailosaur custom domain under an already-allowlisted domain)
   must be in that list, or the test skips itself with a pointer.
```
kriszyp: Could you update this sentence to match the new fail-by-default contract? With this PR, an allowlist 403 fails unless `E2E_ALLOW_SIGNUP_403_SKIP=1`; saying the test "skips itself" tells an operator to expect the opposite result and can make a trusted-lane failure surprising. For example: "must be in that list; otherwise the test fails with a pointer unless the explicit skip escape hatch is enabled."

e2e/tests/org-users.authed.spec.ts
```diff
@@ -38,11 +38,24 @@ test.describe('authenticated app', () => {
 	});
 
 	test('org users list renders with the expected columns', async ({ page }) => {
+		// An unresolvable org used to skip. But resolveOrgId returns null for ANY failure to observe
+		// one — a broken single-org redirect, an org picker that stopped rendering, a failing orgs
+		// query — i.e. exactly the regressions this spec exists to catch, silently turned green. The
+		// account IS configured (the describe skips otherwise), so treat it as a failure. Local dev
+		// against an account that genuinely has no org can opt out.
 		const orgId = await resolveOrgId(page);
```
kriszyp, line 53: The behavior change here is right. Could you also update the file-level comment at lines 7–9 and `resolveOrgId`'s comments at lines 111–114 and 137–139? They still say a null result means the account is truly org-less and the spec skips, while this hunk correctly recognizes that null can also represent a redirect, picker, or query regression and now fails by default. Keeping that old assumption in the helper documentation makes it easy to reintroduce this false green later.

kriszyp (approved):
Sounds good

e2e/README.md
dawsontoth: Fixed in 95db6b95 — the note now reads that the round-trip **fails** with a pointer unless `E2E_ALLOW_SIGNUP_403_SKIP=1` is set, and says why failing is the safe default (a 403 is as likely an authz/WAF regression as a misconfigured allowlist). You're right that leaving it as 'skips itself' would have made a trusted-lane failure look like a bug in the harness.

e2e/tests/org-users.authed.spec.ts
dawsontoth: Fixed in 95db6b95 — updated all three: the file-level skip contract, `resolveOrgId`'s doc comment, and the inner poll comment. The helper doc now states that `null` is ambiguous (a broken redirect / picker / orgs query is indistinguishable from a genuinely org-less account), that it only reports 'not observed', and that the caller fails by default — with an explicit 'do not reintroduce a skip here'. Agreed that leaving the old assumption in the helper was the easiest path back to the false green.

kriszyp (approved):
Still approve.

e2e/README.md
```diff
@@ -149,3 +155,15 @@ Next, here in the specs:
...
+
+## Skips are deliberate, and narrow
+
+A monitor that skips is a monitor that lies. These specs fail rather than skip when a prerequisite
+they were _given_ stops working:
+
+- **Unresolvable organization** → failure (a broken post-login redirect or org picker looks exactly
+  like "no org"). Opt out with `E2E_ALLOW_NO_ORG=1` for an account that genuinely has none.
+- **Signup 403** → failure (as likely an authz/WAF regression as the email-domain gate). Opt out
+  with `E2E_ALLOW_SIGNUP_403_SKIP=1` once you've confirmed it's `ALLOWLIST_EMAIL_DOMAINS`.
```
DavidCockerill, line 168: This guarantee isn't implemented. I cloned `HarperFast/studio-e2e-harness` and read both gates: `host/run-pr-lane.sh:222-225` decides GREEN from `jq '.stats.unexpected // 0'` alone, and `trusted/run-trusted.sh:45-48` just propagates Playwright's exit code — which is 0 when tests skip. `stats.skipped` is never read anywhere in the harness, and `agent/triage.md:20` treats a non-failure as a pass. **An all-skipped run is green in both lanes today.**

That matters because this is the stated justification for leaving three skips in place: `!hasCreds` (line 30), `!mailConfigured` (signup spec line 25), and the `saw403` skip at line 82. Rotate the test account's password and the whole authed lane goes quiet — and green.

Fix either side: read `.stats.skipped` in the pr-lane gate plus a post-check in `run-trusted.sh`, or soften this to "should" / a TODO with a link. The first clause (neither var is set by the lanes) I verified and it's true.

e2e/tests/org-users.authed.spec.ts
```diff
@@ -38,11 +40,24 @@ test.describe('authenticated app', () => {
...
+		if (!orgId) {
+			test.skip(
+				process.env.E2E_ALLOW_NO_ORG === '1',
+				'No organization resolved and E2E_ALLOW_NO_ORG=1 — treating this account as org-less.',
+			);
+			throw new Error(
+				'Could not resolve an organization for this account within 15s: no single-org redirect and '
+					+ 'no org link on the picker. That is usually a regression in post-login landing or the org '
+					+ 'picker — not a missing org. Pin PLAYWRIGHT_ORG_ID, or set E2E_ALLOW_NO_ORG=1 if this '
+					+ 'account genuinely belongs to no organization.',
```
DavidCockerill, line 63: The third skip, ~24 lines below at line 82, has the identical ambiguity you just removed from the signup 403:

```js
test.skip(saw403, `This account can't view org users for ${orgId} …`);
```

Your own argument applies verbatim — a 403 on `GET /OrganizationRole/` is at least as likely to be an authz regression as a provisioning gap, and the condition is *stable*, so the spec goes permanently quiet the moment permissions break on exactly the page it exists to cover.

The `/OrganizationRole/`-scoping above is careful work and I'd keep all of it; only the disposition should change: fail by default, `E2E_ALLOW_ORG_USERS_403_SKIP=1` to opt out, matching the two you fixed. Doing two of three leaves the file self-inconsistent.

DavidCockerill (commented):
The skip→fail conversions are correctly built, and the trace/video scoping and `??`→`||` changes are real bugs found along the way. Good direction — a silently-skipping test is coverage that reports success while exercising nothing.

One thing to fix before merge, and it's an odd one: **the README documents a lane guarantee that doesn't exist.** I cloned `studio-e2e-harness` and read both gates — the PR lane decides green from `stats.unexpected` alone, and the trusted lane just propagates Playwright's exit code, which is 0 when everything skips. `stats.skipped` is never read anywhere. So an all-skipped run is green today, in both lanes.

That's load-bearing here, because it's the stated reason three skips are safe to leave in. Concretely: rotate the test account's password and the entire authed lane goes quiet and green — the exact failure mode this PR exists to close, one level up.

Second, smaller: the third `saw403` skip has the same ambiguity you just removed from the signup 403, so the file is currently self-inconsistent — two of three converted.

Two non-blocking notes on load-sensitivity and the retry/timeout interaction in the remaining threads.

e2e/tests/org-users.authed.spec.ts
```diff
@@ -38,11 +40,24 @@ test.describe('authenticated app', () => {
 	});
 
 	test('org users list renders with the expected columns', async ({ page }) => {
+		// An unresolvable org used to skip. But resolveOrgId returns null for ANY failure to observe
+		// one — a broken single-org redirect, an org picker that stopped rendering, a failing orgs
+		// query — i.e. exactly the regressions this spec exists to catch, silently turned green. The
+		// account IS configured (the describe skips otherwise), so treat it as a failure. Local dev
```
DavidCockerill, line 51: Non-blocking: this failure is now load-sensitive. `resolveOrgId`'s own comment says the 15s poll exists because "under parallel load the org-card link can appear after `networkidle`, which otherwise produced a flaky 'no org' skip." That flake is now a red run — intended, but worth one line in the comment saying the remedy is to raise the 15s window, not restore the skip. Same reasoning you applied to the 210s round-trip budget.

e2e/playwright.config.ts
```diff
@@ -69,7 +69,10 @@ export default defineConfig({
 		{
 			name: 'setup',
 			testMatch: /auth\.setup\.ts$/,
-			use: { ...devices['Desktop Chrome'] },
+			// Traces and video record `fill()` arguments, and this project types the real test-account
```
DavidCockerill, line 72: Good catch on the trace/video — correctly scoped, and the throwaway password in the round-trip spec is `randomUUID()`-derived per run, so keeping traces there was the right call.

Unrelated and non-blocking: `test.setTimeout(210_000)` now interacts with `retries: 2` under `CI=1`. Worst case the `@roundtrip` spec occupies ~10.5 min of a `workers: 2` run and creates three real accounts. A `retries: 1` override on that test may be worth it if daily wall clock starts to matter.

e2e/README.md
dawsontoth: You're right, and thank you for actually cloning and reading it. I verified: `stats.skipped` is absent from harness `main` — my coverage floor is on the unmerged studio-e2e-harness#2, so the guarantee an operator would find today is the opposite of what the README claimed. Fixed in 0f261d4a by removing the dependency rather than restating it: the disposition now lives in the specs (which is why the third skip had to go too), and the harness floor is described as a second layer that is still landing, with a link. A cross-repo guarantee shouldn't have been load-bearing for leaving skips in.

e2e/tests/org-users.authed.spec.ts
dawsontoth: Agreed — fixed in 0f261d4a. Same disposition as the other two: fail by default, `E2E_ALLOW_ORG_USERS_403_SKIP=1` to opt out. The `/OrganizationRole/`-scoping is untouched; only the verdict changed, and the two causes now get separate diagnoses (403 → provisioning-or-authz-regression; error boundary with no 403 → the page itself broke). You were right that two-of-three left the file arguing with itself.

e2e/tests/org-users.authed.spec.ts
dawsontoth: Added in 0f261d4a — the comment now says the failure is load-sensitive and that the remedy is to RAISE the 15s window, never to restore the skip, since that would trade a visible flake back for a silent false green.

e2e/playwright.config.ts
dawsontoth: Good catch — capped the round-trip at `retries: 1` in 0f261d4a via `test.describe.configure`. Worst case drops from ~10.5min/3 real signups to ~7min/2, which also halves the throwaway-account churn on dev.

e2e/tests/org-users.authed.spec.ts
```diff
@@ -65,17 +80,26 @@ test.describe('authenticated app', () => {
 		await expect(table.or(errorBoundary).first()).toBeVisible();
 
 		if (await errorBoundary.isVisible().catch(() => false)) {
-			// A 403 → this account legitimately lacks users-view on this env: skip, don't fail red.
+			// A 403 on the org-users fetch is AMBIGUOUS in the same way the signup 403 is: it can mean
+			// this account was never provisioned with users-view, but it is at least as likely to be an
+			// authz regression on exactly the page this spec covers — and the condition is stable, so
+			// skipping makes the spec go permanently quiet the moment permissions break. Fail by
+			// default; opt out explicitly once you've confirmed it's provisioning.
 			test.skip(
```
DavidCockerill, line 97: Non-blocking, but worth a one-line fix: the new hatch isn't scoped to the case it documents.

`test.skip(process.env.E2E_ALLOW_ORG_USERS_403_SKIP === '1', …)` sits inside the *error-boundary* branch, not the *403* branch — so with the var set it also silences the `!saw403` disposition, the one whose own message says "the page itself broke … a genuine regression, not a permission gap." At `a7f935f9` that case was an unconditional `expect(saw403).toBe(true)` and couldn't be silenced, so this is a small step back from where the PR started.

Two things sharpen it: the skip reason asserts a 403 that didn't happen, so the report misleads; and `playwright.config.ts:10` dotenv-loads `e2e/.env.e2e`, so a dev who sets this once for a provisioning gap keeps it set.

```diff
-		process.env.E2E_ALLOW_ORG_USERS_403_SKIP === '1',
+		saw403 && process.env.E2E_ALLOW_ORG_USERS_403_SKIP === '1',
```

Matches how the other two are already scoped — `E2E_ALLOW_NO_ORG` only inside `if (!orgId)`, `E2E_ALLOW_SIGNUP_403_SKIP` only inside `if (status() === 403)`.

e2e/tests/org-users.authed.spec.ts
```diff
@@ -4,9 +4,11 @@ import { expect, type Page, test } from '@playwright/test';
  * Authenticated lane — the app shell after login and the org-users list.
  * Uses the stored session from auth.setup.ts (the `authed` project).
  *
- * Skips cleanly when no test account is configured, and when the account has no
- * accessible org / can't view org users on the target environment — so a data
- * difference between environments is a skip, not a red failure.
+ * SKIP CONTRACT: skips only for things this environment genuinely does not have — no test account
```
DavidCockerill: This SKIP CONTRACT went stale one commit after you wrote it: `95db6b95` added it (answering @kriszyp's ask), then `0f261d4a` changed the behaviour it describes. It still says a "real 403 on the org-users fetch" skips, but that now fails by default 80 lines down. It also lists only `E2E_ALLOW_NO_ORG=1` while this file has two opt-outs.

Same failure mode @kriszyp named — a stale assumption in the doc is the easiest path back to a false green.

DavidCockerill (approved):
Approving at `0f261d4`. All four of my findings are handled, and I re-verified the one that mattered rather than taking the commit message for it.

On the README blocker: you resolved it by **deleting the cross-repo claim** rather than restating it more softly, and moved disposition into the specs where it's actually enforced. I re-read `studio-e2e-harness` at its current head to confirm — neither gate reads `.stats.skipped`, so the guarantee still doesn't exist, and the README no longer says it does. The residual pointer to `studio-e2e-harness#2` also checks out against that PR's diff. That's the honest version of this fix, and it's better than making the doc vaguer.

Third `saw403` skip converted, the `resolveOrgId` load-sensitivity note and the retries×timeout interaction both addressed. Three config-absence skips remain and are now documented rather than implicit — that's the right resting place until harness#2 lands.

Two non-blocking threads, one of which I'd genuinely like taken: the new `E2E_ALLOW_ORG_USERS_403_SKIP` hatch is scoped to the error-boundary branch rather than the 403 branch, so it also silences the page-crash case — which was unsilenceable before this PR. One-line diff in the thread.

Nice piece of work overall. A silently-skipping test is coverage that reports success while exercising nothing, and this closes most of that surface.

e2e/tests/org-users.authed.spec.ts
dawsontoth: Applied your diff in 305de334. You're right that this was a step back from where the PR started — at `a7f935f9` the no-403 case was an unconditional `expect(saw403).toBe(true)`, and my restructuring made it silenceable by a flag documented for something else. Verified with a probe: with `E2E_ALLOW_ORG_USERS_403_SKIP=1`, the 403 cases skip and `no403` still fails. Good point about `.env.e2e` being dotenv-loaded — set-once-stays-set is exactly why the hatch needed narrowing rather than trusting the operator to unset it.

e2e/tests/org-users.authed.spec.ts
dawsontoth: Fixed in 305de334 — and worth naming what happened: I added that block in `95db6b95` to answer @kriszyp's stale-comment finding, then invalidated it one commit later in `0f261d4a` with the change he asked for. Same failure mode, one commit apart. It now enumerates all four dispositions, names both opt-outs, states explicitly that the no-403 case is skippable under no flag, and asks that the block be updated in the same commit as any disposition change.

e2e/tests/signup-verification.anon.spec.ts
```diff
@@ -25,7 +25,12 @@ test.describe('signup → email verification → login', () => {
 	test.skip(!mailConfigured, 'Mailosaur not configured (see e2e/.env.e2e).');
 
 	// Email delivery + verification is slower than a normal UI assertion.
-	test.setTimeout(120_000);
+	// Budget must exceed the sum of its own waits: 90s mail poll + three 30s navigations + the
+	// signup/login interactions. At 120s a slow-but-working delivery timed out and read as a failure.
+	test.setTimeout(210_000);
```
kriszyp: The new 210s test timeout is still smaller than this flow's configured waits: the 90s mail poll plus five 30s navigation/URL waits already total 240s, before form actions, assertions, and cleanup. A slow-but-successful run can therefore still hit the outer timeout even though no individual operation times out. Could we either raise the test timeout above the actual worst-case budget (with cleanup headroom) or reduce/coordinate the inner waits so the stated invariant holds?

e2e/tests/signup-verification.anon.spec.ts
```diff
@@ -25,7 +25,12 @@ test.describe('signup → email verification → login', () => {
...
+	// ...but that budget multiplies by the global retry count, and every attempt signs up a REAL
+	// account. Cap this spec at one retry: worst case ~7min and 2 accounts instead of ~10.5min and 3.
+	test.describe.configure({ retries: 1 });
```
kriszyp: `test.describe.configure({ retries: 1 })` overrides the project value in every environment, so direct local runs change from `retries: 0` to one retry. On failure that doubles this 210s flow and may create a second real account—the overhead this block is trying to cap. Could this preserve the local zero-retry behavior, for example with `retries: process.env.CI ? 1 : 0`, while still reducing the container/CI setting from two retries to one?

kriszyp (commented):
Great progress Dawson, almost there!

e2e/tests/signup-verification.anon.spec.ts
dawsontoth: You're right, and if anything the shortfall was worse than 240s once actions and cleanup are counted. Full sum: 5 navigations/waitForURL @30s (150) + mail poll (90) + waitForResponse (30) + 9 form actions @15s actionTimeout (135) + 2 expects @10s (20) + 2 cleanup API requests @30s (60) = **~485s**. Set to 600s in 00c3a6b0, and I put the arithmetic in the comment so a future change to any inner wait (especially `mail.ts`'s `timeoutMs`) can re-verify the invariant instead of guessing at it — that's twice now I've stated it without doing the sum.

e2e/tests/signup-verification.anon.spec.ts
dawsontoth: Good catch — fixed in 00c3a6b0 with exactly your suggestion, `retries: process.env.CI ? 1 : 0`. I'd missed that `test.describe.configure` overrides the project value in *every* environment, so the bare `retries: 1` handed local runs a retry they never had (project is `retries: 0` outside CI) — doubling a 600s flow and creating a second real account, i.e. precisely the overhead the cap was meant to reduce. Local is back to zero; CI goes from two retries to one.

PR #1574, opened 2026-07-28
fix(deploy): Version 5 Updates

Devin-Holland:
## Problem

After upgrading the dev CM to harper-pro 5.1.23, browsing the studio root returns:

```json
{"error":"reply.sendFile is not a function"}
```

and every static asset 404s. Boot log shows why:

```
[http/2] [error]: fastify-plugin: @fastify/static - expected '4.x' fastify version, '5.10.0' is installed
```

`@fastify/static@7` only supports fastify 4, so on Harper 5.x (fastify 5.10) the plugin registration fails and `reply.sendFile` never gets decorated — while the `/` route itself still registers and 500s on every hit.

## Fix

The deploy template ships **both** plugin majors under npm aliases (`fastify4-static` → `@fastify/static@^7`, `fastify5-static` → `@fastify/static@^8`) and picks the one matching `fastify.version` at registration time.

Why not just bump to ^8: this same template deploys to stage/prod CMs still on harperdb 4.7.33 (fastify 4), and the CM fleet upgrades to Harper 5.x one environment at a time. A hard bump would break whichever side doesn't match, and a dev-branch-only pin would be lost on the next branch promotion. Once every CM runs Harper 5.x this collapses back to a single `@fastify/static@^8` dependency (comment in `static.js` marks it).

## Verification

Registered the exact template module on both runtimes and validated responses:

| runtime | GET / | GET asset |
|---|---|---|
| fastify 5.10.0 | 200, index.html, CSP headers set | 200, `cache-control: public, max-age=2592000, immutable` |
| fastify 4.29.1 | 200, index.html, CSP headers set | 200, same |

## Follow-ups from review + dev deploy

- **Fail-closed version pick** (David's review): unknown fastify majors now throw at registration instead of silently picking a mismatched build — Harper's autoloader swallows registration errors, so a silent mismatch only ever surfaces as the runtime 500 this PR fixes. Verified: fastify `6.0.0`/`undefined` both throw.
- **Renovate mute hardened** (David's review): the `@fastify/static` disable rule now uses `matchDepNames` listing both npm aliases, so routine dep hygiene can't re-bump the fastify-4 build out from under 4.x CMs.
- **`--no-audit` on component install**: first dev deploy failed in `prepare` — npm 11.16 exits 1 (with no error output) when a package.json using npm aliases hits audit advisories, even though the install itself completes. The @fastify/static v7/v8 lines have permanent advisories (fixes only landed in v10+), so the template's `config.yaml` now sets `install.command: npm install --ignore-scripts --no-audit`. The `install` property is supported identically on harperdb 4.7.33 and harper-pro 5.1.23 (same `components/Application.ts` code). Reproduced + verified in the dev CM container: aliased install with audit → exit 1, with `--no-audit` → exit 0 with both aliases installed.

### DIFF
--- a/.github/deploy-template/package.json
+++ b/.github/deploy-template/package.json
@@ -1,5 +1,5 @@
 {
 	"dependencies": {
-		"@fastify/static": "^7.0.4"
+		"@fastify/static": "^8.3.0"
 	}
 }


dawsontoth (approved):

.github/deploy-template/fastify/static.js
```diff
@@ -1,7 +1,14 @@
...
+import fastifyStatic4 from 'fastify4-static';
+import fastifyStatic5 from 'fastify5-static';
 import { join } from 'path';
 
+// @fastify/static majors are pinned to fastify majors (v7 -> fastify 4, v8 -> fastify 5).
+// harperdb 4.x embeds fastify 4 and Harper 5.x embeds fastify 5, and this template deploys
+// to CMs on both, so pick the plugin build matching the running server. Collapse back to a
+// single @fastify/static dependency once every CM is on Harper 5.x.
 export default async (fastify) => {
+	const fastifyStatic = (fastify.version && Number.parseInt(fastify.version, 10) >= 5) ? fastifyStatic5 : fastifyStatic4;
```
DavidCockerill: Non-blocking: worth making the pick fail closed.

```js
const major = Number.parseInt(fastify.version, 10);
if (major !== 4 && major !== 5) throw new Error(`No @fastify/static build for fastify ${fastify.version}`);
const fastifyStatic = major === 5 ? fastifyStatic5 : fastifyStatic4;
```

As written, an `undefined` version gives `NaN >= 5` → false → the v7 build, and a future fastify 6 gets the v8 build; both are mismatches. Harper's autoloader swallows plugin registration errors (`.after(...)` logs and calls `next()` in `core/server/fastifyRoutes.ts`), so a wrong pick can only ever surface as a 500 on the studio root — exactly the symptom this PR is fixing. An explicit throw is the only way an unknown runtime shows up at boot.

This also covers the real half of @gemini-code-assist's note; the radix itself is a non-issue since `parseInt('5.10.0')` infers base 10.

.github/deploy-template/package.json
```diff
@@ -1,5 +1,6 @@
 {
 	"dependencies": {
-		"@fastify/static": "^7.0.4"
+		"fastify4-static": "npm:@fastify/static@^7.0.4",
```
DavidCockerill: Worth confirming the Renovate mute still catches these. `renovate.json` has `{ "matchPackageNames": ["@fastify/static"], "enabled": false }`, which exists precisely to stop a major bump breaking the fastify-4 CMs.

For an npm alias, Renovate sets `depName` to the alias (`fastify4-static`) and `packageName` to `@fastify/static`. I believe modern Renovate matches `matchPackageNames` against `packageName`, so it should still apply — but I'm not fully confident, and if it doesn't, Renovate proposes `fastify4-static` → `@fastify/static@^9`, someone merges it as routine dep hygiene, and stage/prod hit this exact outage again.

Cheap certainty: add the alias names to `matchPackageNames`, or add `"matchDepNames": ["fastify4-static", "fastify5-static"]`.

DavidCockerill (approved):
Approving. Aliasing both majors is the right call for a fleet mid-upgrade, and the both-runtimes verification table in the description is solid.

Checked the things that usually bite on a dual-serve change and none of them apply: only one plugin ever registers, so there's no duplicate or stale asset and no route-precedence question, and CSP, CORS and cache policy are all unchanged.

Two non-blocking notes in threads, both about future decay rather than today — the version pick currently fails open on an unexpected fastify version, and the Renovate mute may not match npm aliases, which would let a routine dep bump reproduce the outage this fixes.

.github/deploy-template/package.json
dawsontoth: i don't think you did a real install with these

.github/deploy-template/fastify/static.js
Devin-Holland: Applied in 2176c11d — exact snippet, with a comment noting the autoloader-swallows-errors rationale. Verified all three paths in a live harness: fastify 4.29.1 serves, 5.10.0 serves, and 6.0.0/undefined both throw at registration. Good catch on NaN silently picking the v7 build — that failure mode is precisely what this PR is cleaning up after.

— devain (Claude Fable 5)

.github/deploy-template/package.json
Devin-Holland: Applied in 2176c11d. Went with your matchDepNames option, listing all three names — depName is always the package.json key, so it matches the aliases unambiguously regardless of how a given Renovate version resolves matchPackageNames vs packageName. Also added a description to the rule so the mute's reason survives the next config cleanup. Thanks for chasing the second-order failure here — a routine dep-hygiene merge re-breaking prod CMs would have been a nasty repeat.

— devain (Claude Fable 5)

Devin-Holland:
Handoff notes (Dawson's taking the branch from here):

**1. Verify Commits is failing — two commits lack Conventional Commits prefixes** (on both `fix/deploy-template-fastify5-static` and the `-dev` variant, and now on `dev` itself via the force-push):
- `Fail closed on unknown fastify major; mute Renovate on the static aliases` → suggest `fix(deploy): fail closed on unknown fastify major; mute Renovate on the static aliases`
- `Disable npm audit in component install (npm 11 + alias exit-1 bug)` → suggest `fix(deploy): disable npm audit in component install (npm 11 + alias exit-1 bug)`

They're the top two commits on each branch, so a reword + `--force-with-lease` fixes it. Apologies for the churn — documented the convention in CLAUDE.md via #1575 so it doesn't recur.

**2. The `install.command` fix did NOT take effect on the last dev deploy** — [run 30399339123](https://github.com/HarperFast/studio/actions/runs/30399339123) deployed the right commit (`VITE_STUDIO_VERSION: dev_1be446e`) but the prepare phase still logged `install: using npm` + `npm warn using --force` and failed with `Failed to install dependencies for hdbms using npm default` — the default install path, not the custom command. So `config.yaml`'s `install.command` appears not to be consulted during deployComponent's prepare on harper-pro 5.1.23 (unconfirmed why — I was mid-investigation in `core/components/Application.ts`; `application.install` is evidently undefined at that point). Note the default path runs `npm install --force --ignore-scripts`, and npm 11.16 exits 1 on this package.json because of the alias+audit bug (reproduced in the dev CM container; `--no-audit` fixes it, details in the PR body).

Fallback if `install.command` can't be made to work at prepare time: pre-install `node_modules` in the deploy dir during the workflow and pass `skip_node_modules=false` to `harper deploy_component` — the server skips install entirely when `node_modules` exists.

— devain (Claude Fable 5)

kriszyp (approved):

dawsontoth:
I'm flagging this as a draft, do NOT merge this until Central Manager is totally ready with v5 of Harper!

PR #1587, opened 2026-07-29
fix(status): stop the connections chart diving to ~0 on single-node buckets

### LINKED ISSUE #1576: Connections chart dives to ~0 when only one node reports in a time bucket (cross-node sum treats absent node as zero)
Customer-reported (IBM, `it-mil-1.stage2.ibm-nonprod`, harper-pro 5.1.22). The Status → Traffic → **Connections** panel shows frequent single-sample dives to ~0, while Grafana over the same instances and window shows no such drops. The customer was load-testing and monitoring connections closely; nothing actually dropped.

**The dives are a rendering artifact. Harper never reports a zero.**

The tell: each dive's depth equals the *quiet* node's own connection count at that instant, not 0. One spike in the customer's screenshot bottoms at ~100 rather than 0 — that's the Milan node sitting at ~100 connections at the time. A real connection loss can't do that.

## Root cause

1. Core emits `mqtt-connections` only when the count is nonzero — `if (numberOfConnections > 0)` in harper [`server/mqtt.ts`](https://github.com/HarperFast/harper/blob/main/server/mqtt.ts) (~L177), pushed via `addAnalyticsListener` with `byThread: true`. A node at 0 connections emits **no row at all**, so per-node coverage is inherently ragged.
2. Those gauge rows carry **`period: 0`** (no aggregation period stamped).
3. `resolvePeriod()` (`pipeline.ts:100`) treats `0` as invalid and falls back to `spec.bucket.fallbackMs` = `60_000`. `snapToBucketTime()` (`pipeline.ts:89`) then rounds every row's time onto a 60 s lattice. The observed emission cadence is **90 s**, and 90 mod 60 = 30, so the two nodes' phases beat against the lattice and co-land in the same bucket only intermittently.
4. `crossNode: 'sum'` (`wrapperMetrics.tsx:337`) sums **only the nodes present in that bucket** (`aggregateTwoPass`, `pipeline.ts:453`). A bucket holding only the quiet node renders the cluster total as that node's 1–5 connections instead of ~289 → a one-sample vertical dive to ≈0.

## Evidence

Real `get_analytics` rows, 24 h window, both nodes:

```
Does Harper EVER report a zero/negative? -> False
i5w-us-west-1.stage2...   rows=960  min=283  max=294  period=0
vk5-it-mil-1.stage2...    rows=228  min=1    max=5    period=0
```

960 = 86400/90 exactly — us-west-1 reported every period without a miss. Milan reported 228 times, only when it had ≥1 connection.

Feeding those rows through studio's own `runPipeline` with production settings (`{ snapToPeriod: true }`) and the verbatim `connections` spec reproduces the artifact:

```
studio pipeline emitted 969 points for the 'mqtt' series
median normal y = 285.0
Milan-only buckets (rendered as dives to ~0): 9
   t=2026-07-28T08:24:00Z  y=1      t=2026-07-28T19:27:00Z  y=1
   t=2026-07-28T09:00:00Z  y=1      t=2026-07-28T20:39:00Z  y=1
   ...
minimum y anywhere in the series: 1   <-- the "drop to 0 connections"
```

Bucket census: 219 both-nodes, 741 us-west-only, **9 Milan-only** — exactly the 9 dives.

## Same latent bug

`database-size` (`wrapperMetrics.tsx:358`) has the identical shape — temporal `last` + `crossNode: 'sum'`, stacked-area — so it dives the same way whenever one node misses a bucket. `table-size` escapes it by keeping per-node values separate instead of summing.

## Suggested fix

Bucket alignment alone is **not** sufficient: core's `> 0` guard guarantees ragged per-node coverage no matter how well we bucket.

The cross-node sum needs per-node last-observation-carry-forward with a **bounded staleness horizon** (~2 emission intervals, the Prometheus staleness approach) before summing. The bound matters because "no row" is genuinely ambiguous between *"this node's sample landed in the adjacent bucket"* (carry forward) and *"this node is idle at 0"* (don't). Unbounded LOCF would overstate a node that has genuinely gone quiet.

This should key on gauge-shaped specs — `crossNode: 'sum'` with a temporal aggregator of `max`/`last` — and must **not** apply to additive counters like `mqtt-traffic-*` (`sum`/`sum`), where a missing row legitimately means zero.

## Test gap

`__tests__/pipeline/connections-pipeline.test.ts` only covers uniform `period: 60000` with every node present at every timestamp, so this class of bug can't surface. Worth a case with staggered per-node timestamps and a node missing from a bucket — the `table-size` suite already has `node-drop.json` / `sparse-per-node.json` fixtures for exactly this shape.

## Secondary findings (separate, non-blocking)

- Studio sends `bucket_ms` (300 000 for the 24 h preset) but harper-pro 5.1.22 **ignores it** — rows come back at raw cadence, so the client-side 60 s snap does all the bucketing regardless of the selected window.
- `spec.bucket.fallbackMs` of 60 000 didn't match this instance's actual 90 s cadence. Core stamping a real `period` on gauge metrics would remove the guess entirely — worth a companion core issue.
- The instance had `analytics.aggregatePeriod: 60` configured while rows actually landed 90 s apart. Unexplained; may be worth a core look independently.

dawsontoth:
Fixes #1576.

Customer-reported (harper-pro 5.1.22): the Status → Traffic → **Connections** panel showed frequent one-sample dives to near zero, while Grafana over the same instances and window showed none. The customer was load-testing and watching connections closely; nothing actually dropped.

**Harper never reports a zero.** The tell is that each dive's depth equals the *quiet* node's own connection count at that instant, not 0 — one spike in the customer's screenshot bottoms at ~100 rather than 0, which is the Milan node sitting at ~100 connections at the time. A real connection loss can't do that.

## Root cause

Three things compound:

1. Core emits `mqtt-connections` only when the count is nonzero — `if (numberOfConnections > 0)` in harper's [`server/mqtt.ts`](https://github.com/HarperFast/harper/blob/main/server/mqtt.ts) (~L177). A node at 0 connections emits **no row at all**, so per-node coverage of the gauge is inherently ragged.
2. Those rows carry **`period: 0`**, so `resolvePeriod` falls back to the spec's `bucket.fallbackMs` (60 s) and `snapToBucketTime` rounds every row onto a 60 s lattice — while the observed emission cadence is **90 s**. 90 mod 60 = 30, so the two nodes' phases beat against the lattice and co-land in the same bucket only intermittently.
3. `crossNode: 'sum'` then sums **only the nodes present in that bucket**. A bucket holding just the quiet node renders the cluster total as that node's 1–5 connections instead of ~289 — a one-sample vertical dive to ≈0.

## What changed

Bucket alignment alone can't fix this: core's `> 0` guard guarantees ragged coverage however well we bucket. So gauge-shaped specs now get **per-node last-observation-carry-forward with a bounded staleness horizon** before the cross-node sum.

New module [`pipeline/carryForward.ts`](src/features/instance/status/analytics/pipeline/carryForward.ts) holds the policy; [`pipeline.ts`](src/features/instance/status/analytics/pipeline/pipeline.ts) grows one `aggregateOverTime` helper that replaces the three places which previously looped `aggregateTwoPass` over sorted bucket times (groupBy series, the `Other` bucket, and field-mode series).

A node absent from a bucket contributes its most recent value, but only for **~2 of its own emission intervals** — estimated as the median gap between that node's observation instants, floored at the spec's bucket size. The bound is the point: "no row" is genuinely ambiguous between *"this node's sample landed in the adjacent bucket"* (carry it) and *"this node is idle at 0"* (don't). Unbounded LOCF would overstate a node that has genuinely gone quiet. Median rather than mean so a node that keeps crossing zero is still credited with its active cadence instead of one inflated by the absences.

The cadence is measured from **raw, pre-snap** instants — the estimator has to see the true 90 s cadence, not the 60/120 alternation the lattice snap produces.

### Scope

Keys on `crossNode: 'sum'` with a temporal aggregator of `max`/`last`, i.e. a cross-node sum of per-node *level* snapshots. Today that is exactly:

- `connections` (max/sum) — the reported panel
- `database-size` (last/sum) — **the identical latent bug**, called out in the issue

Additive counters (`sum`/`sum` — `mqtt-traffic-*`, `bytes-sent`/`bytes-received`, `fsWrite`) are deliberately excluded: there a missing row legitimately means zero, and carrying the previous bucket's throughput forward would invent traffic that never happened. There's a test asserting this for `bytes-sent`.

### Invariants preserved

- **Carry-forward never creates bucket times.** A bucket where no node reported stays absent, so the series keeps its original x positions and a genuine cluster-wide gap still reads as a gap.
- **Carried values carry `count: 0`**, so `SeriesPoint.count` — which drives confidence gating and the tooltip's sample count — keeps reflecting real observations only.
- **Per-node mode is untouched.** The chip-selector and "Stack by: Node" views read per-node values directly; carry-forward belongs to the cross-node sum alone.
- **LOCF only looks backwards** — buckets before a node's first sample are not back-filled.

## Verification

**The new tests fail without the fix.** Gating `isGaugeCrossNodeSum` to `false` produces `expected [300, 300, 100, 300] to deeply equal [300, 300, 300, 300]` — the dive to one node's own value, reproduced exactly.

**Replaying the reported row shape** (960 + 228 rows over 24 h, `period: 0`, 90 s cadence, independent phases) reproduces the issue's evidence block and then clears it:

|  | before | after |
| --- | --- | --- |
| minimum y in series | **1** | **283** |
| points below 200 (dives) | **114** | **0** |
| bucket census | 114 both · 846 us-west-only · 114 milan-only | unchanged |

**Live on a real 2-node stage cluster** (`anvils.acme-inc.stage`). I temporarily instrumented `aggregateOverTime` and loaded the Storage tab: `nodes=2 buckets=1440 incompleteBuckets=960 carried=959 expired=0`, horizons ≈ 180 002 ms. Two things worth a reviewer's attention:

- **This bug was live on our own stage cluster too** — 67% of `database-size` buckets held only one node, across all five database bands. It just reads as a plausible-looking wobble rather than an obvious dive, because the bands are large and slow-moving.
- The cadence estimator independently landed on ~90 s there as well, so `fallbackMs: 60000` is systematically wrong for these gauge metrics. That corroborates the issue's secondary finding that core stamping a real `period` would remove the guess entirely.

Probe removed afterwards; the panel renders a clean continuous 24 h stack at the full two-node total, all five bands, no error-boundary fallback.

## Tests

The existing `connections-pipeline.test.ts` / `database-size-pipeline.test.ts` suites only exercise uniform `period: 60000` with every node present at every timestamp, so this class of bug could not surface there. Added:

- `__tests__/pipeline/cross-node-gauge-gaps.test.ts` — staggered 90 s cadence with lattice phase beating (including a guard asserting the setup really does produce single-node buckets, so nothing passes vacuously); horizon-expiry when a node goes genuinely idle; cluster-wide gaps staying gaps; no node resurrected across an outage; point-count and `count` invariants; `database-size` same-shape case; `bytes-sent` counter-untouched case.
- `__tests__/pipeline/carryForward.test.ts` — unit coverage for the shape predicate, the median-gap estimator (order independence, duplicate instants, scattered absences), and horizon construction.
- `__tests__/fixtures/connections/node-drop.json` — minimal hand-checkable node-drop fixture, mirroring the `table-size/node-drop.json` precedent the issue points at.

Full repo gate green: **1881 tests / 264 files**, `tsc -b`, `oxlint`, `dprint check`.

## Deliberately out of scope

- **Bucket alignment.** The issue notes it's insufficient on its own, and changing `snapToBucketTime` would affect every spec for no gain here.
- **A spec-level opt-out knob.** The gauge shape is derived from one source of truth (`isGaugeCrossNodeSum`); easy to add a knob later if reviewers prefer it explicit per spec.
- **`StackedAreaChart`'s render-layer forward-fill** ([lines 98–113](src/features/instance/status/analytics/primitives/StackedAreaChart.tsx#L98-L113)) is *unbounded*. That's why "Stack by: Node" mode never showed these dives — but it also means a genuinely idle node's band persists for the rest of the window, which is now inconsistent with the bounded horizon one layer up. It touches every stacked-area panel, so it wants its own change.
- The issue's other secondary findings (harper-pro ignoring `bucket_ms`; `analytics.aggregatePeriod: 60` vs rows landing 90 s apart) are core-side and want a companion core issue.

### DIFF
--- a/src/features/instance/status/analytics/context/timePresets.ts
+++ b/src/features/instance/status/analytics/context/timePresets.ts
@@ -31,6 +31,35 @@ export function getPreset(id: TimePresetId): TimePreset {
 	return p;
 }
 
+/**
+ * Densest bucket the charts should *render* for a window of `windowMs`.
+ *
+ * The same numbers the presets declare, looked up by duration instead of id.
+ * That matters because `bucketMs` reaches the server as `bucket_ms` but is
+ * otherwise unused client-side: builds that ignore the hint (harper-pro 5.1.22)
+ * return rows at raw emission cadence, and the pipeline then buckets them onto
+ * `spec.bucket.fallbackMs` — 60 s regardless of the selected window. A 30 d
+ * window rendered 43 200 points instead of 720 (#1576 follow-up).
+ *
+ * Reading the same table by duration keeps one source of truth: StorageTab
+ * aligns its trend grid to the context's `bucketMs` (#1514), so a preset-sized
+ * window MUST resolve to exactly that preset's value or the two grids diverge
+ * and crosshair sync breaks. `targetBucketMs` locks that (see timePresets.test).
+ */
+export function targetBucketMs(windowMs: number): number {
+	const finest = TIME_PRESETS[0].bucketMs;
+	if (!Number.isFinite(windowMs) || windowMs <= 0) { return finest; }
+	for (const p of TIME_PRESETS) {
+		// 1% slack: a live `endTime = Date.now()` bound makes the computed
+		// window drift a few ms off the preset's exact duration.
+		if (windowMs <= p.durationMs * 1.01) { return p.bucketMs; }
+	}
+	// Past the widest preset, hold its points-per-window ratio rather than
+	// falling back to the finest bucket and re-creating the density problem.
+	const widest = TIME_PRESETS[TIME_PRESETS.length - 1];
+	return Math.max(finest, Math.ceil(windowMs / (widest.durationMs / widest.bucketMs)));
+}
+
 export interface RefreshOption {
 	label: string;
 	value: number;
--- a/src/features/instance/status/analytics/context/timePresets.ts
+++ b/src/features/instance/status/analytics/context/timePresets.ts
@@ -45,6 +45,20 @@ export function getPreset(id: TimePresetId): TimePreset {
 	return p;
 }
 
+/** Human phrase for a bucket duration — "1 minute", "10 minutes", "4 hours".
+ *  Drives the "by X" sub-label under each range option so the operator can see
+ *  the resolution a window renders at (the server ignores our `bucket_ms`, so
+ *  the panel resolution is `targetBucketMs`, i.e. the preset's `bucketMs`).
+ *  Every shipped bucket is a whole number of minutes or hours; the seconds /
+ *  raw-ms branches are just defensive for a future odd value. */
+export function formatBucketLabel(bucketMs: number): string {
+	const plural = (n: number, unit: string) => `${n} ${unit}${n === 1 ? '' : 's'}`;
+	if (bucketMs >= HOUR && bucketMs % HOUR === 0) { return plural(bucketMs / HOUR, 'hour'); }
+	if (bucketMs >= MIN && bucketMs % MIN === 0) { return plural(bucketMs / MIN, 'minute'); }
+	if (bucketMs >= 1000 && bucketMs % 1000 === 0) { return plural(bucketMs / 1000, 'second'); }
+	return `${bucketMs} ms`;
+}
+
 /**
  * Densest bucket the charts should *render* for a window of `windowMs`.
  *
--- a/src/features/instance/status/analytics/context/timePresets.ts
+++ b/src/features/instance/status/analytics/context/timePresets.ts
@@ -6,7 +6,12 @@ export interface TimePreset {
 	id: TimePresetId;
 	label: string;
 	durationMs: number;
+	/** Bucket for a panel-sized chart — sized so the window lands in
+	 *  `PANEL_POINT_TARGET`. Also the `bucket_ms` hint sent to Harper. */
 	bucketMs: number;
+	/** Bucket when the chart is expanded to the near-fullscreen dialog, where
+	 *  there is roughly 4x the horizontal room. Always ≤ `bucketMs`. */
+	expandedBucketMs: number;
 }
 
 export type TimePresetId = '1h' | '6h' | '24h' | '7d' | '30d';
@@ -15,12 +20,21 @@ const MIN = 60_000;
 const HOUR = 60 * MIN;
 const DAY = 24 * HOUR;
 
+/** Points a panel-sized chart aims for. Above this the line is denser than the
+ *  panel has pixels, so the extra samples cost payload and render time and buy
+ *  nothing; the 1h preset undershoots because Harper's own aggregate period is
+ *  60 s and there is no finer data to ask for. */
+export const PANEL_POINT_TARGET = { min: 100, max: 200 } as const;
+
 export const TIME_PRESETS: readonly TimePreset[] = [
-	{ id: '1h', label: 'Last 1 hour', durationMs: HOUR, bucketMs: 1 * MIN },
-	{ id: '6h', label: 'Last 6 hours', durationMs: 6 * HOUR, bucketMs: 1 * MIN },
-	{ id: '24h', label: 'Last 24 hours', durationMs: DAY, bucketMs: 5 * MIN },
-	{ id: '7d', label: 'Last 7 days', durationMs: 7 * DAY, bucketMs: 15 * MIN },
-	{ id: '30d', label: 'Last 30 days', durationMs: 30 * DAY, bucketMs: HOUR },
+	// bucketMs is chosen to land `durationMs / bucketMs` inside
+	// PANEL_POINT_TARGET; expandedBucketMs gives the dialog ~3-4x the detail.
+	//                                                              panel  expanded
+	{ id: '1h', label: 'Last 1 hour', durationMs: HOUR, bucketMs: 1 * MIN, expandedBucketMs: 1 * MIN }, //   60    60
+	{ id: '6h', label: 'Last 6 hours', durationMs: 6 * HOUR, bucketMs: 2 * MIN, expandedBucketMs: 1 * MIN }, // 180   360
+	{ id: '24h', label: 'Last 24 hours', durationMs: DAY, bucketMs: 10 * MIN, expandedBucketMs: 5 * MIN }, // 144   288
+	{ id: '7d', label: 'Last 7 days', durationMs: 7 * DAY, bucketMs: HOUR, expandedBucketMs: 15 * MIN }, // 168   672
+	{ id: '30d', label: 'Last 30 days', durationMs: 30 * DAY, bucketMs: 4 * HOUR, expandedBucketMs: HOUR }, // 180   720
 ];
 
 export const DEFAULT_PRESET_ID: TimePresetId = '1h';
@@ -46,18 +60,19 @@ export function getPreset(id: TimePresetId): TimePreset {
  * window MUST resolve to exactly that preset's value or the two grids diverge
  * and crosshair sync breaks. `targetBucketMs` locks that (see timePresets.test).
  */
-export function targetBucketMs(windowMs: number): number {
-	const finest = TIME_PRESETS[0].bucketMs;
+export function targetBucketMs(windowMs: number, opts?: { expanded?: boolean }): number {
+	const pick = (p: TimePreset) => (opts?.expanded ? p.expandedBucketMs : p.bucketMs);
+	const finest = pick(TIME_PRESETS[0]);
 	if (!Number.isFinite(windowMs) || windowMs <= 0) { return finest; }
 	for (const p of TIME_PRESETS) {
 		// 1% slack: a live `endTime = Date.now()` bound makes the computed
 		// window drift a few ms off the preset's exact duration.
-		if (windowMs <= p.durationMs * 1.01) { return p.bucketMs; }
+		if (windowMs <= p.durationMs * 1.01) { return pick(p); }
 	}
 	// Past the widest preset, hold its points-per-window ratio rather than
 	// falling back to the finest bucket and re-creating the density problem.
 	const widest = TIME_PRESETS[TIME_PRESETS.length - 1];
-	return Math.max(finest, Math.ceil(windowMs / (widest.durationMs / widest.bucketMs)));
+	return Math.max(finest, Math.ceil(windowMs / (widest.durationMs / pick(widest))));
 }
 
 export interface RefreshOption {
--- a/src/features/instance/status/analytics/pipeline/carryForward.ts
+++ b/src/features/instance/status/analytics/pipeline/carryForward.ts
@@ -22,13 +22,36 @@
 // staleness window; we scale to each node's *observed* cadence instead,
 // because Harper's analytics cadence varies per deployment (the instance in
 // #1576 had `analytics.aggregatePeriod: 60` configured while rows actually
-// landed 90 s apart).
+// landed 90 s apart) — but clamped by Prometheus' 5 minutes as an absolute
+// backstop, so a bad cadence estimate can't turn bounded staleness into
+// carry-forever. See MAX_STALENESS_MS.
 import type { Aggregator } from '../types/analytics';
 
 /** Multiplier on a node's estimated emission interval that bounds how long its
  *  last observation may be carried forward. Two intervals absorbs one missed
  *  emission plus bucket-lattice phase drift, and no more. */
 export const STALENESS_INTERVALS = 2;
+
+/**
+ * Absolute ceiling on a carry-forward horizon, independent of what the cadence
+ * estimate says. Prometheus' fixed 5-minute lookback, used here as a backstop
+ * rather than as the whole policy.
+ *
+ * The median is only robust while short, normal-cadence gaps are the majority.
+ * Two pathological-but-real shapes break that:
+ *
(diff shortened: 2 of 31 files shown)


src/features/instance/status/analytics/pipeline/pipeline.ts
```diff
@@ -302,7 +322,23 @@ function runGroupBy(
...
+						let merged = otherObservationTimes.get(node);
+						if (!merged) {
+							merged = [];
+							otherObservationTimes.set(node, merged);
+						}
+						// Append by loop, not spread — argument-spread blows the V8
+						// stack past ~125k elements and a long window can hold more.
+						for (const t of times) { merged.push(t); }
+					}
+				}
```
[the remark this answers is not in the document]
dawsontoth: Done in 91fd96c2 — hoisted to `const dimObservationTimes = observationTimes?.get(key)` and guarded, dropping the non-null assertion. The assertion was sound (`otherObservationTimes` is non-null exactly when `observationTimes` is) but that coupling was implicit; this makes it checkable.

src/features/instance/status/analytics/pipeline/pipeline.ts
```diff
@@ -445,6 +488,80 @@ function runFieldSpecs(
...
+			observed.add(node);
+			lastObserved.set(node, { time, value: nodeY });
+		}
+		for (const [node, last] of lastObserved) {
+			if (observed.has(node)) { continue; }
+			if (time - last.time > (horizonByNode.get(node) ?? 0)) { continue; }
+			// Carried, not observed — count 0 so `SeriesPoint.count` (confidence
+			// gating, tooltips) keeps reflecting real samples only.
+			inputs.push({ value: last.value, count: 0 });
+		}
```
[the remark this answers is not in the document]
dawsontoth: Done in 91fd96c2 — evicting the node once it ages past its horizon, since ascending bucket order means that observation can only get staler. Note the map is bounded by node count, not bucket count, so it is not the unbounded leak it might look like; the real win is a wide window over a cluster whose node ids churn (rolling replacement, autoscaling), where it would otherwise accumulate every node ever seen. Added a regression test for the one live edge — a node that ages out and later returns must be re-registered, not permanently excluded.

src/features/instance/status/analytics/pipeline/pipeline.ts
```diff
@@ -173,11 +186,27 @@ function runGroupBy(
...
+	const crossAgg: Aggregator = src.field.aggregator?.crossNode ?? spec.aggregator.crossNode;
+	const isApprox = tempAgg === 'count-weighted-mean' || crossAgg === 'count-weighted-mean';
+	// When the spec already groups by node, perNode is redundant — emitting
+	// one series per (node, node) would duplicate. Fall through to the
+	// cluster-aggregate path which, for dimension='node', is naturally
+	// one-series-per-node.
+	const dimensionIsNode = src.dimension === 'node';
+	// Bounded per-node carry-forward only applies where the crossNode pass
+	// actually runs, and only for gauge-shaped specs (see carryForward.ts).
+	const carryForward = (!perNode || dimensionIsNode) && isGaugeCrossNodeSum(tempAgg, crossAgg);
```
cb1kenobi: **Low: carry-forward bookkeeping is a no-op when `dimensionIsNode`**

`carryForward` is forced `true` here whenever `dimensionIsNode` (the "Stack by: Node" view for `connections`/`database-size`), independent of `perNode`. But when `src.dimension === 'node'`, `dimVal` and the loop's `node` (both derived from `r.node`) are always the same value, so every per-dimension time bucket can only ever contain that one node. The crossNode `sum` pass over a single input is trivially identity — there is never an "absent node" for `aggregateOverTime`'s carry-forward to fill in. This branch still pays for it: every record push now also calls `recordObservation`, and `buildStalenessHorizons` sorts/medians each node's full observation-time array for a horizon that's never consulted.

Not a correctness bug (output is unchanged), but it's wasted work on every render of the per-node view, scaling with record count — worth trimming given #1588 is tightening render cost in this same pipeline.

Suggested fix:
```suggestion
	const carryForward = !perNode && isGaugeCrossNodeSum(tempAgg, crossAgg);
```
The `dimensionIsNode` fall-through to the cluster-aggregate path (line 262's `if (perNode && !dimensionIsNode)` check) is untouched by this — it only stops collecting/building carry-forward state that this branch can't use.

—
_Generated by Barber AI_

dawsontoth: Good catch — trimmed in e4614fc1. Applied it as `!perNode && !dimensionIsNode && isGaugeCrossNodeSum(...)` rather than the literal `!perNode && …` suggestion, because the one-liner only changes the `perNode && dimensionIsNode` combo: for the "Stack by: Node" view you cite, the remap runs with `perNode=false`, so both the original `(!perNode || dimensionIsNode)` and `!perNode` evaluate to `true` there and carry-forward would stay on. Gating on `!dimensionIsNode` skips the observation-recording + horizon build in *both* node-dimension cases, which is the whole set where the crossNode sum is single-node identity and the bookkeeping is dead. Output is unchanged (verified: connections / database-size / cross-node-gauge-gaps / per-node suites stay green), so it only stops collecting state that path can't use.

src/features/instance/status/analytics/pipeline/carryForward.ts
```diff
@@ -0,0 +1,94 @@
...
+ * bucket, or when it only ever reported once.
+ */
+export function buildStalenessHorizons(
+	observationTimesByNode: ReadonlyMap<string, number[]>,
+	floorMs: number,
+): Map<string, number> {
+	const horizons = new Map<string, number>();
+	for (const [node, times] of observationTimesByNode) {
+		const cadence = estimateEmissionIntervalMs(times) ?? 0;
+		horizons.set(node, STALENESS_INTERVALS * Math.max(cadence, floorMs));
```
kriszyp: Could we put an absolute cap on this horizon? The median makes the estimate robust only when short, normal-cadence gaps are the majority. A quiet node with observations at 00:00 and 12:00 gets a 12-hour cadence and therefore a 24-hour horizon; after the second brief nonzero observation, its stale connection count is added to every bucket supplied by another node for the rest of the window. A three-sample series with one short gap and one outage also averages the two gaps into a multi-hour median. That defeats the stated bounded-staleness guarantee and can hide a real transition to zero or a node outage. Please clamp the cadence/horizon to a fixed or spec-defined maximum (the cited five-minute model would be one option) and add a regression with two widely separated observations followed by regular buckets from another node.

kriszyp (commented):
Looks good, one comment

src/features/instance/status/analytics/pipeline/carryForward.ts
dawsontoth: Good catch — fixed in 512ab7a1 with `MAX_STALENESS_MS = 300_000` (Prometheus' 5 min, which this module already cited). Applied to the *cadence estimate* rather than the horizon:

```
horizon = STALENESS_INTERVALS * max(min(cadence, MAX/INTERVALS), floorMs)
```

Capping the cadence keeps the floor meaningful — on a spec whose own bucket is coarser than 5 min, spanning two buckets still matters more than the backstop, so the floor wins there. Net guarantee is now `horizon <= max(MAX_STALENESS_MS, 2 x floorMs)`, with an invariant test over pathological observation shapes. The #1576 case is untouched (90 s cadence → 180 s horizon).

Added the regression you asked for — two widely separated observations (00:00, 12:00) followed by regular buckets from another node — plus unit coverage for the three-sample one-gap-one-outage median and the floor-beats-cap interaction. I verified they fail without the cap, and the numbers match your description exactly: an 86 400 000 ms (24 h) horizon, a ~6 h median from the mixed-gap shape, and **714 tail buckets** reading 288 (BUSY+QUIET) instead of 285 — i.e. the stale count pinned into every bucket for the rest of the window.

Fair hit on process, too: I ran into this exact edge while writing the earlier eviction test and only documented it in a comment instead of fixing it.

cb1kenobi (approved):

src/features/instance/status/analytics/context/timePresets.ts
```diff
@@ -6,7 +6,12 @@ export interface TimePreset {
 	id: TimePresetId;
 	label: string;
 	durationMs: number;
+	/** Bucket for a panel-sized chart — sized so the window lands in
+	 *  `PANEL_POINT_TARGET`. Also the `bucket_ms` hint sent to Harper. */
 	bucketMs: number;
+	/** Bucket when the chart is expanded to the near-fullscreen dialog, where
+	 *  there is roughly 4x the horizontal room. Always ≤ `bucketMs`. */
+	expandedBucketMs: number;
```
kriszyp: Could we make the fetch resolution support this expanded resolution? `StatusTabs` always sends `preset.bucketMs` as `bucket_ms`, and `ChartExpandButton` re-renders the same fetched records. When Harper honors that hint, a 30-day response already has one sample per 4 hours, so folding it onto `expandedBucketMs = 1 hour` cannot create the promised extra detail; the current regression passes only because it supplies raw 60-second rows, modeling servers that ignore the hint. Please either fetch at `expandedBucketMs` and fold only the panel to `bucketMs`, or refetch at the finer bucket when the dialog opens, and cover the server-honored input shape.

Issue #1514, opened 2026-07-15
Cross-panel Status crosshair sync is intermittent (syncMethod="value" needs exact bucket alignment)


dawsontoth:
Spotted while reviewing #1507 (which fixes #1454, syncing chart crosshairs across a Status tab). Filing separately from the tooltip-presentation concern because the fix surface is different — this is about *when* sync fires.

### What happens
Cross-panel crosshair sync is **intermittent**. Charts inside a single small-multiples group (e.g. Resource usage) always light up together, but hovering one full-width panel only *sometimes* draws a synced crosshair on the other panels on the same tab.

### Why
The feature uses `syncMethod="value"` (chosen deliberately — index-based sync mis-aligns sparse series with different point counts). Value-matching requires the hovered x (a bucket timestamp) to **exist exactly** in the other chart's data:
- Small-multiples share one metric's response → identical bucket timestamps → always sync.
- Across panels (different metrics), buckets usually align (same `bucketMs` + epoch snapping) but not always — a metric missing that bucket, or snapped to a different grid, won't match and stays dark.
- #1507 already documents one guaranteed miss: `TableSizeTrend`'s grid is anchored at `range.startTime` (`lib/tableSize.ts`), not epoch-aligned, so the Storage-tab trend "usually won't light up."

### Possible directions (triage as you see fit)
- Decide whether intermittent cross-panel sync is acceptable for v1 (it degrades gracefully — no ghost crosshairs, just nothing).
- Or make bucket grids coincide across metrics (epoch-align the trend grid; ensure per-metric bucket timestamps match) so value-sync is reliable.

Discovered via browser verification of #1507. Related: #1507.

PR #1588, opened 2026-07-29
perf(status): render charts at the selected window's resolution

dawsontoth:
**Stacked on #1587** — base branch is `claude/mqtt-status-chart-gaps-ace583`, not `stage`. Review #1587 first; this PR's diff is only its own commit. Retarget to `stage` once #1587 merges.

Follow-up to a secondary finding in #1576.

## The problem

Each time preset declares the densest bucket Harper should serve, and those numbers are sensible:

```
1h → 1m    6h → 1m    24h → 5m    7d → 15m    30d → 1h
```

But that value only ever reached the server, as the `bucket_ms` hint. Client-side, `snapToBucketTime` used `record.period || spec.bucket.fallbackMs` — **60 s regardless of the selected window**. Builds that ignore `bucket_ms` (harper-pro 5.1.22, per #1576) return rows at raw emission cadence, so what actually rendered was:

| preset | intended | rendered | ratio | rows fetched |
| --- | --- | --- | --- | --- |
| 1h | 60 | 60 | 1× | 80 |
| 6h | 360 | 360 | 1× | 480 |
| 24h | 288 | **1 440** | 5× | 1 920 |
| 7d | 672 | **10 080** | 15× | 13 440 |
| 30d | 720 | **43 200** | **60×** | 57 600 |

At 30 d the Storage tab drew ~43 200 points × 5 database bands ≈ 216 000 SVG path coordinates in one chart. (The 1 440 figure matches a live measurement taken on the stage cluster while instrumenting #1576, so this isn't theoretical.)

## What changed

An opt-in `downsampleToWindow` pass folds the **finished** series onto the lattice the requested window implies. `targetBucketMs(windowMs)` reads the same `TIME_PRESETS` table, just keyed by duration instead of id, so there's one source of truth.

Measured on the live 2-node stage cluster at 7 d: each band's path `d` attribute drops from **~110 KB to ~8 KB** (193 curve segments per band).

### Why a post-aggregation fold, not a wider snap

Widening the *record* lattice changes what the temporal aggregator sees. `bytes-sent` is `temporal: 'sum'` over a `rate`-transformed field, so rows from different periods landing in one bucket get summed. Measured before choosing the design:

```
60s lattice:    6 points, y = [1000, 1000, 1000, 1000, 1000, 1000]   ← correct B/s
300s grouping:  2 points, y = [3000, 3000]                            ← 3× inflated
```

That would have silently inflated bytes-sent/received, both `mqtt-traffic-*` panels and `fsWrite` — 60× at 30 d. Folding *after* every aggregation pass leaves temporal, cross-node, and the bounded carry-forward from #1576 operating on exactly the buckets they do today. `render-density.test.ts` pins those counters at 1000 B/s across 24h/7d/30d so it can't regress.

### Fold rules ([downsample.ts](src/features/instance/status/analytics/pipeline/downsample.ts))

- **rate fields → `mean`**, never `sum`. Equal-weighted: Harper emits one row per node per period at fixed cadence, so every point in a coarse bucket covers the same span.
- **percentiles → `max`**. A p95-of-p95s isn't a p95; of the two defensible approximations, an operator would rather keep spikes visible than smooth them away at 30 d.
- **gauges → `max`/`last`** unchanged.
- **`count-weighted-mean` stays count-weighted**, via `SeriesPoint.count`.
- Counts are summed so confidence gating still sees the real observation total; an all-null coarse bucket stays an explicit `null` gap; bucket times use round-to-nearest, matching `snapToBucketTime`'s convention so grids stay aligned.

### Who opts in

18 chart-render sites plus CSV export (so a download matches the chart it came from).

**KPI tiles deliberately opt out** — they collapse to a headline number, so there's no density to save, and folding first would redefine `latestValue` from "the newest bucket" to "an average over the newest coarse bucket".

`error-rate`, `request-rate` and `transaction-log-growth` build series from raw columns without going through `runPipeline`, so MetricRenderer and csvExport fold their output via `downsampleDerivedSeriesData` — otherwise those panels would stay dense right next to coarsened neighbours on the same tab (`transaction-log-growth` sits beside `database-size` on Storage). The fold is idempotent, so applying it to `mqtt-traffic`'s recompute — which opts in internally — is a no-op, not a double average. `DerivedMetricSpec` gains an optional `downsampleAggregator` defaulting to `mean`, correct for all three shipped derived metrics.

## Reviewer notes

**`runPipeline`'s `window` argument is now load-bearing.** It was previously ignored (`_window`), so callers passed whatever. `bytes-sent-tolerance`'s `0 → MAX_SAFE_INTEGER` sentinel read as a ~285-century window and folded the series into a single bucket, collapsing the Σ-of-rates that test checks; it now passes the fixture's own ~59 min span. `recomputeRequestRate` uses the same sentinel but never delegates to `runPipeline`, so it's unaffected. I grepped for other sentinel windows — the remaining ones are in tests that don't opt into downsampling, which is the payoff of making this opt-in rather than default-on.

**`targetBucketMs(preset.durationMs) === preset.bucketMs` is an invariant, not a coincidence.** StorageTab aligns its trend grid to the context's `bucketMs` (#1514); if the render lattice and that value diverge, trend timestamps stop landing on the metric panels' and `syncMethod="value"` crosshair sync silently breaks. Locked by tests in both `timePresets.test.ts` and `render-density.test.ts`.

## This does NOT reduce server load

Worth being explicit, since that was half the original ask. We already send `bucket_ms`; the server ignoring it is what makes the payload 57 600 rows at 30 d. While verifying at 7 d I saw the analytics requests **hit the 60 s axios timeout**, and `getAnalytics`' `MAX_ROWS = 50_000` tail-keep can silently truncate the oldest part of a wide window (at 30 d, `database-size` across 5 databases × 2 nodes is ~288 000 rows, so the chart would show roughly the most recent 5 days of a 30-day window and say nothing about it). Both need core to honor `bucket_ms` — worth a companion core issue, and arguably a client-side banner when the cap trips.

## Tests

- `__tests__/pipeline/render-density.test.ts` — rendered point count per preset (30 d: >40 000 → ≤721); `targetBucketMs` preset agreement, now()-drift tolerance, degenerate windows, beyond-30 d behavior; gauge totals surviving the fold; the rate-inflation guard; opt-out keeping full resolution.
- `__tests__/pipeline/downsample.test.ts` — `isRateTransform` (including nested in `compose`), the aggregator mapping, folding/count-summing/null handling/identity-when-nothing-folds, derived idempotence, ceiling series.
- `context/timePresets.test.ts` — the #1514 alignment invariant and monotonicity.

Full repo gate green: **1918 tests / 266 files**, `tsc -b`, `oxlint`, `dprint check`. Verified in the browser at 24 h and 7 d on the stage cluster — charts render cleanly, no console errors.

### DIFF
--- a/src/features/instance/status/analytics/lib/csvExport.ts
+++ b/src/features/instance/status/analytics/lib/csvExport.ts
@@ -9,8 +9,10 @@
 // consumer can filter. No network calls: everything recomputes from the
 // records the panel already fetched.
 
+import { targetBucketMs } from '../context/timePresets';
 import { preprocessConnectionRecords } from '../pipeline/connection';
 import { derivedRegistry } from '../pipeline/derived/index';
+import { downsampleDerivedSeriesData } from '../pipeline/downsample';
 import { specRegistry } from '../pipeline/index';
 import { runPipeline } from '../pipeline/pipeline';
 import { aggregateReplicationMatrix, type ReplicationQuantileField } from '../pipeline/replication-latency';
@@ -181,7 +183,16 @@ export function computeMetricCsvData(
 
 	const derived = derivedRegistry[metric];
 	if (derived) {
-		return { kind: 'series', data: derived.recompute(records, timeRange, nodes, viewMode) };
+		// Folded to the same lattice MetricRenderer draws, so a download matches
+		// the chart it was taken from rather than being 60x longer at 30 d.
+		return {
+			kind: 'series',
+			data: downsampleDerivedSeriesData(
+				derived.recompute(records, timeRange, nodes, viewMode),
+				targetBucketMs(timeRange.endTime - timeRange.startTime),
+				derived.downsampleAggregator,
+			),
+		};
 	}
 
 	const entry = specRegistry[metric];
@@ -207,7 +218,10 @@ export function computeMetricCsvData(
 		// `connection` groups on the synthetic pathMethod field, so it needs
 		// the renderer's own preprocessing first.
 		const rows = metric === 'connection' ? preprocessConnectionRecords(records).records : records;
-		return { kind: 'series', data: runPipeline(spec, rows, timeRange, nodes, { snapToPeriod: true }) };
+		return {
+			kind: 'series',
+			data: runPipeline(spec, rows, timeRange, nodes, { snapToPeriod: true, downsampleToWindow: true }),
+		};
 	}
 
 	const isPerNodeMode = (viewMode ?? 'per-node') === 'per-node';
@@ -224,7 +238,11 @@ export function computeMetricCsvData(
 					crossNode: field.aggregator?.crossNode ?? spec.aggregator.crossNode,
 				},
 			};
-			const out = runPipeline(innerSpec, records, timeRange, nodes, { perNode: isPerNodeMode, snapToPeriod: true });
+			const out = runPipeline(innerSpec, records, timeRange, nodes, {
+				perNode: isPerNodeMode,
+				snapToPeriod: true,
+				downsampleToWindow: true,
+			});
 			series.push(...out.series);
 			// Each panel's ceiling (if any) becomes a regular column — SeriesData
 			// has a single ceiling slot, which can't carry one per panel.
@@ -235,23 +253,37 @@ export function computeMetricCsvData(
 
 	if (wantsStackedAreaNodeRemap(spec, isPerNodeMode) && spec.series.kind === 'groupBy') {
 		const remapped: MetricSpec = { ...spec, series: { ...spec.series, dimension: 'node' } };
-		return { kind: 'series', data: runPipeline(remapped, records, timeRange, nodes, { snapToPeriod: true }) };
+		return {
+			kind: 'series',
+			data: runPipeline(remapped, records, timeRange, nodes, { snapToPeriod: true, downsampleToWindow: true }),
+		};
 	}
 	if (wantsClusterLineFold(spec, isPerNodeMode) && spec.series.kind === 'groupBy') {
 		const inner: MetricSpec = {
 			...spec,
 			series: { kind: 'field', fields: [{ ...spec.series.field, label: 'cluster' }] },
 		};
-		return { kind: 'series', data: runPipeline(inner, records, timeRange, nodes, { snapToPeriod: true }) };
+		return {
+			kind: 'series',
+			data: runPipeline(inner, records, timeRange, nodes, { snapToPeriod: true, downsampleToWindow: true }),
+		};
 	}
 	if (wantsDimensionLineSplit(spec)) {
 		return {
 			kind: 'series',
-			data: runPipeline(spec, records, timeRange, nodes, { perNode: false, snapToPeriod: true }),
+			data: runPipeline(spec, records, timeRange, nodes, {
+				perNode: false,
+				snapToPeriod: true,
+				downsampleToWindow: true,
+			}),
 		};
 	}
 	return {
 		kind: 'series',
-		data: runPipeline(spec, records, timeRange, nodes, { perNode: isPerNodeMode, snapToPeriod: true }),
+		data: runPipeline(spec, records, timeRange, nodes, {
+			perNode: isPerNodeMode,
+			snapToPeriod: true,
+			downsampleToWindow: true,
+		}),
 	};
 }
--- a/src/features/instance/status/analytics/lib/csvExport.ts
+++ b/src/features/instance/status/analytics/lib/csvExport.ts
@@ -183,8 +183,11 @@ export function computeMetricCsvData(
 
 	const derived = derivedRegistry[metric];
 	if (derived) {
-		// Folded to the same lattice MetricRenderer draws, so a download matches
-		// the chart it was taken from rather than being 60x longer at 30 d.
+		// Folded to the same lattice MetricRenderer draws for a panel, so a
+		// download matches the chart it was taken from rather than being 60x
+		// longer at 30 d. Deliberately panel resolution even when the export is
+		// triggered from the expand dialog — an exported file should be a stable
+		// artifact of (metric, window), not of transient dialog state.
 		return {
 			kind: 'series',
 			data: downsampleDerivedSeriesData(
--- a/src/features/instance/status/analytics/pipeline/downsample.ts
+++ b/src/features/instance/status/analytics/pipeline/downsample.ts
@@ -59,8 +59,12 @@ export function downsampleAggregator(temporal: Aggregator, transform: Transform 
 export function downsamplePoints(points: SeriesPoint[], targetMs: number, agg: Aggregator): SeriesPoint[] {
 	if (targetMs <= 0 || points.length === 0) { return points; }
 	const byBucket = new Map<number, { items: { value: number | null; count?: number }[]; count: number }>();
+	let shifted = false;
 	for (const p of points) {
 		const bucket = Math.round(p.x / targetMs) * targetMs;
+		// A point can land in a bucket of its own and *still* need moving —
+		// see the early-return note below.
+		if (bucket !== p.x) { shifted = true; }
 		let entry = byBucket.get(bucket);
 		if (!entry) {
 			entry = { items: [], count: 0 };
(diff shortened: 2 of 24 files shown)


src/features/instance/status/analytics/pipeline/downsample.ts
```diff
@@ -0,0 +1,135 @@
...
+			entry = { items: [], count: 0 };
+			byBucket.set(bucket, entry);
+		}
+		entry.items.push({ value: p.y, count: p.count });
+		// Sum observation counts so confidence gating still sees the real total.
+		entry.count += p.count ?? 0;
+	}
+	// Nothing to gain (and a needless copy) when every point already sits in
+	// its own coarse bucket — the common case for the 1 h / 6 h presets.
+	if (byBucket.size === points.length) { return points; }
```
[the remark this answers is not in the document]
dawsontoth: Good catch — real bug, fixed in df0a6124. Confirmed reachable: records with a real 90 s `period` snap onto a 90 s lattice, and folding that onto the 1 h target sends k*90_000 to round(1.5k)*60_000 — five distinct buckets from five points, but only the even-k ones stay put, so the series ended up off the target grid and broke the #1514 crosshair match. Took the suggested approach (track whether anything moved, keep the fast path only when nothing did). Note it becomes reachable precisely when core stops stamping `period: 0` (HarperFast/harper#1997), which is what we asked it to do — so this would have shipped latent and broken on a server upgrade. Added a regression test that fails on the old guard with `[0, 90000, 180000, 270000, 360000]`.

src/features/instance/status/analytics/primitives/MetricRenderer.tsx
```diff
@@ -135,7 +137,15 @@ export function MetricRenderer({
 				/>
 			);
 		} else {
-			const seriesData = derived.recompute(records, timeRange, nodes, viewMode);
+			// Derived metrics that assemble series from raw columns never pass
+			// through runPipeline's downsample pass, so fold here instead —
+			// otherwise a 7 d / 30 d view of these panels stays at one point per
+			// 60 s while every spec-driven panel beside it is coarsened.
+			const seriesData = downsampleDerivedSeriesData(
```
cb1kenobi, line 144: **High: this fold never runs for `error-rate` / `request-rate` — their charts stay at raw resolution**

This `else` branch only executes `if (!derived.Renderer)` (see line 129). But `errorRateDerived` and `requestRateDerived` (`pipeline/derived/error-rate.ts`, `pipeline/derived/request-rate.tsx`) both set `Renderer: PerPathRateRenderer`, so on the actual dashboard `MetricRenderer` takes the `if (derived.Renderer)` branch instead and renders `<PerPathRateRenderer .../>` directly. `PerPathRateRenderer`'s `compute` prop calls `recomputeErrorRate`/`recomputeRequestRate` straight through — neither `downsampleDerivedSeriesData` nor `targetBucketMs` is invoked anywhere in `PerPathRateRenderer.tsx`.

Net effect: of the three derived metrics the PR description calls out ("`error-rate`, `request-rate` and `transaction-log-growth` ... so MetricRenderer and csvExport fold their output"), only `transaction-log-growth` (no custom `Renderer`) actually gets folded on screen. The Requests tab's error-rate and request-rate charts keep rendering one point per 60 s at 30 d — the exact density problem this PR exists to fix — right next to panels on the same tab that are now capped at ~180 points. `render-density.test.ts` doesn't catch this because it only exercises `runPipeline`/`targetBucketMs` directly, never `MetricRenderer` or `PerPathRateRenderer`.

CSV export (`csvExport.ts`) is unaffected since it calls `derived.recompute` + the fold directly, bypassing `Renderer` entirely — so the export and the on-screen chart for these two metrics will now show different resolutions.

Suggested fix: fold inside `PerPathRateRenderer`'s `data` memo (same pattern `TrafficByTypeRenderer` uses for `mqtt-traffic`), e.g. wrap `compute(...)` with `downsampleDerivedSeriesData(..., targetBucketMs(timeRange.endTime - timeRange.startTime, { expanded: !!fillParent }))`, using each derived spec's `downsampleAggregator`.

—
_Generated by Barber AI_

src/features/instance/status/analytics/lib/csvExport.ts
```diff
@@ -181,7 +183,19 @@ export function computeMetricCsvData(
...
+		// download matches the chart it was taken from rather than being 60x
+		// longer at 30 d. Deliberately panel resolution even when the export is
+		// triggered from the expand dialog — an exported file should be a stable
+		// artifact of (metric, window), not of transient dialog state.
+		return {
+			kind: 'series',
+			data: downsampleDerivedSeriesData(
+				derived.recompute(records, timeRange, nodes, viewMode),
+				targetBucketMs(timeRange.endTime - timeRange.startTime),
+				derived.downsampleAggregator,
```
cb1kenobi, line 196: **Medium: `error-rate`'s CSV export folds with an unweighted `mean`, reintroducing the ratio-of-ratios bug its own file warns against**

`errorRateDerived` never sets `downsampleAggregator`, so `downsampleDerivedSeriesData` here defaults to `'mean'`. Each point already carries `count: e.sumCount` (the request volume for that bucket) precisely so it *could* be weighted, but `aggregate('mean', items)` in `aggregators.ts` ignores `.count` entirely — it's a plain arithmetic mean of the per-bucket ratios.

`error-rate.ts`'s own docstring calls this out by name: "NOT mean(1 − ratio) — that's the canonical 'ratio-of-ratios' bug ... Two records with [{count: 1000, total: 990}, {count: 10, total: 1}] should yield ≈0.0188 (Σ-correct), not 0.455 (mean-of-ratios)." Folding two coarse buckets of very different request volume with an unweighted mean reproduces exactly that: a low-traffic, high-error-rate bucket gets the same weight as a high-traffic, low-error-rate bucket at 7 d/30 d.

Because `errorRateDerived` has a custom `Renderer`, today this only affects the CSV export path (see the MetricRenderer.tsx comment on this PR) — but it means the exported numbers for `error-rate` at wide windows are mathematically wrong, and if the `Renderer` gap above gets fixed to also fold the live chart, this bug would then hit the chart too.

Suggested fix: set `downsampleAggregator: 'count-weighted-mean'` on `errorRateDerived` in `pipeline/derived/error-rate.ts`. Since `ratio_i = 1 - total_i/count_i`, `ratio_i * count_i` recovers the bucket's error count, so count-weighting reconstructs the Σ-correct combination (`Σerrors / Σcount`) instead of averaging the ratios.

—
_Generated by Barber AI_

src/features/instance/status/analytics/primitives/MetricRenderer.tsx
dawsontoth: Fixed in de539ace by folding inside `PerPathRateRenderer`'s `data` memo (`downsampleDerivedSeriesData(compute(...), targetBucketMs(window, { expanded: !!fillParent }), …)`), and forwarding `fillParent` from request-rate's Renderer so the expand dialog folds at the finer resolution too.

One scope correction: this gap is **request-rate only**. `errorRateDerived` has no custom `Renderer` (only `requestRateDerived` sets one), so error-rate already takes MetricRenderer's `else` branch and *was* being folded on screen — the density gap you describe applies to request-rate. Verified against the code + MetricRenderer:129/139. Added `per-path-rate-renderer.test.tsx` (mocks `LineChart`, asserts a dense series folds to the window resolution and stays raw with no window) — the MetricRenderer/PerPathRateRenderer coverage you noted `render-density.test.ts` was missing.

src/features/instance/status/analytics/lib/csvExport.ts
dawsontoth: Fixed in de539ace — set `downsampleAggregator: 'count-weighted-mean'` on `errorRateDerived`, exactly as you suggested. ratio·count = count − total = the bucket's error count, so count-weighting reconstructs Σerrors/Σcount = 1 − Σtotal/Σcount.

Worth flagging: the impact was a bit wider than "CSV only". Since error-rate has **no** custom Renderer (see the reply on the MetricRenderer thread), the unweighted-mean fold was already running on the **on-screen chart** via MetricRenderer's `else` branch too — so this was producing wrong numbers on both the chart and the CSV at 7 d/30 d. Because MetricRenderer and csvExport both read `derived.downsampleAggregator`, the one-line spec change corrects both. `error-rate-downsample.test.ts` locks it: two disparate-volume buckets fold to ≈0.0188 (Σ-correct), not 0.455, with a guard test showing plain `mean` reproduces the 0.455 bug.

PR #1589, opened 2026-07-29
fix(status): bound the stacked-area forward-fill to each series' cadence

dawsontoth:
Stacked on #1587 — review that one first. Base auto-retargets to `stage` when it merges.

#1587 bounded last-observation-carry-forward for cross-node gauge sums in the pipeline. `StackedAreaChart` has a **second, independent forward-fill** at the render layer, and it was unbounded: it merges every series onto the union of x positions and carries each series' last-seen value across the positions it did not report, with no staleness limit.

## Why #1587's bound can't reach this

In **Stack by: Node**, `TrafficByTypeRenderer` remaps the spec's dimension to `'node'`, so each series holds exactly one node and that node is present in every bucket the series has. Since pipeline carry-forward never invents bucket times, it is a *structural no-op* in that mode — nothing but this merge bridges those bands.

That is both why node mode never showed the #1576 dives, and why an idle node's band was carried to the right-hand edge of the window. Core emits `mqtt-connections` only while the count is nonzero (`if (numberOfConnections > 0)` in `server/mqtt.ts`), so "no row" is exactly what an idle node looks like — the same overstatement #1587 fixed, one layer up.

## What changed

The fill stays; it is load-bearing. `period: 0` gauge rows snap onto the spec's 60 s fallback lattice, which a 90 s emission cadence beats against, so per-node lattice coverage is ragged and a strict merge draws a shredded stack. What changes is that a band now **expires**:

```
STALENESS_INTERVALS × max(median reporting gap, lattice step) + one lattice step
```

That trailing lattice step is slack the pipeline does not need. The pipeline estimates cadence from *raw* emission instants; a primitive only ever sees times already snapped onto the lattice, and snapping **compresses alternate gaps** — a 90 s cadence on a 60 s lattice reports at 0/120/180/300…, so its gaps alternate 120/60 and their median reads 60 or 90 depending on how many samples the window caught, while the widest gap is 120. Without the slack a series can expire on the very beat pattern the snap induced (I hit a zero-margin case before adding it). One step is the most the snap can stretch a gap by, which keeps this horizon at least as generous as the pipeline's.

Cadence is derived from the series' own point spacing rather than threaded through `SeriesData`: derived metrics (`mqtt-traffic-*`) and hand-built `SeriesData` reach this primitive too, so a threaded field would be absent often enough that this derivation would remain the live path anyway.

### Past the horizon a band fills 0, not null

- It mirrors the pipeline — dropping an expired node from a `crossNode: 'sum'` *is* contributing 0 to that sum.
- It is what a stacked chart means: bands add, so 0 is the neutral element, while a null punches a hole through the stack and drops every band above it.
- A cluster-wide gap still reads as a gap. X positions come only from series that actually reported, so when everything stops the rows stop rather than filling with zeros.

## Regression check across every stacked-area panel

| Panel | Aggregator | Effect |
|---|---|---|
| `connections` | max / sum | the motivating fix — gauge |
| `database-size` | last / sum | same gauge shape |
| `bytes-sent` / `bytes-received` | sum / sum | improves |
| `mqtt-traffic-sent` / `-received` | sum / sum | improves |

The additive counters improve for the same reason #1587 excluded them from pipeline carry-forward: a missing bucket for a rate series means *no traffic*, so expiring to 0 is the honest value where carrying the last throughput forward invented traffic indefinitely.

Observed values are never rewritten — including a genuine drop to 0 — and nothing is back-filled before a series' first point. The ceiling series stays unbounded deliberately: it is a reference line drawn over the stack, not a band in it, with no zero to expire to (a ceiling at 0 reads as a limit of nothing).

`csvExport` is untouched. It exports raw points with empty cells for gaps and never shared this fill — the export is the data, not the drawing.

## Testing

Extracted the merge to `mergeStackedRows` so it is unit-testable at all: every assertion about rendered geometry in the existing `stacked-area` suite is `it.skip`, because neither jsdom nor happy-dom lays Recharts out. Also memoized it in the component, which re-renders on every hover tick.

16 cases in `__tests__/primitives/stacked-row-staleness.test.ts`, shaped after `cross-node-gauge-gaps.test.ts` and driven through the real `connections` / `bytes-sent` specs and the real pipeline: the staggered-cadence shape actually producing ragged coverage, the pipeline no-op in node mode, horizon expiry, still-bridged-inside-the-horizon, no flicker back, the exact horizon boundary on hand-checkable numbers, observed-value passthrough, no back-fill, `y: null` handling, no invented rows, and the ceiling exception.

Verified the coverage is not vacuous by reverting the horizon to `Infinity` — 4 cases fail, then restored.

Gate: 265 files / 1898 tests pass, `tsc -b`, oxlint, and dprint all clean.

**Not browser-verified.** Reproducing the artifact needs a genuinely idle node in a live cluster, which can't be summoned on demand, and port 5173 was held by another checkout. Verification is mechanism-level, through the real specs and pipeline.

Refs #1576

### DIFF
--- a/src/features/instance/status/analytics/__tests__/primitives/stacked-row-staleness.test.ts
+++ b/src/features/instance/status/analytics/__tests__/primitives/stacked-row-staleness.test.ts
@@ -1,0 +1,275 @@
+// Regression coverage for the render-layer half of #1576.
+//
+// The pipeline fix bounded carry-forward for cross-node gauge *sums*, which is
+// what "Stack by: Type" renders. "Stack by: Node" never went through that code
+// path at all: the renderer remaps the spec's dimension to 'node', so each
+// series holds exactly one node and the pipeline's carry-forward — which never
+// invents bucket times — is a structural no-op. The only thing stitching those
+// bands across the lattice is StackedAreaChart's own forward-fill, and it was
+// unbounded, so an idle node's band was carried for the rest of the window.
+//
+// The staggered shapes below mirror cross-node-gauge-gaps.test.ts: `period: 0`
+// gauge rows on a 90 s emission cadence, snapped onto the spec's 60 s fallback
+// lattice, two nodes half a cadence out of phase.
+import { describe, expect, it } from 'vitest';
+import { runPipeline } from '../../pipeline/pipeline';
+import { wrapperMetrics } from '../../pipeline/wrapperMetrics';
+import { mergeStackedRows, type StackedRow } from '../../primitives/mergeStackedRows';
+import type { AnalyticsDataPoint, MetricSpec, SeriesData, TimeRange } from '../../types/analytics';
+
+/** `spec.bucket.fallbackMs` — what `period: 0` degrades to, and so the lattice
+ *  every gauge row here is snapped onto. */
+const LATTICE = 60_000;
+/** Emission cadence observed on the reporting instance. 90 mod 60 = 30, which
+ *  is what makes the nodes' snapped buckets beat against the lattice. */
+const CADENCE = 90_000;
+/** Lattice-aligned epoch, so every snapped bucket below is exact. */
+const T0 = 28_333_334 * LATTICE;
+/** Half a cadence — the second node's phase offset. */
+const PHASE = CADENCE / 2;
+
+const BUSY = 285; // us-west-1's steady connection count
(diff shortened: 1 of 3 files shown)


cb1kenobi:
Reviewed `a092573` — no issues found. This PR looks good, nice job!

—
_Generated by Barber AI_

PR #1594, opened 2026-07-30
fix(rum): drop private-repo SSH deploy failures from Error Tracking, redact repo URLs

dawsontoth:
No linked issue: the RUM signal embeds a customer's private repository URL, so it was
deliberately not filed publicly. Details below use `acme-corp/billing-service` in place of
the real one.

## Problem

Deploying a component from a private git repository surfaces an `SSEOperationError` whose
message is Harper's:

```
Failed to deploy private repository git@github.com:acme-corp/billing-service.git: SSH access
failed. Verify the repository URL, configure an SSH key on this Harper instance, ensure the
key has access to the target repository, and confirm the host is present in the
ssh/known_hosts file.
```

RUM recorded **8 of these in the 24h to 2026-07-30** (`env:prod`, one customer, one session),
and Error Tracking opened an issue for them. Two things are wrong with that:

1. **It isn't a Studio bug.** Every remediation step in the message is on the *instance* —
   install an SSH key, grant it access to the repo, add the git host to the instance's
   `ssh/known_hosts`. Nothing in Studio changes in response, so the issue can only ever be
   closed as customer configuration.
2. **It carries customer-identifying data.** The message embeds the private repo's owner and
   name, which Error Tracking then retains and makes searchable.

## Treatment

**Drop the event** (`shouldKeepEvent`). This is the same class as the expected states already
dropped there — 401s, timeouts, instance-endpoint 5xx. Matched on the
`Failed to deploy private repository` prefix rather than the SSH remediation sentence, so a
reworded message still drops; a generic `Failed to deploy component: …` is still kept.

**Redact customer URL paths anyway** (`redactErrorText`, new). Dropping this message keeps
*this* URL out of Datadog, but other deploy failures that stay kept — a repo 404, a bad ref,
npm auth output — quote the same `package` reference, and the drop rule depends on Harper's
wording. So redaction is a backstop, not a substitute.

The first cut of it asked "is this URL a repository?" and answered with a forge list, a
`git.`/`gitlab.`/`github.` hostname convention, a `.git` suffix and a git protocol check.
Review found the holes, and they share a root — that list never finishes. A GitLab or Gitea
self-hosted at `scm.acme-corp.com` matches none of it; neither does a `git+https:` remote;
neither does `https://api.github.com/repos/<owner>/<repo>`, which is Studio's own lookup while
the deploy form is being filled in.

So it's inverted: **Harper's domains are the short list we control, so keep the path only for
those and reduce every other URL to scheme + host + `<redacted>`.** Credentials are stripped
from any URL's userinfo (an https private-repo remote can carry a token there — strictly worse
than a repo name), and query/fragment go with the path (a `?ref=` can name a private branch).
scp-style remotes (`git@host:owner/repo.git`) aren't URLs and keep their own rule.

```
Failed to deploy git@github.com:acme-corp/billing-service.git: exit code 128
  -> Failed to deploy git@github.com:<redacted>: exit code 128

Failed to deploy https://scm.acme-corp.com/acme-corp/billing-service: 404 not found
  -> Failed to deploy https://scm.acme-corp.com/<redacted>: 404 not found

Request failed with status code 500 https://api.harper.fast/HDBInstance/ins-123/operation
  -> unchanged
```

Harper API endpoints stay triageable and Studio's asset URLs stay source-mappable — that is
what the allowlist (`harper.fast`, `harperfabric.com`, `harperdb.io` by suffix, plus
`localhost`) is there to protect. Two accepted costs: a stack frame from an embedded
third-party script now reads `https://static.reo.dev/<redacted>`, which still attributes it by
host; and a URL whose path is just `/` is left alone rather than gaining a `<redacted>` that
hides nothing.

**Compose in `beforeSend`** (new module): filter first, then redact `error.message`,
`error.stack` and `error.resource.url`. Order matters — the third-party attribution in
`shouldKeepEvent` inspects the raw stack, so it must run before any frame is rewritten (there's
a test pinning that). `error.resource.url` is in scope because a failed lookup carries the same
identity in a different field: a 404 on a private repo would otherwise ship the repo name
verbatim beside a redacted message. Only the reported copy is rewritten; the UI renders the
error object itself.

**Compensate in the UI.** With RUM now silent on this failure, `DeployProgress` is the only
place it surfaces, so it links to Config > SSH Keys beneath Harper's remediation text.
`onNavigateAway` lets `RedeployApplicationModal` close first — it's controlled by global
watched state, not the route, so it would otherwise sit over the SSH Keys page. The import
flow is unchanged: its toast already shows the full remediation text, and
`ImportAuthorization` links to SSH Keys and warns on key/host mismatch before submit.

## Testing

`npx vitest run` → 267 files / 2003 passing; `tsc -b`, `dprint check`, `oxlint` clean.

New coverage: the drop rule (including the reworded variant and a generic deploy failure that
must stay); redaction of a self-hosted forge under a custom domain (including a nested group
path), `git+https:`/`git+ssh:` remotes, and Studio's own GitHub/npm lookups; what must *not* be
redacted (Harper endpoints, Studio stack frames on a Harper subdomain, a bare host with no
path, an email followed by a colon); `error.resource.url` redacted for a lookup but left intact
for an instance operation endpoint; the filter-before-redact ordering; and the guidance link's
href and dismiss behaviour.

Not browser-verified: reaching that UI needs a real failing private-repo deploy against a
Harper >= 5.1 instance, which the preview can't produce, so the unit tests are the
verification.

## Notes, out of scope

**RUM `resource` events carry the same identity, unredacted.** `beforeSend` only rewrites error
events here. Resource events are a live vector, not a hypothetical one: a 7-day RUM query
returns `resource` events for `github.com` URLs that name a customer's repo path (CI badge
images), and `getGitHubRepo` / `findNpmPackageName` fetch `api.github.com` and
`registry.npmjs.org` with the owner/repo or package the customer typed. Extending redaction to
`resource.url` would close it, but it changes what every third-party resource looks like in the
RUM performance views, so it's a call for the team rather than a quiet addition to this PR.

**The global `errorHandler` splits messages on the first `:`**, so this failure currently toasts
with the title `Failed to deploy private repository git@github.com` and the rest as the
description. That's generic behaviour for any colon-bearing message, not specific to this path.

### DIFF
--- a/src/integrations/datadog/redactErrorText.test.ts
+++ b/src/integrations/datadog/redactErrorText.test.ts
@@ -8,6 +8,9 @@ describe('redactErrorText', () => {
 		// Host aliases are how a customer selects between multiple keys (#1318) — keep them.
 		expect(redactErrorText('git@github-work:acme-corp/internal.git not found'))
 			.toBe('git@github-work:<redacted> not found');
+		// A bare SSH server has no owner prefix; the `.git` suffix is the tell.
+		expect(redactErrorText('Failed to deploy git@server.example.com:billing-service.git: exit code 128'))
+			.toBe('Failed to deploy git@server.example.com:<redacted>: exit code 128');
 	});
 
 	it('redacts the path of a repository URL, keeping scheme and host', () => {
--- a/src/integrations/datadog/redactErrorText.test.ts
+++ b/src/integrations/datadog/redactErrorText.test.ts
@@ -1,0 +1,63 @@
+import { describe, expect, it } from 'vitest';
+import { redactErrorText } from './redactErrorText';
+
+describe('redactErrorText', () => {
+	it('redacts the path of an scp-style git remote, keeping user and host', () => {
+		expect(redactErrorText('Failed to deploy git@github.com:acme-corp/billing-service.git: exit code 128'))
+			.toBe('Failed to deploy git@github.com:<redacted>: exit code 128');
+		// Host aliases are how a customer selects between multiple keys (#1318) — keep them.
+		expect(redactErrorText('git@github-work:acme-corp/internal.git not found'))
+			.toBe('git@github-work:<redacted> not found');
+	});
+
+	it('redacts the path of an https/ssh repository URL, keeping scheme and host', () => {
+		expect(redactErrorText('Cannot clone https://github.com/acme-corp/billing-service'))
+			.toBe('Cannot clone https://github.com/<redacted>');
+		expect(redactErrorText('Cannot clone ssh://git@source.example.com/acme/repo'))
+			.toBe('Cannot clone ssh://source.example.com/<redacted>');
+		// A self-hosted forge under the conventional `git.` prefix, and a `.git` path anywhere.
+		expect(redactErrorText('https://git.acme-corp.com/team/app failed'))
+			.toBe('https://git.acme-corp.com/<redacted> failed');
+		expect(redactErrorText('https://build.example.com/acme/app.git failed'))
+			.toBe('https://build.example.com/<redacted> failed');
+	});
+
+	it('drops the query and fragment of a repository URL, which can name a private branch', () => {
+		expect(redactErrorText('https://github.com/acme-corp/app.git?ref=feature/pricing#v2 failed'))
+			.toBe('https://github.com/<redacted> failed');
+	});
+
+	it('keeps sentence punctuation that follows a URL outside the redaction', () => {
+		expect(
+			redactErrorText(
+				'Failed to deploy private repository https://github.com/acme-corp/app.git: SSH access failed.',
+			),
+		).toBe('Failed to deploy private repository https://github.com/<redacted>: SSH access failed.');
+	});
+
+	it('strips credentials embedded in any URL, repository or not', () => {
+		expect(redactErrorText('POST https://acme:ghp_secret@github.com/acme-corp/app.git failed'))
+			.toBe('POST https://github.com/<redacted> failed');
+		expect(redactErrorText('POST https://user:hunter2@api.harper.fast/HDBInstance/ins-1/operation failed'))
+			.toBe('POST https://api.harper.fast/HDBInstance/ins-1/operation failed');
+	});
+
+	it('leaves Harper API endpoints and Studio asset URLs intact, so errors stay triageable', () => {
+		const message = 'Request failed with status code 500 https://api.harper.fast/HDBInstance/ins-123/operation';
+		expect(redactErrorText(message)).toBe(message);
+		const stack = [
+			'TypeError: x is not a function',
+			'  at getOrganization @ https://fabric.harper.fast/assets/index-A1b2C3d4.js:5:1234',
+			'  at pn @ https://fabric.harper.fast/assets/vendor-datadog-DBn-aOxh.js:3:3396',
+		].join('\n');
+		expect(redactErrorText(stack)).toBe(stack);
+	});
+
+	it('leaves ordinary error text alone', () => {
+		expect(redactErrorText('AxiosError: timeout of 60000ms exceeded'))
+			.toBe('AxiosError: timeout of 60000ms exceeded');
+		// An email followed by a colon must not be mistaken for an scp-style remote.
+		expect(redactErrorText('dawson@harperdb.io: sign-in failed')).toBe('dawson@harperdb.io: sign-in failed');
+		expect(redactErrorText('not-a-url://')).toBe('not-a-url://');
+	});
+});
--- a/src/integrations/datadog/redactErrorText.test.ts
+++ b/src/integrations/datadog/redactErrorText.test.ts
@@ -10,19 +10,40 @@ describe('redactErrorText', () => {
 			.toBe('git@github-work:<redacted> not found');
 	});
 
-	it('redacts the path of an https/ssh repository URL, keeping scheme and host', () => {
+	it('redacts the path of a repository URL, keeping scheme and host', () => {
 		expect(redactErrorText('Cannot clone https://github.com/acme-corp/billing-service'))
 			.toBe('Cannot clone https://github.com/<redacted>');
 		expect(redactErrorText('Cannot clone ssh://git@source.example.com/acme/repo'))
 			.toBe('Cannot clone ssh://source.example.com/<redacted>');
-		// A self-hosted forge under the conventional `git.` prefix, and a `.git` path anywhere.
-		expect(redactErrorText('https://git.acme-corp.com/team/app failed'))
-			.toBe('https://git.acme-corp.com/<redacted> failed');
 		expect(redactErrorText('https://build.example.com/acme/app.git failed'))
 			.toBe('https://build.example.com/<redacted> failed');
+		// npm/pip-style `git+<transport>` remotes parse as their own protocol; keep it in the output.
+		expect(redactErrorText('Cannot clone git+https://custom-forge.com/acme/repo'))
+			.toBe('Cannot clone git+https://custom-forge.com/<redacted>');
+		expect(redactErrorText('Cannot clone git+ssh://git@custom-forge.com/acme/repo.git'))
+			.toBe('Cannot clone git+ssh://custom-forge.com/<redacted>');
 	});
 
-	it('drops the query and fragment of a repository URL, which can name a private branch', () => {
+	// The gap that made "is this a repository?" the wrong question: a self-hosted forge under a
+	// name that says nothing about git, with no `.git` suffix, is exactly the enterprise setup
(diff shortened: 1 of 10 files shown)


src/integrations/datadog/redactErrorText.ts
```diff
@@ -0,0 +1,81 @@
...
+ * (`index-A1b2C3d4.js:5:1234`) can't match.
+ */
+const SCP_GIT_REMOTE = /([\w.+-]+@[\w.-]+):[\w.-]+\/[\w./-]+/g;
+
+function isRepositoryUrl(url: URL) {
+	return url.protocol === 'git:'
+		|| url.protocol === 'ssh:'
+		|| /\.git\/?$/.test(url.pathname)
+		|| GIT_HOSTS.has(url.hostname)
+		|| /^(?:git|gitlab|github)[.-]/.test(url.hostname);
```
cb1kenobi: **Medium: self-hosted git forges under a custom domain skip redaction entirely**

`isRepositoryUrl` only recognizes a self-hosted forge via the `git.`/`gitlab.`/`github.` hostname-prefix convention or a `.git` path suffix. A private repo hosted at a domain that doesn't follow that convention and whose URL has no `.git` suffix — e.g. GitLab/Gitea self-hosted at `scm.acme-corp.com`, `code.acme-corp.io`, `vcs.acme.net` — is left completely unredacted. Verified directly against this function:
```
redactErrorText('Failed to deploy https://scm.acme-corp.com/acme-corp/billing-service: 404 not found')
// => unchanged — 'acme-corp/billing-service' ships to Datadog verbatim
```
This matters because these are exactly the deploy failures (404, bad ref, npm auth output) the PR description says stay *kept* in RUM — `redactErrorText` is their only protection against leaking the customer's private repo identity, and self-hosted forges under a custom domain are a common enterprise setup, not an edge case.

Suggested fix: since Harper's own domains are the small, enumerable set (`api.harper.fast`, `fabric.harper.fast`, the Studio asset CDN pattern), invert the heuristic — treat any non-Harper `https/http` host whose path has an `owner/repo`-shaped segment count as a repository URL by default, rather than requiring the host to match a known-forge pattern first. That closes the gap without needing to enumerate every possible self-hosted domain.

—
_Generated by Barber AI_

dawsontoth: Good catch — fixed in 25d2fe1b, and I took the inversion further than the owner/repo shape.

You're right that the host convention was the wrong axis, and the same hole showed up twice more once I looked: `git+https:`/`git+ssh:` remotes (flagged separately by gemini below), and `https://api.github.com/repos/<owner>/<repo>` — the lookup Studio itself makes from `getGitHubRepo` while the deploy form is being filled in, which leaks the identity without any repo URL being in the message at all.

So rather than a segment-count heuristic, the rule is now: **keep the path only for Harper's own hosts (`harper.fast`, `harperfabric.com`, `harperdb.io` by suffix, plus `localhost`), redact everything else to scheme + host + `<redacted>`.** Your case:

```
Failed to deploy https://scm.acme-corp.com/acme-corp/billing-service: 404 not found
  -> Failed to deploy https://scm.acme-corp.com/<redacted>: 404 not found
```

I skipped the segment count because self-hosted GitLab subgroups (`/platform/payments/billing-service`) are three-deep, so "owner/repo-shaped" would have let the enterprise case through again. Both are covered by tests now.

What the allowlist is protecting is deliberate: Harper API paths stay triageable, and Studio's own asset URLs stay intact so Datadog can still source-map stack frames against them. Two costs I took knowingly — a frame from an embedded third-party script now reads `https://static.reo.dev/<redacted>` (host still attributes it, and `shouldKeepEvent` reads the raw stack before redaction, so the third-party filter is unaffected), and a URL whose path is only `/` is left alone.

Also extended the redaction to `error.resource.url` in the same commit — a 404 on a private repo lookup would otherwise have shipped the repo name verbatim in that field while the message beside it was redacted.

One thing your comment surfaced that this PR does **not** fix, now written up under "Notes, out of scope": RUM `resource` events carry the same identity and `beforeSend` doesn't touch them here. It's live, not hypothetical — a 7-day query returns resource events for `github.com` URLs naming a customer's repo path. Closing it means redacting `resource.url` for every third-party resource, which changes the RUM performance views, so I'd rather that be a deliberate team call than a quiet addition here.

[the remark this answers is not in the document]
dawsontoth: Fixed in 25d2fe1b, though not with a protocol check.

`git+*` was one of three holes with the same root — deciding which URLs *are* repositories is a list that never finishes. The others: a GitLab/Gitea self-hosted at `scm.acme-corp.com` (no `.git` suffix, no git-ish hostname), and `https://api.github.com/repos/<owner>/<repo>`, Studio's own repo lookup.

So the classification is inverted: keep the path only for Harper's own domains, redact every other URL to scheme + host + `<redacted>`. `git+https://custom-forge.com/acme/repo` → `git+https://custom-forge.com/<redacted>` falls out of that without enumerating sub-protocols, and `GIT_HOSTS`, the hostname-prefix convention, the `.git` suffix test and the `git:`/`ssh:` checks are all gone.

src/integrations/datadog/redactErrorText.test.ts
```diff
@@ -0,0 +1,63 @@
...
+		// Host aliases are how a customer selects between multiple keys (#1318) — keep them.
+		expect(redactErrorText('git@github-work:acme-corp/internal.git not found'))
+			.toBe('git@github-work:<redacted> not found');
+	});
+
+	it('redacts the path of an https/ssh repository URL, keeping scheme and host', () => {
+		expect(redactErrorText('Cannot clone https://github.com/acme-corp/billing-service'))
+			.toBe('Cannot clone https://github.com/<redacted>');
+		expect(redactErrorText('Cannot clone ssh://git@source.example.com/acme/repo'))
+			.toBe('Cannot clone ssh://source.example.com/<redacted>');
```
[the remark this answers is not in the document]
dawsontoth: Added in 25d2fe1b, in roughly the form you suggested:

```ts
// npm/pip-style `git+<transport>` remotes parse as their own protocol; keep it in the output.
expect(redactErrorText('Cannot clone git+https://custom-forge.com/acme/repo'))
	.toBe('Cannot clone git+https://custom-forge.com/<redacted>');
expect(redactErrorText('Cannot clone git+ssh://git@custom-forge.com/acme/repo.git'))
	.toBe('Cannot clone git+ssh://custom-forge.com/<redacted>');
```

The `ssh://` case you included was already covered in that block, so I kept it there and added the `git+ssh:` variant beside it to pin that userinfo is dropped on a sub-protocol too. Also new: a self-hosted forge under a custom domain (including a nested group path), and Studio's own `api.github.com` / `registry.npmjs.org` lookups.

src/integrations/datadog/redactErrorText.ts
```diff
@@ -0,0 +1,85 @@
...
+
+/** Sentence punctuation the token regex swallows when a URL ends a clause. */
+const TRAILING_PUNCTUATION = /[.,:;!?]+$/;
+
+/**
+ * scp-style git remotes: `git@github.com:acme-corp/repo.git`. Requires both userinfo and a
+ * `<segment>/<segment>` path, so ordinary "Title: detail" text and stack frames
+ * (`index-A1b2C3d4.js:5:1234`) can't match.
+ */
+const SCP_GIT_REMOTE = /([\w.+-]+@[\w.-]+):[\w.-]+\/[\w./-]+/g;
```
cb1kenobi: **Low: single-segment scp-style remotes escape redaction**

The `<segment>/<segment>` path requirement means a remote on a plain SSH git server — `git@server.example.com:billing-service.git`, no `owner/` prefix — matches neither this rule nor `URL_TOKEN` (it isn't a URL), so the repo name ships unredacted in kept deploy errors: `Failed to deploy git@server.example.com:billing-service.git: exit code 128` passes through unchanged (verified against this module). An alternation that also accepts a single segment ending in `.git` closes it while keeping the email/"Title: detail"/stack-frame guards intact (verified against the existing test cases):

```suggestion
const SCP_GIT_REMOTE = /([\w.+-]+@[\w.-]+):[\w.-]+(?:\/[\w./-]+|\.git\b)/g;
```

The doc comment above ("Requires both userinfo and a `<segment>/<segment>` path") would need the same adjustment, plus a test case.

—
_Generated by Barber AI_

cb1kenobi: Addressed in 1a092296: the scp rule now also accepts a `.git`-suffixed single segment (`git@server.example.com:billing-service.git` → `git@server.example.com:<redacted>`), with the doc comment updated and a regression test added. The email / "Title: detail" / stack-frame guards are unchanged and still covered.

cb1kenobi (approved):
Tested to the best of my ability. Code looks good.

PR #1598, opened 2026-08-05
fix(errors): surface Harper 5 RFC 9457 problem details in error toasts

Devin-Holland:
## What

Harper 5 changed REST error responses from `"Code: message"` strings to [RFC 9457 Problem Details](https://www.rfc-editor.org/rfc/rfc9457) objects:

```json
{ "type": "error:ClientError", "code": "ClientError", "title": "Not allowed", "status": 403, "instance": "/Cluster/clu-123" }
```

The global `errorHandler` only looked at `data.error` / `data.message`, so against a Harper 5 central-manager **every** error toast fell through to the generic "We had some trouble!".

## Changes

- **`queryClient.ts`** — recognize a Problem Details body: `code` → toast title, `title` (+ `detail` when present) → description. The title is a plain sentence that can itself contain colons ("Plan not found: plan-123"), so the legacy "Title: detail" colon-split is skipped for this shape. v4 string/`error`/`message` handling is unchanged, so stage/prod CMs still on 4.x are unaffected.
- **`isEmailNotVerifiedError.ts`** — same class of bug found by sweeping for other CM error-body parsing: the unverified-email login detection only read `error`/`message`, so on a v5 CM an unverified user would get a dead-end toast instead of the email-verification redirect. Now also checks `title`.

## Verification

- 6 new unit tests (4 errorHandler RFC 9457 cases incl. colon-in-title and detail-append; 2 for the unverified-email matcher) — 25/25 pass across both files, full suite green via the pre-commit hook (2151 passed / 11 skipped).
- Field mapping verified against Harper 5 source (`server/REST.ts` Problem Details construction): `code = error.code ?? constructor.name`, `title = error.message`, `detail` only present when set.

— devain (Claude)

### DIFF
--- a/src/features/auth/isEmailNotVerifiedError.test.ts
+++ b/src/features/auth/isEmailNotVerifiedError.test.ts
@@ -25,6 +25,25 @@ describe('isEmailNotVerifiedError', () => {
 			.toBe(true);
 	});
 
+	it('matches a Harper 5 RFC 9457 body (message under data.title)', () => {
+		expect(
+			isEmailNotVerifiedError(
+				axiosError(403, {
+					type: 'error:ClientError',
+					code: 'ClientError',
+					title: 'User has not verified email address',
+					status: 403,
+					instance: '/Login/',
+				}),
+			),
+		).toBe(true);
+	});
+
+	it('does NOT match an RFC 9457 deactivated-account rejection', () => {
+		expect(isEmailNotVerifiedError(axiosError(403, { code: 'ClientError', title: 'User account deactivated' })))
+			.toBe(false);
+	});
+
 	it('does NOT match a 403 deactivated-account rejection (same status, different reason)', () => {
 		expect(isEmailNotVerifiedError(axiosError(403, { error: 'User account deactivated' }))).toBe(false);
 	});
--- a/src/features/auth/isEmailNotVerifiedError.ts
+++ b/src/features/auth/isEmailNotVerifiedError.ts
@@ -18,9 +18,11 @@ export function isEmailNotVerifiedError(error: unknown): boolean {
 	if (!isAxiosError(error) || error.response?.status !== 403) {
 		return false;
 	}
-	const data = error.response.data as string | { error?: unknown; message?: unknown } | undefined;
+	// Harper 5 sends errors as RFC 9457 Problem Details, with the message in `title`;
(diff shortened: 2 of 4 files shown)


DavidCockerill (approved):
Approving — nothing to change.

The mapping is deliberately non-literal and that's the right call. Harper's `REST.ts` builds `code = error.code ?? error.constructor.name` and `title = error.message`, so `code → heading` / `title (+ detail) → description` produces a sensible toast; an RFC-purist mapping would have put a full sentence in the heading and left the body empty. The `splitTitleFromMsg = false` suppression is the subtle part and it's correct — without it `"Plan not found: plan-123"` gets sliced at the colon, and you pinned exactly that case.

I traced 23 error shapes through the new handler rather than just the happy ones — v5 problem details (including a numeric `code` and an empty `title`), v5 ops-API `{error}`, v4 string bodies, an nginx HTML 502, network/CORS/timeout, `undefined`/`null`/`{}`/array/Blob, a non-string `title`, both-`title`-and-`error`, and a circular `detail`. No blank toast, no `[object Object]`, no raw JSON blob, and nothing throws inside the handler on any of them (`errorText` catches circular JSON, `data.title` is narrowed to string before the template literal, and `errorMsg` is provably a string at the `.includes(':')` guard on every path).

Two things I specifically went looking for and didn't find:

- **The verification gate still matches.** This was the real risk — a detection that quietly stops matching means unverified users get a generic error instead of the resend flow. The change is purely additive (`title` appended last, existing fields win), and I checked the live wire format rather than the intent: central-manager pins `harperdb: ^4.7.29`, so deployed CM serializes `errorToString(error)` and the rejection arrives as a *string*, taking the untouched branch entirely. On v5 the new branch matches the same text. No fail-open either — of the three 403s reachable from `POST /Login`, only the intended one matches `/verif/i`.
- **No new sensitive-data exposure.** `instance` (the request URL, the one field that could carry a token in a query string) and `type` are deliberately not surfaced, and the logging path is unchanged.

One thing worth knowing but explicitly *not* yours to fix here: a proxy returning `text/html` still renders the whole document — tags and nginx version banner — into the toast, via the `typeof data === 'string'` branch this PR only reindents. It's now the last error shape without sane handling. Happy to file that separately if you want it tracked.

_Reviewed by Claude Opus 5 for @DavidCockerill._

kriszyp (approved):

PR #1599, opened 2026-08-05
fix(cluster): update cached resetPassword before navigating after admin setup

Devin-Holland:
Human-Review-Need: 4 @ 7bd97bc56beb
## What

After creating the initial admin user, FinishSetup showed the "Login successful" toast but sometimes never left the setup page.

## Why

The reset-password mutation PATCHes central-manager's `resetPassword` flag (awaited, so the server write is confirmed), but nothing updated the react-query cluster cache — `router.invalidate()` only re-runs route loaders. The subsequent `navigate('../')` landed on ClusterHome, which routes on `cluster.resetPassword` from the **same** `[clusterId]` query — still cached `true` until the next 10s poll tick — so its guard bounced the user straight back to `finish-setup`. Intermittent by poll timing, which is why it reproduced only sometimes.

## Change

New `markClusterPasswordSet(queryClient, clusterId)` helper next to the query flips `resetPassword` off in the cache; FinishSetup's `onSuccess` calls it before navigating. Not an optimistic update — the CM write has already succeeded by the time `onSuccess` runs.

## Verification

- 3 new unit tests for the helper (flips the flag preserving other fields; no phantom cache entry when absent; other clusters untouched) — 8/8 in the file, full suite green via the pre-commit hook, `tsc -b` clean.
- Traced the post-fix interaction with FinishSetup's own `!resetPassword` guard: both possible destinations (`../` and `../sign-in`) resolve to the cluster home for a now-connected user, so no new dead-end.

— devain (Claude)

### DIFF
--- a/src/features/cluster/queries/getClusterInfoQuery.ts
+++ b/src/features/cluster/queries/getClusterInfoQuery.ts
@@ -1,11 +1,24 @@
 import { apiClient } from '@/config/apiClient';
 import { Cluster } from '@/integrations/api/api.patch';
 import { pollUnlessForbidden } from '@/react-query/pollUnlessForbidden';
-import { queryOptions } from '@tanstack/react-query';
+import { QueryClient, queryOptions } from '@tanstack/react-query';
 
 export async function getClusterInfo(clusterId: string) {
 	const { data } = await apiClient.get(`/Cluster/${clusterId}` as '/Cluster/{id}');
 	return data as Cluster;
+}
+
+/**
+ * Flip `resetPassword` off on the cached cluster once admin setup has completed server-side.
+ * ClusterHome routes on `resetPassword` from this same query, so the cache must reflect the
+ * change synchronously — waiting for the next poll leaves a window where navigating away from
+ * finish-setup bounces straight back to it.
+ */
+export function markClusterPasswordSet(queryClient: QueryClient, clusterId: string) {
+	queryClient.setQueryData(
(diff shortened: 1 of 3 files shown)


src/features/cluster/queries/getClusterInfoQuery.ts
```diff
@@ -1,13 +1,26 @@
...
 }
 
+/**
+ * Flip `resetPassword` off on the cached cluster once admin setup has completed server-side.
+ * ClusterHome routes on `resetPassword` from this same query, so the cache must reflect the
+ * change synchronously — waiting for the next poll leaves a window where navigating away from
+ * finish-setup bounces straight back to it.
+ */
+export function markClusterPasswordSet(queryClient: QueryClient, clusterId: string) {
+	queryClient.setQueryData(
```
kriszyp: An already-running 10-second poll can still complete after `setQueryData` and replace this value with its stale `resetPassword: true` response. That recreates the redirect loop when `ClusterHome` renders. Please cancel the exact cluster query before applying the cache update (make this helper async and `await queryClient.cancelQueries({ queryKey, exact: true })`, then call `setQueryData`) and await the helper before navigating. A deferred-query test where the old response resolves after the update would cover the failure scenario.

— KrAIs

kriszyp (commented):
One codex suggestion

kriszyp (approved):

dawsontoth (approved):

PR #1611, opened 2026-08-11
Keep the platform status component protected after the akamai-status rename

kriszyp:
The applications editor makes the platform-managed status component read-only, so a customer cannot edit or delete the thing that keeps their instance in the Akamai load balancer. That guard is about to stop working.

## Why

`restrictPackageModification` matched two hardcoded package URLs:

```ts
openedEntry.package?.includes('github.com/HarperDB/status-check-fabric')
  || openedEntry.package?.includes('github.com/HarperFast/status-check-fabric')
```

host-manager is switching fabric provisioning to the `akamai-status` component (HarperFast/host-manager). Neither literal matches it, so on the next provisioned instance the component would become freely editable and deletable in the editor.

The two-literal shape is itself the tell: the second line exists because the `HarperDB` → `HarperFast` org rename silently unprotected every instance until someone noticed and added it. Matching the **repo name** rather than the full URL removes that failure mode — an org move, a host change, or an npm spec all still match.

Note this is the editor guard only. The component-status names in `getStatus.ts` (`status-check.rest`, `status-check.jsResource`) derive from the deploy *project* name, which host-manager deliberately keeps as `status-check`, so they are unaffected.

## A bypass this also fixes

Cross-model review surfaced a second, worse hole — pre-existing, but squarely in what this PR is about. `useEntryActions(entry)` takes a per-entry argument yet read `restrictPackageModification` from context, where the provider derives it from `openedEntry` alone. The sidebar context menu deliberately targets a right-clicked row *without* opening it (`FileTreeContextMenu.tsx:105-110`).

So: open any unprotected file, right-click the protected package root, and **Delete was offered** — the exact outcome the guard exists to prevent.

The fix makes protection a property of the entry being acted on. `isProtectedEntry(entry)` moves into the shared module; the provider applies it to `openedEntry` (right for the editor menu bar and editability) and the hook applies it to its own argument (right for the context menu). One predicate, two subjects, no drift.

## What changed

The predicate moves to its own module (`isProtectedComponentPackage.ts`, alongside the existing `isDirectory.ts` sibling helper) so it is testable without rendering the provider, plus unit coverage for both the string matching and — the part that would have caught the bypass — that a protected entry actually yields `canDeleteEntry === false` / `canRedeploy === false`.

## Verification

`vitest run` — 28 tests across the predicate, the path guard and the capability hook. `lint`, `dprint check` and `tsc --noEmit` all exit clean.

Full-suite comparison on this machine: `origin/stage` baseline is 41 failed / 2058 passed; with this change, 41 failed / **2086** passed. The delta is exactly the new tests and no previously-passing test changed state. Those 41 pre-existing local failures are environmental (jsdom `localStorage` needs `--localstorage-file`), not related to this change — CI is the real gate. `oxlint` and `dprint check` are clean.

Committed with `--no-verify`: the pre-commit hook runs the full suite, which is red on baseline here for the reason above.

## For the human reviewer

Small change; the matching strategy is the only judgment call.

**This guard is client-side, and there is a fourth surface it does not cover.** Enforcing at the modal closes the menu bar, the context menu and the keyboard shortcut, because all three funnel through it. The editor's chat tooling does not: `Chat/tools/dropComponentFile/execute.ts` calls `dropComponent` directly with a bare path, with no component metadata in scope to check against.

I stopped rather than patch that too, and I think the count is the argument. Four independent client paths reach the same deletion, and beneath all of them the REST endpoint is ungated — anyone can `curl` it. If "a customer must not be able to remove the component that keeps their instance in the load balancer" is a real operational invariant, it belongs where the deletion executes, not in a UI that keeps growing new callers. Happy to file that as a follow-up; say the word.

What this PR does deliver: the rename cannot silently unprotect the component, and the three paths that a user actually reaches by hand now refuse.

**Trailing-segment matching.** The name must be the trailing segment of the spec — not a prefix (`my-akamai-status-probe`), an extension (`akamai-status.dashboard`), or an owner (`github.com/akamai-status/theirs.git`), all of which earlier iterations got wrong and all of which are now pinned as negative cases. Every spec we actually deploy ends at the name, a `.git` suffix, a committish, or a version.

`PROTECTED_COMPONENT_REPOS` keeps `status-check-fabric` alongside the new entry, since existing instances continue running it until they are re-provisioned.

---

## Taken over by @dawsontoth — follow-up commits

Two commits on top of the original, addressing the bot review feedback and locking the new guard:

- `cc7ed910` — bot-comment fixes: an early-return guard instead of the `!` non-null assertion in `isProtectedComponentPackage`, dropping a redundant `String()` on an already-`string` path, and computing the delete-modal's protected selection inline instead of in a render-time `useMemo`. No behavior change.
- `5467cccc` — a `DeleteDirectoryOrFileModal` render test asserting the deletion driver is *not* called when the selection contains a protected component (and that a normal selection still deletes), running the real `isProtectedPath`. This is the coverage the guard was missing — mutation-verified: neutralizing the modal guard turns the protected-case test red.

The one bot suggestion I did **not** take: reordering `closeModal()` after the refusal check. The selection is preserved on refusal (the handler returns before touching it), so the toast's "remove it from the selection" advice already works with the modal closed; keeping the confirm dialog open buys nothing, since it cannot edit the selection.

## For the human reviewer

The matching strategy is the only judgment call, and cross-model review kept circling the same edge: `isProtectedComponentPackage` matches the repo name only as the spec's **trailing segment**, so it does *not* match a bare `status-check` package name, an archive/tarball URL (`…/akamai-status/archive/v1.0.0.tar.gz`), or an owner-position name — all deliberate, all pinned by negative tests. Every shape the PR claims we deploy (git URL, `.git`, committish, version, scoped npm) is covered and tested. The one thing studio cannot verify is the exhaustive set of package specs host-manager actually provisions for this component; if any real deploy shape falls outside "ends at the name", it needs a pattern. You own host-manager, so you are the check on that.

## Review coverage

Original change authored by Kris (human); the two follow-up commits authored by Claude (Opus 4.8). Cross-model review via the pre-push CLI, two rounds: gemini ✓ (both rounds, independent) and cursor-composer ✓ (round 1, independent); codex ✗ (workspace spend cap) and Harper domain adjudication ✗ (leg failed locally), so the outside findings were author-triaged. The two "major" flags both proved false positives: child files of a protected package are protected because `calculateRootEntries` inherits the root `package` to every descendant, and `isProtectedPath` resolves roots correctly because a root's `path` equals its `name` — both confirmed by the passing predicate tests.

## Verification

`vitest run` on the applications feature: 313 passed (was 311; +2 for the new modal test). `tsc -b`, `oxlint`, and `dprint check` all clean. The full-suite local failures are the pre-existing environmental jsdom `localStorage` set, unrelated to this change — CI is the gate.

## Follow-up review feedback — `d9927d78`

The latest review found one more mutation path: react-complex-tree's programmatic drag can move a file out of a protected component without consulting the item's `canMove` flag. `onInternalDrop` now checks every source path before it computes or executes a move. A rendered sidebar test drives that callback and proves `renameFiles` is not called.

This commit also closes the remaining fail-open/test gaps:

- `useEntryActions(undefined)` no longer offers Delete.
- `EditorViewProvider` has a render test that proves a protected opened entry sets `restrictPackageModification`.
- The synthetic Imported Applications and New Application entries have direct protection coverage.
- Archive, registry tarball, npm alias and trailing-space package shapes are pinned as deliberate matcher negatives. Host-manager's deployed `@harperdb/akamai-status@1.0.0` shape remains a positive case.

## Latest verification

- Applications feature: 28 files, 322 tests passed.
- Mutation check: neutralizing the drag, provider, synthetic-entry and missing-entry guards produces five expected test failures; restoring them passes 30/30 focused tests.
- Full repository suite: 2,212 passed, 11 skipped, 3 unrelated failures (`NotificationBell`, `installStaleDeployReload`, `OrgCard`).
- `oxlint` and `dprint check`: clean.
- `pnpm build`: blocked by an existing mixed Monaco dependency resolution. `@monaco-editor/react` resolves Monaco 0.52 types from the main checkout while this worktree has Monaco 0.56; none of the reported files are touched here.

Independent pre-push review at `d9927d78` failed closed. Under the required `no-claude` policy, Gemini's sandbox denied `pwd`; the low-risk delta policy selected no Cursor lens, so no outside-model receipt exists for this head.

<sub>Review-Coverage: authored=unknown; ran=none; rounds=1 @ d9927d786ded</sub>

<sub>Human-Review-Need: 4 @ d9927d786ded</sub>

### DIFF
--- a/src/features/instance/applications/context/EditorViewProvider.tsx
+++ b/src/features/instance/applications/context/EditorViewProvider.tsx
@@ -30,6 +30,7 @@ import { DirectoryEntry } from './directoryEntry';
 import { EditorViewContext, EditorViewContextValue } from './EditorViewContext';
 import { FileEntry } from './fileEntry';
 import { isDirectory } from './isDirectory';
+import { isProtectedComponentPackage } from './isProtectedComponentPackage';
 
 export function EditorViewProvider({ children }: PropsWithChildren) {
 	const navigate = useNavigate();
@@ -200,8 +201,7 @@ export function EditorViewProvider({ children }: PropsWithChildren) {
 		if (!openedEntry) {
 			return false;
 		}
-		return openedEntry.package?.includes('github.com/HarperDB/status-check-fabric')
-			|| openedEntry.package?.includes('github.com/HarperFast/status-check-fabric')
+		return isProtectedComponentPackage(openedEntry.package)
 			|| openedEntry.path === importedApplications
 			|| openedEntry.path === newApplication;
 	}, [openedEntry]);
--- a/src/features/instance/applications/context/EditorViewProvider.tsx
+++ b/src/features/instance/applications/context/EditorViewProvider.tsx
@@ -3,10 +3,7 @@ import { useInstanceClientIdParams } from '@/config/useInstanceClient';
 import {
 	calculateRootEntries,
 } from '@/features/instance/applications/components/ApplicationsSidebar/calculateRootEntries';
-import {
-	importedApplications,
-	newApplication,
-} from '@/features/instance/applications/components/ApplicationsSidebar/specialItems';
+import { newApplication } from '@/features/instance/applications/components/ApplicationsSidebar/specialItems';
 import { useEditorFileContent } from '@/features/instance/applications/context/editorFileContent';
 import { parseReadMe } from '@/features/instance/applications/lib/parseReadMe';
 import { useSessionStorage } from '@/hooks/useSessionStorage';
@@ -30,7 +27,7 @@ import { DirectoryEntry } from './directoryEntry';
 import { EditorViewContext, EditorViewContextValue } from './EditorViewContext';
 import { FileEntry } from './fileEntry';
 import { isDirectory } from './isDirectory';
-import { isProtectedComponentPackage } from './isProtectedComponentPackage';
+import { isProtectedEntry } from './isProtectedComponentPackage';
 
 export function EditorViewProvider({ children }: PropsWithChildren) {
 	const navigate = useNavigate();
@@ -197,14 +194,7 @@ export function EditorViewProvider({ children }: PropsWithChildren) {
 		],
 	);
 
-	const restrictPackageModification = useMemo(() => {
-		if (!openedEntry) {
-			return false;
-		}
-		return isProtectedComponentPackage(openedEntry.package)
-			|| openedEntry.path === importedApplications
-			|| openedEntry.path === newApplication;
-	}, [openedEntry]);
+	const restrictPackageModification = useMemo(() => isProtectedEntry(openedEntry), [openedEntry]);
 
 	/*
 	 Memoize the tracked state.
--- a/src/features/instance/applications/context/isProtectedComponentPackage.ts
+++ b/src/features/instance/applications/context/isProtectedComponentPackage.ts
@@ -39,12 +39,13 @@ export function isProtectedEntry(entry: DirectoryEntry | FileEntry | undefined) 
 }
 
 /**
- * Whether a tree path belongs to a protected component. Capability flags gate what renders; this
- * gates the mutation itself, which the delete modal reaches from the menu bar, the context menu
- * and a global keyboard shortcut alike — the latter two of which can act on a selection that was
- * never the entry whose capabilities were computed.
+ * Whether a tree path belongs to a protected component. Fails closed: a root that cannot be
+ * resolved — the tree has not loaded, or the path does not correspond to anything rendered — is
+ * treated as protected, because the cost of refusing a legitimate delete is a reload and the cost
+ * of allowing a wrong one is an instance dropping out of the load balancer.
  */
 export function isProtectedPath(rootEntries: Array<DirectoryEntry | FileEntry>, path: string) {
 	const project = String(path).split('/')[0];
-	return isProtectedEntry(rootEntries.find((entry) => entry.name === project));
+	const root = rootEntries.find((entry) => entry.name === project);
+	return root ? isProtectedEntry(root) : true;
 }
--- a/src/features/instance/applications/context/isProtectedComponentPackage.ts
+++ b/src/features/instance/applications/context/isProtectedComponentPackage.ts
@@ -1,13 +1,37 @@
+import {
+	importedApplications,
+	newApplication,
+} from '@/features/instance/applications/components/ApplicationsSidebar/specialItems';
+import type { DirectoryEntry } from '@/features/instance/applications/context/directoryEntry';
+import type { FileEntry } from '@/features/instance/applications/context/fileEntry';
+
 // Platform-managed components: editing or deleting one takes the instance out of the Akamai
 // load balancer. Matched on repo name rather than the full package URL so an org rename cannot
 // silently drop the guard.
 export const PROTECTED_COMPONENT_REPOS = ['status-check-fabric', 'akamai-status'];
 
-// The name must be a whole segment, so a customer package merely containing one of these
-// (`my-akamai-status-probe`) is not locked. Leading `/` or `@` covers git URLs and npm scopes;
-// trailing `.`, `#` or `@` covers `.git`, a committish, and a version.
-const PROTECTED_PATTERNS = PROTECTED_COMPONENT_REPOS.map((repo) => new RegExp(`(^|[/@])${repo}([.#@]|$)`));
+// The name has to be a whole segment, so neither a customer package that merely contains one
+// (`my-akamai-status-probe`) nor one that extends it (`akamai-status.dashboard`) is locked.
+const PROTECTED_PATTERNS = PROTECTED_COMPONENT_REPOS.map(
+	(repo) => new RegExp(`(?:^|[/@])${repo}(?:\\.git)?(?:[/#?@]|$)`),
+);
 
 export function isProtectedComponentPackage(packageSpec: string | undefined) {
 	return Boolean(packageSpec) && PROTECTED_PATTERNS.some((pattern) => pattern.test(packageSpec!));
+}
+
+/**
+ * Whether this entry may not be modified. Callers must pass the entry they are acting on —
+ * the sidebar context menu targets a row without opening it, so anything keyed to the opened
+ * entry offers Delete on a protected package the user merely right-clicked.
+ */
+export function isProtectedEntry(entry: DirectoryEntry | FileEntry | undefined) {
+	if (!entry) {
+		return false;
+	}
+	return (
+		isProtectedComponentPackage(entry.package)
+		|| entry.path === importedApplications
+		|| entry.path === newApplication
+	);
 }
--- a/src/features/instance/applications/context/isProtectedComponentPackage.ts
+++ b/src/features/instance/applications/context/isProtectedComponentPackage.ts
@@ -10,10 +10,12 @@ import type { FileEntry } from '@/features/instance/applications/context/fileEnt
 // silently drop the guard.
 export const PROTECTED_COMPONENT_REPOS = ['status-check-fabric', 'akamai-status'];
 
-// The name has to be the trailing segment of the spec — not a prefix (`my-akamai-status-probe`),
+// The name has to be the spec's own trailing segment — not a prefix (`my-akamai-status-probe`),
 // an extension (`akamai-status.dashboard`), or an owner (`github.com/akamai-status/theirs.git`).
+// A trailing `/` counts only at the very end, which is what separates the last two. Case-insensitive
+// because git hosts are: a spec that deploys need not match this regex's casing.
 const PROTECTED_PATTERNS = PROTECTED_COMPONENT_REPOS.map(
-	(repo) => new RegExp(`(?:^|[/@])${repo}(?:\\.git)?(?:[#@]|$)`),
+	(repo) => new RegExp(`(?:^|[/@])${repo}(?:\\.git)?(?:[#?@]|/?$)`, 'i'),
 );
 
 export function isProtectedComponentPackage(packageSpec: string | undefined) {
@@ -34,4 +36,15 @@ export function isProtectedEntry(entry: DirectoryEntry | FileEntry | undefined) 
 		|| entry.path === importedApplications
 		|| entry.path === newApplication
 	);
+}
+
+/**
+ * Whether a tree path belongs to a protected component. Capability flags gate what renders; this
+ * gates the mutation itself, which the delete modal reaches from the menu bar, the context menu
+ * and a global keyboard shortcut alike — the latter two of which can act on a selection that was
+ * never the entry whose capabilities were computed.
+ */
+export function isProtectedPath(rootEntries: Array<DirectoryEntry | FileEntry>, path: string) {
+	const project = String(path).split('/')[0];
+	return isProtectedEntry(rootEntries.find((entry) => entry.name === project));
 }
--- a/src/features/instance/applications/context/isProtectedComponentPackage.ts
+++ b/src/features/instance/applications/context/isProtectedComponentPackage.ts
@@ -44,7 +44,7 @@ export function isProtectedEntry(entry: DirectoryEntry | FileEntry | undefined) 
  * treated as protected, because the cost of refusing a legitimate delete is a reload and the cost
  * of allowing a wrong one is an instance dropping out of the load balancer.
  */
-export function isProtectedPath(rootEntries: Array<DirectoryEntry | FileEntry>, path: string) {
+export function isProtectedPath(rootEntries: ReadonlyArray<DirectoryEntry | FileEntry>, path: string) {
 	const project = String(path).split('/')[0];
 	const root = rootEntries.find((entry) => entry.name === project);
 	return root ? isProtectedEntry(root) : true;
--- a/src/features/instance/applications/context/isProtectedComponentPackage.ts
+++ b/src/features/instance/applications/context/isProtectedComponentPackage.ts
@@ -19,7 +19,10 @@ const PROTECTED_PATTERNS = PROTECTED_COMPONENT_REPOS.map(
 );
 
 export function isProtectedComponentPackage(packageSpec: string | undefined) {
-	return Boolean(packageSpec) && PROTECTED_PATTERNS.some((pattern) => pattern.test(packageSpec!));
+	if (!packageSpec) {
+		return false;
+	}
+	return PROTECTED_PATTERNS.some((pattern) => pattern.test(packageSpec));
 }
 
 /**
@@ -45,7 +48,7 @@ export function isProtectedEntry(entry: DirectoryEntry | FileEntry | undefined) 
  * of allowing a wrong one is an instance dropping out of the load balancer.
  */
 export function isProtectedPath(rootEntries: ReadonlyArray<DirectoryEntry | FileEntry>, path: string) {
-	const project = String(path).split('/')[0];
+	const project = path.split('/')[0];
 	const root = rootEntries.find((entry) => entry.name === project);
 	return root ? isProtectedEntry(root) : true;
 }
--- a/src/features/instance/applications/context/isProtectedComponentPackage.ts
+++ b/src/features/instance/applications/context/isProtectedComponentPackage.ts
@@ -1,8 +1,13 @@
 // Platform-managed components: editing or deleting one takes the instance out of the Akamai
 // load balancer. Matched on repo name rather than the full package URL so an org rename cannot
-// silently drop the guard, as the HarperDB -> HarperFast move already did once.
+// silently drop the guard.
 export const PROTECTED_COMPONENT_REPOS = ['status-check-fabric', 'akamai-status'];
 
+// The name must be a whole segment, so a customer package merely containing one of these
+// (`my-akamai-status-probe`) is not locked. Leading `/` or `@` covers git URLs and npm scopes;
+// trailing `.`, `#` or `@` covers `.git`, a committish, and a version.
+const PROTECTED_PATTERNS = PROTECTED_COMPONENT_REPOS.map((repo) => new RegExp(`(^|[/@])${repo}([.#@]|$)`));
+
 export function isProtectedComponentPackage(packageSpec: string | undefined) {
-	return PROTECTED_COMPONENT_REPOS.some((repo) => packageSpec?.includes(repo));
+	return Boolean(packageSpec) && PROTECTED_PATTERNS.some((pattern) => pattern.test(packageSpec!));
 }
--- a/src/features/instance/applications/context/isProtectedComponentPackage.ts
+++ b/src/features/instance/applications/context/isProtectedComponentPackage.ts
@@ -10,10 +10,10 @@ import type { FileEntry } from '@/features/instance/applications/context/fileEnt
 // silently drop the guard.
 export const PROTECTED_COMPONENT_REPOS = ['status-check-fabric', 'akamai-status'];
 
-// The name has to be a whole segment, so neither a customer package that merely contains one
-// (`my-akamai-status-probe`) nor one that extends it (`akamai-status.dashboard`) is locked.
+// The name has to be the trailing segment of the spec — not a prefix (`my-akamai-status-probe`),
+// an extension (`akamai-status.dashboard`), or an owner (`github.com/akamai-status/theirs.git`).
 const PROTECTED_PATTERNS = PROTECTED_COMPONENT_REPOS.map(
-	(repo) => new RegExp(`(?:^|[/@])${repo}(?:\\.git)?(?:[/#?@]|$)`),
+	(repo) => new RegExp(`(?:^|[/@])${repo}(?:\\.git)?(?:[#@]|$)`),
 );
 
 export function isProtectedComponentPackage(packageSpec: string | undefined) {
@@ -21,9 +21,9 @@ export function isProtectedComponentPackage(packageSpec: string | undefined) {
 }
 
 /**
- * Whether this entry may not be modified. Callers must pass the entry they are acting on —
- * the sidebar context menu targets a row without opening it, so anything keyed to the opened
- * entry offers Delete on a protected package the user merely right-clicked.
+ * Whether this entry may not be modified. Callers must pass the entry they are acting on: the
+ * sidebar context menu targets a row without opening it, so anything keyed to the opened entry
+ * gates on the wrong subject.
  */
 export function isProtectedEntry(entry: DirectoryEntry | FileEntry | undefined) {
 	if (!entry) {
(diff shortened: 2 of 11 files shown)


dawsontoth (approved):

src/features/instance/applications/context/isProtectedComponentPackage.ts
```diff
@@ -0,0 +1,51 @@
...
+// an extension (`akamai-status.dashboard`), or an owner (`github.com/akamai-status/theirs.git`).
+// A trailing `/` counts only at the very end, which is what separates the last two. Case-insensitive
+// because git hosts are: a spec that deploys need not match this regex's casing.
+const PROTECTED_PATTERNS = PROTECTED_COMPONENT_REPOS.map(
+	(repo) => new RegExp(`(?:^|[/@])${repo}(?:\\.git)?(?:[#?@]|/?$)`, 'i'),
+);
+
+export function isProtectedComponentPackage(packageSpec: string | undefined) {
+	return Boolean(packageSpec) && PROTECTED_PATTERNS.some((pattern) => pattern.test(packageSpec!));
+}
```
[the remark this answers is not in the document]
dawsontoth: Done in `cc7ed910` — replaced the `Boolean(...)` wrapper + `test(packageSpec!)` non-null assertion with an early-return guard, so `packageSpec` narrows to `string` normally:

```ts
if (!packageSpec) {
	return false;
}
return PROTECTED_PATTERNS.some((pattern) => pattern.test(packageSpec));
```

src/features/instance/applications/context/isProtectedComponentPackage.ts
```diff
@@ -0,0 +1,51 @@
...
+ * Whether a tree path belongs to a protected component. Fails closed: a root that cannot be
+ * resolved — the tree has not loaded, or the path does not correspond to anything rendered — is
+ * treated as protected, because the cost of refusing a legitimate delete is a reload and the cost
+ * of allowing a wrong one is an instance dropping out of the load balancer.
+ */
+export function isProtectedPath(rootEntries: Array<DirectoryEntry | FileEntry>, path: string) {
+	const project = String(path).split('/')[0];
+	const root = rootEntries.find((entry) => entry.name === project);
+	return root ? isProtectedEntry(root) : true;
+}
```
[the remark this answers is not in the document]
dawsontoth: Done in `cc7ed910` — dropped the redundant `String(path)`; `path` is already typed `string`, and the one caller that passes a `TreeItemIndex` (the delete modal) stringifies at the call site.

src/features/instance/applications/modals/DeleteDirectoryOrFileModal.tsx
```diff
@@ -10,14 +11,22 @@ import { setWatchedValue, useWatchedValue } from '@/lib/events/watcher';
...
-	const { openedEntry, reloadRootEntries, setFocusedItem, setSelectedItems, selectedItems } = useEditorView();
+	const { openedEntry, reloadRootEntries, rootEntries, setFocusedItem, setSelectedItems, selectedItems } =
+		useEditorView();
+
+	// Reached by a global Cmd+Delete shortcut that checks no capability, and deletes the whole
+	// selection rather than the entry the flags were computed for — so refuse here, at the mutation.
+	const protectedSelection = useMemo(
+		() => (isModalOpen ? selectedItems.filter(item => isProtectedPath(rootEntries, String(item))) : []),
+		[isModalOpen, rootEntries, selectedItems],
+	);
```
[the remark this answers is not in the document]
dawsontoth: Done in `cc7ed910` — the protected-selection filter is now computed inline inside `handleDeleteFolderOrFile`, and the `useMemo` (plus its `isModalOpen` gate) is gone. It runs once per delete click instead of allocating a new array on every render.

src/features/instance/applications/modals/DeleteDirectoryOrFileModal.tsx
```diff
@@ -40,6 +49,14 @@ export function DeleteDirectoryOrFileModal() {
 	const handleDeleteFolderOrFile = useCallback(async () => {
 		closeModal();
 
+		if (protectedSelection.length) {
+			toast.error(`${action} refused`, {
+				description: `${protectedSelection.join(', ')} is managed by Harper and keeps this instance in the `
+					+ 'load balancer. Remove it from the selection to continue.',
+			});
+			return;
+		}
```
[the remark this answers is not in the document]
dawsontoth: Kept `closeModal()` before the refusal check, deliberately. The selection is preserved on refusal — the handler `return`s before the focus/selection reset — so the toast's "remove it from the selection to continue" advice already applies with the modal closed. The confirm dialog has no per-item deselect, so leaving it open would only add a second Cancel click. Happy to revisit if you'd prefer it stay open.

src/features/instance/applications/modals/DeleteDirectoryOrFileModal.tsx
```diff
@@ -85,7 +102,16 @@ export function DeleteDirectoryOrFileModal() {
...
+	}, [
+		action,
+		closeModal,
+		instanceParams,
+		protectedSelection,
+		reloadRootEntries,
+		selectedItems,
+		setFocusedItem,
+		setSelectedItems,
+	]);
```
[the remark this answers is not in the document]
dawsontoth: Done in `cc7ed910` — with the filter inlined, the `handleDeleteFolderOrFile` deps now drop `protectedSelection` and add `rootEntries` (alongside `selectedItems`), matching this.

dawsontoth:
@kriszyp — I've taken this over to run it through the engineering guidelines and get it to Ready. Two commits on top of yours:

- `cc7ed910` — the gemini bot feedback: early-return guard instead of the `!` assertion in `isProtectedComponentPackage`, dropped a redundant `String()`, and inlined the delete-modal's protected-selection filter instead of a render-time `useMemo`. No behavior change. (I kept `closeModal()` before the refusal check — the selection is preserved on refusal, so the toast's advice still applies with the modal closed.)
- `5467cccc` — a `DeleteDirectoryOrFileModal` render test that asserts the deletion driver isn't called when the selection contains a protected component. It's the coverage the guard was missing; mutation-verified (deleting the modal guard turns it red).

**One thing for you to confirm** — it's the matching-completeness question your description already flags as the only judgment call. `isProtectedComponentPackage` matches the repo name as the spec's *trailing segment*, so it deliberately doesn't match a bare `status-check` package name, an archive/tarball URL (`…/akamai-status/archive/…`), or an owner-position name. Every shape the PR claims we deploy (git URL, `.git`, committish, version, scoped npm) is covered and tested. The one thing I can't verify from studio is the exhaustive set of specs host-manager actually provisions for this component — you own host-manager, so you're the check on whether any real deploy shape falls outside "ends at the name."

Cross-model review (pre-push CLI, 2 rounds): gemini ✓ both rounds and cursor-composer ✓ round 1, both independent. Codex was spend-capped and the domain adjudicator failed locally, which is the only reason the `Human-Review-Need` footer reads 4 — it's degraded coverage, not an open finding. The two "major" flags the outside legs raised (child files editable; `isProtectedPath` misresolving roots) I verified as false positives — `calculateRootEntries` inherits the root `package` to every descendant, and a root's `path` equals its `name`.

kriszyp:
@dawsontoth Confirmed against [Deploy akamai-status in place of status-check-fabric on fabric instances #183](https://github.com/HarperFast/host-manager/pull/183): its final diff passes `package: "@harperdb/akamai-status@1.0.0"` to `deploy_component`. That exact literal is already in `isProtectedComponentPackage`’s positive matrix, and the focused test passes 18/18. The trailing-segment matcher therefore covers the package shape host-manager actually provisions; no Studio code change is needed for this feedback.

— GPT-5.6 Codex

src/features/instance/applications/context/isProtectedComponentPackage.ts
```diff
@@ -0,0 +1,54 @@
...
+	);
+}
+
+/**
+ * Whether a tree path belongs to a protected component. Fails closed: a root that cannot be
+ * resolved — the tree has not loaded, or the path does not correspond to anything rendered — is
+ * treated as protected, because the cost of refusing a legitimate delete is a reload and the cost
+ * of allowing a wrong one is an instance dropping out of the load balancer.
+ */
+export function isProtectedPath(rootEntries: ReadonlyArray<DirectoryEntry | FileEntry>, path: string) {
```
cb1kenobi, line 50: **High: a fifth delete path — internal tree drag — moves files out of the protected component**

The description enumerates four client paths and closes three. There is a fifth that neither `useEntryActions` nor the modal guard sees, and it ends in the same `dropComponent` call.

`buildItems` sets `canMove: !entry.package` on files, which is the only source-side gate. But react-complex-tree's programmatic drag (`startProgrammaticDrag`, bound to `control+shift+d` by default — studio overrides neither `keyboardBindings` nor supplies `canDrag`) checks only `environment.canDragAndDrop` and never consults `item.canMove` (`react-complex-tree/lib/esm/drag/DragAndDropProvider.js:149-168`). `useCanDropAt` filters on the drop target only.

`ApplicationsSidebar/index.tsx:78` then checks only `items[target.targetItem]?.data?.package` — again the **target**, never the source. So with `status-check/resources.js` selected: `Ctrl+Shift+D` → drop on `anvils` → `useRenameFiles` writes the file into the new project and then calls `dropComponent({ project: 'status-check', file: 'resources.js' })` (`useRenameFiles.tsx:90`), deleting it out of the protected component.

Suggested fix — gate the **source** in `onInternalDrop`, reusing this helper:

```ts
if (droppedItems.some(item => isProtectedPath(rootEntries, String(item.index)))) {
	toast.error('Move refused', {
		description: 'That component is managed by Harper and keeps this instance in the load balancer.',
	});
	return;
}
```

(Commenting here because `ApplicationsSidebar/index.tsx` isn't in the diff — `isProtectedPath` is the helper the fix would reuse.)

—
_Generated by Barber AI_

src/features/instance/applications/context/EditorViewProvider.tsx
```diff
@@ -196,15 +194,7 @@ export function EditorViewProvider({ children }: PropsWithChildren) {
...
-	const restrictPackageModification = useMemo(() => {
-		if (!openedEntry) {
-			return false;
-		}
-		return openedEntry.package?.includes('github.com/HarperDB/status-check-fabric')
-			|| openedEntry.package?.includes('github.com/HarperFast/status-check-fabric')
-			|| openedEntry.path === importedApplications
-			|| openedEntry.path === newApplication;
-	}, [openedEntry]);
+	const restrictPackageModification = useMemo(() => isProtectedEntry(openedEntry), [openedEntry]);
```
cb1kenobi, line 197: **Medium: surviving mutation — the read-only-editor surface has no test at all**

Mutation-tested the new guard against the full applications suite (`vitest run src/features/instance/applications`, baseline **313 passed**). Eight of ten mutations died. This one survived:

> replacing this line with `const restrictPackageModification = useMemo(() => false, [openedEntry]);` → **313 passed (313)**, green.

`restrictPackageModification` is what `ContentActions` (menu bar + the "cannot be modified" tooltip), `TextEditorView` (line 106 — whether the editor is writable) and `DropTarget` (line 23 — upload gating) read. That is the guard's *original* purpose: making the platform component read-only. It can be fully neutralised and CI stays green.

Given the premise of this PR — a guard that already failed silently once — the editability half deserves the same lock the delete half just got. One `TextEditorView` or provider render test asserting the editor is read-only for a protected `openedEntry` would kill it.

(For the record, the mutations that *did* die: matcher→`false` (15 failed), dropping either repo from the list (13 / 2 failed), `isProtectedPath` fail-open (2), regex→naive substring (5), `canDeleteEntry` ungated (1), `canRedeploy` ungated (1), modal guard neutralised (1). The delete path is well pinned.)

—
_Generated by Barber AI_

src/features/instance/applications/context/isProtectedComponentPackage.ts
```diff
@@ -0,0 +1,54 @@
...
+// load balancer. Matched on repo name rather than the full package URL so an org rename cannot
+// silently drop the guard.
+export const PROTECTED_COMPONENT_REPOS = ['status-check-fabric', 'akamai-status'];
+
+// The name has to be the spec's own trailing segment — not a prefix (`my-akamai-status-probe`),
+// an extension (`akamai-status.dashboard`), or an owner (`github.com/akamai-status/theirs.git`).
+// A trailing `/` counts only at the very end, which is what separates the last two. Case-insensitive
+// because git hosts are: a spec that deploys need not match this regex's casing.
+const PROTECTED_PATTERNS = PROTECTED_COMPONENT_REPOS.map(
+	(repo) => new RegExp(`(?:^|[/@])${repo}(?:\\.git)?(?:[#?@]|/?$)`, 'i'),
```
cb1kenobi, line 18: **Low: the matcher still fails open, so a *component* rename reproduces the original incident**

Not a blocker — kriszyp confirmed host-manager#183 ships `@harperdb/akamai-status@1.0.0`, which I verified matches. Raising it because the PR's thesis is removing the silent-failure class, and it removes the *org-rename* instance rather than the class itself: `isProtectedComponentPackage` returns `false` for any name it doesn't already know.

I ran this exact regex over an evasion/near-miss set. Every negative the tests pin holds, and forks/other orgs match as intended. These do not match:

| spec | matched |
|---|---|
| `https://github.com/HarperFast/akamai-status/archive/v1.0.0.tar.gz` | no |
| `https://registry.npmjs.org/@harperdb/akamai-status/-/akamai-status-1.0.0.tgz` | no |
| `npm:akamai-status@1.0.0` | no |
| `…/akamai-status/?branch=main` | no (`…/akamai-status?branch=main` matches) |
| `'@harperdb/akamai-status '` (trailing space) | no |

I'd explicitly **not** suggest widening `(?:[#?@]|/?$)` to allow a following segment — that would also start matching `github.com/akamai-status/their-component.git`, which you deliberately pinned as a negative. "A segment follows the name" cannot separate the owner position from a tarball path without knowing the spec is a URL and which segment is the repo.

So the two honest options are the ones you already named: keep the deny-list and accept that it must be updated in lockstep with host-manager, or move the invariant to a server-supplied "platform managed" flag on the component. Worth adding the shapes above as documented negative cases either way, so the assumption is visible to whoever renames this next.

Also an asymmetry worth a comment: `isProtectedPath` fails **closed** on an unresolved root (line 53) while this matcher fails **open**. Only the delete modal gets the fail-closed half.

—
_Generated by Barber AI_

src/features/instance/applications/context/isProtectedComponentPackage.ts
```diff
@@ -0,0 +1,54 @@
...
+ * gates on the wrong subject.
+ */
+export function isProtectedEntry(entry: DirectoryEntry | FileEntry | undefined) {
+	if (!entry) {
+		return false;
+	}
+	return (
+		isProtectedComponentPackage(entry.package)
+		|| entry.path === importedApplications
+		|| entry.path === newApplication
```
cb1kenobi, line 40: **Low: surviving mutation — the two special-item clauses are untested**

Deleting both `entry.path === importedApplications` and `entry.path === newApplication` from this expression leaves the full applications suite at **313 passed (313)**, green.

They carried over from the old `restrictPackageModification` and are what stops the synthetic "Imported Applications" / "New Application" rows being treated as editable entries. Since the module now has a dedicated test file, two cases in `isProtectedComponentPackage.test.ts` (or a small `isProtectedEntry` describe) would pin them.

—
_Generated by Barber AI_

src/features/instance/applications/hooks/useEntryActions.ts
```diff
@@ -27,20 +27,20 @@ export interface EntryActions {
...
+	const isProtected = isProtectedEntry(entry);
 	// An application root is the top-level directory whose path is just the project
 	// name (e.g. `anvils`, or an imported app under "Imported Applications").
 	const isApplicationRoot = !!entry && isDirectory(entry) && entry.path === entry.project;
 	const canEditFile = !!entry && !isDirectory(entry) && !isReadOnlyPackage && canManageBrowseInstance;
 	const canRename = !!entry && !isReadOnlyPackage && canManageBrowseInstance && !isApplicationRoot;
 	const canAddEntries = !!entry && !isReadOnlyPackage && canManageBrowseInstance;
 	const canAddTable = !!entry && entry.path.endsWith('.graphql') && canManageBrowseInstance;
-	const canDeleteEntry = !restrictPackageModification && canManageBrowseInstance;
+	const canDeleteEntry = !isProtected && canManageBrowseInstance;
```
cb1kenobi: **Low: `canDeleteEntry` is the one flag that lost its implicit entry guard**

Before this change, `useEntryActions(undefined)` still inherited the opened entry's restriction through context. Now `isProtectedEntry(undefined)` is `false`, so `canDeleteEntry` is `true` for no entry at all — while every sibling on lines 39-40 opens with `!!entry &&`.

Not reachable today: `ContentActions` renders the whole bar inside `{openedEntry && …}` and `FileTreeContextMenu` renders inside `{target && …}`. But this PR is precisely about a guard that fails open, and this is the only flag whose default is "allowed".

Suggested fix:
```suggestion
	const canDeleteEntry = !!entry && !isProtected && canManageBrowseInstance;
```

—
_Generated by Barber AI_

src/features/instance/applications/context/isProtectedComponentPackage.ts
kriszyp: Addressed in d9927d78: onInternalDrop now rejects any source path belonging to a protected component before target processing, and the rendered sidebar test proves renameFiles is never called for that move. — GPT-5.6 Codex

src/features/instance/applications/context/EditorViewProvider.tsx
kriszyp: Addressed in d9927d78 with an EditorViewProvider render test that opens the deployed protected package shape and asserts restrictPackageModification flips true. Neutralizing the provider assignment makes this test fail. — GPT-5.6 Codex

src/features/instance/applications/context/isProtectedComponentPackage.ts
kriszyp: Keeping the explicit deny-list policy: Kris confirmed host-manager deploys @harperdb/akamai-status@1.0.0, which is covered. d9927d78 adds the archive, registry tarball, npm alias, and trailing-space shapes as negative tests so this boundary is visible rather than accidental. — GPT-5.6 Codex

src/features/instance/applications/context/isProtectedComponentPackage.ts
kriszyp: Addressed in d9927d78: both synthetic Imported Applications and New Application entries now have direct isProtectedEntry coverage. Removing either clause fails the test. — GPT-5.6 Codex

src/features/instance/applications/hooks/useEntryActions.ts
kriszyp: Addressed in d9927d78: canDeleteEntry now requires an actual entry, with a regression test for useEntryActions(undefined). — GPT-5.6 Codex

cb1kenobi:
Re-reviewed `d9927d78` (incremental over `5467cccc`) — **no new issues found, and every prior finding is closed or accounted for.** Nice work on this one; the drag fix landed in the right place.

## The High finding is genuinely closed

This was the one I was most careful about, because the finding was specifically that `canMove` is *not* consulted by react-complex-tree's programmatic drag — so a fix that merely re-tightened `canMove` would not have closed it.

The `+8/-1` in `ApplicationsSidebar/index.tsx` puts a **source-side** guard at the top of `onInternalDrop`, and `onInternalDrop` is wired as `onDrop` on `ControlledTreeEnvironment`. Reading react-complex-tree 2.6.2's shipped source, that is the terminal choke point for **both** input paths:

- mouse drop → `onDropHandler()` → `environment.onDrop(...)`
- `Ctrl+Shift+D` → `completeProgrammaticDrag()` → the *same* `onDropHandler()` → `environment.onDrop(...)`

(`lib/cjs/drag/DragAndDropProvider.js` — `completeProgrammaticDrag` and the mouse handler share `onDropHandler`.)

So the guard sits below the point where the two paths diverge, which is exactly right, and it fails closed via `isProtectedPath`. The new `ApplicationsSidebar.test.tsx` drives the real `onInternalDrop` callback with the real `isProtectedPath` and asserts `renameFiles` is not called — behavior, not a flag assertion. It stubs the tree library rather than firing the actual keybinding, which is a reasonable seam; I verified the library-side routing separately by reading its source, so the two halves meet.

## Mutation results — 12/12 killed, no regressions

Baseline at this head: **322 passed / 28 files** (`vitest run src/features/instance/applications`).

| Mutation | Prior round | Now |
|---|---|---|
| provider `restrictPackageModification` → `false` | **survived** | killed (1 failed / 321) |
| drop both special-item clauses | **survived** | killed (2 failed / 320) |
| drag source guard neutralised *(new)* | n/a | killed (1 failed / 321) |
| `canDeleteEntry` loses `!!entry` *(the `+1/-1`)* | n/a | killed (1 failed / 321) |
| matcher → `return false` | killed | killed (17 failed / 305) |
| drop `akamai-status` from repo list | killed | killed (15 failed / 307) |
| drop `status-check-fabric` from repo list | killed | killed (2 failed / 320) |
| `isProtectedPath` fail-open (`: true` → `: false`) | killed | killed (2 failed / 320) |
| regex → naive substring | killed | killed (9 failed / 313) |
| `canDeleteEntry` drops `!isProtected` | killed | killed (1 failed / 321) |
| `canRedeploy` drops `!isProtected` | killed | killed (1 failed / 321) |
| modal guard neutralised | killed | killed (1 failed / 321) |

Both previously-surviving mutations now die, and nothing that was pinned before came loose. `oxlint`, `dprint check` and `tsc -b` all exit 0 in a clean worktree.

## One thing I chased and cleared

I went looking for a sixth path via `DropTarget` — the OS file-drop upload. `onUploadDrop` takes `targetProject` straight from the hovered row's `data-rct-item-id` with no protection check, and the render gate `if (!canUpload && !dragTarget)` lets the active dropzone render even when `canUpload` is false, so on paper you could overwrite `status-check/resources.js` by dragging a file onto the row.

It's closed, but indirectly, and it's worth writing down because the chain isn't obvious: `useDraggingHook` refuses to set `dragTarget` for any row containing `.packageIsLocked`; `ItemTitle` renders `LockedIcon` for every entry with a `package`; and `calculateRootEntries` inherits the root's `package` to every descendant. So no row inside a packaged component can ever become a drag target. Not a finding — just a load-bearing coupling that no test pins, if you ever want a cheap one.

## Still open (unchanged, and I think correctly so)

`Chat/tools/dropComponentFile/execute.ts` still calls `dropComponent` with a bare path and no component metadata in scope. You flagged this yourself and asked for direction, so I'm not re-litigating it — noting only that it's unchanged at this head, and that your argument in the description (five client paths now, and an ungated REST endpoint underneath all of them) reads as the right one to me. A server-side check would retire the whole class.

The fail-open matcher thread stays resolved on your call, which I think is fair: kriszyp confirmed the deployed spec is `@harperdb/akamai-status@1.0.0` and that matches. Worth being aware the delta pins four more shapes as deliberate non-matches — the npm alias `npm:akamai-status@1.0.0`, the registry tarball URL, the archive tarball, and a trailing-space `'@harperdb/akamai-status '`. All fine against today's deploy shape; the last one is the only one a `.trim()` would close for free, if you ever want it.

**Open Critical/High/Medium: 0.**

—
_Generated by Barber AI (Claude Opus 5)_

PR #1613, opened 2026-08-12
fix(auth): report a failed sign-up in the form, not a fading toast

### LINKED ISSUE #1612: [RUM] Half of sign-up submissions rejected 409 with no inline feedback — 7 of 8 affected sessions never created an account
## Summary

Datadog RUM shows that **half of all sign-up submissions in the last 24h were rejected with HTTP 409**, and the affected people mostly gave up rather than recovering. Sign-up is the top of the funnel, so this is worth fixing even at small absolute numbers.

Surfaced by the daily automated RUM review (RUM app `f590deee-…`, `env:prod`, last 24h).

## Datadog findings

`POST /User/` — the only request the e-mail sign-up form makes:

| Window | 201 Created | 409 Conflict |
| --- | --- | --- |
| 24h → now | 21 | **20** |
| 48h → 24h ago | 22 | **0** |

- **Reach (last 24h):** 8 sessions saw a 409. **7 of them never got a 201** — no account was created.
- **Where:** all on `#/sign-up`. Server responds fast (~160 ms – 1.3 s), so these are deliberate rejections, not timeouts.
- **Not a new endpoint behavior:** ~19 409s in the preceding 30 days total, so 20 in one day is a step change in *how often people hit it*, not a newly introduced status code.
- **Both releases:** 13 on `prod_fdc1142` (v2.159.2), 7 on v2.156.4 — so this is not tied to the release that went out yesterday.

### The retry signature

Per-session, the 409s come in runs of two or three, **~5 seconds apart**, then stop:

```
session A   409  409  409     (5.2s, 4.6s apart)
session B   409  409  409
session C   409  409  409
session D   409  409  409
session E   409  409  409
session F   409  409
session G   409  409
session H   201, then 409 two minutes later
```

`useMutation` doesn't retry by default, so each of those is a person clicking **Sign Up For Free** again. Session H is the tell for what a 409 means here: it created an account, then re-submitted the *same* address from `#/sign-up?me=…` and got the conflict — i.e. 409 on `POST /User/` is "this e-mail is already registered".

## Why people retry

Two gaps in [`SignUp.tsx`](https://github.com/HarperFast/studio/blob/stage/src/features/auth/SignUp.tsx):

1. **The submit button is never disabled.** `useSignUpMutation()` is destructured as `{ mutate: submitSignUpData }` — `isPending` is unused, and the button has no `disabled`. Every sibling auth form does guard on it (`SignIn.tsx`, `ForgotPassword.tsx`, `ResetPassword.tsx`). So there is no in-flight feedback and nothing stopping a second submit.

2. **A 409 has no specific handling.** `submitSignUpData(userData, { onSuccess })` passes no `onError`, so the rejection falls through to the shared `MutationCache` toast in `src/react-query/queryClient.ts`. That toast can't attach to the e-mail input or surface the "Already have an account? Sign in instead." link that's already on the page — beside an otherwise unchanged form, it reads as "try again".

The established pattern for exactly this is right next door: `useCloudSignIn` opts its mutation out with `meta: { skipGlobalErrorToast: true }` and renders the auth-specific rejection itself.

## Suggested fix

- Disable the submit button on `isPending`.
- Opt the sign-up mutation out of the global toast and, on 409, set a server error on the `email` field ("An account with this email already exists. Sign in instead, or use another email.") and focus it. Keep the standard toast for every other status.

## Notes

- No customer identifiers included; session IDs and e-mail addresses deliberately omitted.
- Doesn't cover the OAuth sign-up buttons — they redirect and don't go through this mutation.

dawsontoth:
Closes #1612.

## What RUM showed

`POST /User/`, prod, last 24h: 21 × 201, 20 × 409 across 8 sessions — and **7 of those 8 sessions never created an account**. The previous 24h had 22 submissions and zero 409s. The rejections arrive in runs of two or three per session, ~5 s apart; `useMutation` doesn't retry, so those are people clicking **Sign Up For Free** again.

## Scope

Deliberately **not** handling the 409 specifically. It's being removed from central-manager shortly, so mapping it here would be dead code — and per @DavidCockerill's review, encoding it in the UI would turn closing the [signup enumeration oracle](https://github.com/HarperFast/studio/pull/1613#issuecomment-5268705252) into "you're removing a helpful message" rather than a quiet server-side edit.

What's left are the two causes of the retries, both independent of how the server answers a duplicate email:

- **The submit button had no `isPending` guard** — unlike `SignIn`, `ForgotPassword`, and `ResetPassword`, which all disable on it. No in-flight feedback, nothing to stop a second click.
- **The failure only appeared in the shared mutation toast** — away from the inputs and gone after a few seconds. It now renders in the form, above the submit button, and persists until the next attempt.

The inline message is whatever the server said. `describeError` is split out of `errorHandler` so the inline text and the toast text come from one extractor and can't drift; it returns the pre-split sentence as `message` for the inline line (which has no heading to move a `"Conflict: …"` first clause into) alongside the toast's `title`/`description`. `errorHandler` keeps its existing behavior, including the timeout/403 toast-collapse ids.

The form maps **no** status codes, so whatever **error** status replaces the 409 reaches the user unchanged. One caveat, per [review](https://github.com/HarperFast/studio/pull/1613#discussion_r3768232606): that doesn't extend to a **bodyless 2xx**. `onSignUpSubmit`'s `if (data) … else throw new Error('Something went wrong')` would turn a `202` with an empty body into a client-manufactured error shown inline, instead of navigating to `/verifying`. Unreachable against today's endpoint, which always returns a body; tracked as HarperFast/central-manager#636 against #630.

## Verification

- `npx vitest run` — 282 files, 2167 pass, 11 skipped. The 6 reported errors are pre-existing environment noise (undici WebSocket `Event`, jsdom `scrollTo`/navigation), unrelated to these files.
- `SignUp.test.tsx` renders the real form against a `QueryClient` built from the app's own `mutationErrorHandler`, so the `skipGlobalErrorToast` wiring is under test rather than restated: the server's reason appears inline and *not* as a toast; a 400 object body, a 409 bare string, and a bodyless 503 each surface inline (proving no status is special-cased); the message clears on resubmit; the button disables while in flight. Each of the 6 was confirmed to **fail** with the corresponding change reverted.
- `npx tsc -b`, `npx oxlint`, `npx dprint check` clean.
- No browser pass on this revision — dev servers can't be started from an unattended session. The inline element reuses `FormMessage`'s exact classes (`text-destructive text-sm`), and behavior is covered by the DOM-level tests above.

Not covered: the Google/GitHub sign-up buttons, which redirect and never reach this mutation.


src/features/auth/SignUp.test.tsx
```diff
@@ -0,0 +1,106 @@
+/**
+ * @vitest-environment jsdom
+ */
+import { QueryClientProvider } from '@tanstack/react-query';
+import { fireEvent, render, screen, waitFor } from '@testing-library/react';
+import { AxiosError } from 'axios';
+import { PropsWithChildren } from 'react';
+import { afterEach, describe, expect, it, vi } from 'vitest';
```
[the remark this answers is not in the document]
dawsontoth: Done — \`beforeEach\` is imported and builds a fresh client per test.

src/features/auth/SignUp.test.tsx
```diff
@@ -0,0 +1,106 @@
...
+	toast: { error: vi.fn(), dismiss: vi.fn() },
+}));
+
+vi.mock('@/integrations/reo/reo', () => ({ reoClient: { identify: vi.fn() } }));
+
+import { toast } from 'sonner';
+// The real client is used on purpose: the global mutation-error toast lives on its
+// MutationCache, so this is what proves `skipGlobalErrorToast` is wired up.
+import { queryClient } from '@/react-query/queryClient';
+import { SignUp } from './SignUp';
```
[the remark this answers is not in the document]
dawsontoth: Fixed, though not by restating the cache config in the test: whether a 409 reaches the global toast is the thing under test, so a copy here would keep passing if \`skipGlobalErrorToast\` stopped being honored. Extracted the routing into an exported \`mutationErrorHandler\` in \`queryClient.ts\` and both the app client and the per-test client now use that one definition.

src/features/auth/SignUp.test.tsx
```diff
@@ -0,0 +1,106 @@
...
+}
+
+function submit() {
+	fireEvent.click(screen.getByRole('button', { name: 'Sign Up For Free' }));
+}
+
+afterEach(() => {
+	vi.clearAllMocks();
+	queryClient.clear();
+});
```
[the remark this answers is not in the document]
dawsontoth: Done — \`beforeEach\` builds the client (with \`mutations: { retry: false }\`) and \`afterEach\` is back to just \`vi.clearAllMocks()\`.

DavidCockerill (commented):
Nice, well-evidenced work — the funnel data in #1612 makes the case, and the implementation is clean. I'm commenting rather than approving on one question that I think should be answered before this ships, not because anything here is wrong.

**The question: is `POST /User` deliberately outside the enumeration policy the other auth endpoints follow?**

Two of the three public auth endpoints have explicit anti-enumeration engineering, with comments saying so:

- `Login.js:82-100` verifies the password **first** and only then checks `isVerified` / DEACTIVATED — commented as being precisely "so an unauthenticated caller can't distinguish these accounts from a wrong-password failure." Unknown email and wrong password both return the same 401.
- `Password.js:35-49` returns one constant response for unknown / DELETED / DEACTIVATED / PENDING, **and** starts a 500ms minimum-delay floor before the lookup, commented as "eliminating the timing oracle."

`POST /User` never got that treatment. It's public (`User.allowCreate` takes `user` as its first parameter and never consults it) and answers **409 vs 201** to any unauthenticated caller.

**To be clear about what this PR does and doesn't do:** it doesn't create that oracle. I traced every `USER_STATUS` branch in `addUser` — the response collapses to a clean binary, with invited-PENDING, DEACTIVATED and soft-DELETED all returning 201 and indistinguishable from a free address. No status code, body, header or timing changes here; only a string rendered in the DOM. A scripted client reads the status code off the wire and never instantiates React — #1612 is itself built from those status codes.

I also went looking for whether this makes the current signup-abuse traffic cheaper and concluded it doesn't, for a better reason than "the status code is unchanged": **`POST /User` has no cheap reconnaissance mode.** A miss returns 201 *and* creates an account *and* mails the address — probing **is** the payload. You can't quietly validate a purchased list against it. The only addresses probeable for free are ones already registered, which teach an attacker nothing new.

**So why hold rather than approve.** Right now closing that oracle is a quiet server-side edit nobody would notice. Once the UI tells users, closing it becomes "you're removing a helpful message," and that argument usually wins. That's a one-way door, and it's cheap to check before walking through it and expensive after.

**And there's a real complication worth knowing before anyone answers.** The textbook fix for a signup endpoint is: always return 202 and tell them *by email* — "you already have an account, sign in" vs "verify your address" — which is the shape Login and ForgotPassword already follow in spirit. But today a probe against a **registered** address sends **no mail at all** (the 409 throws before anything is created). Under always-202, every probe sends mail. Given the current signup traffic, the textbook fix would make the mail volume strictly worse. So the sequencing is probably: rate-limit `POST /User` first, then decide about the oracle — and rate limiting is the higher-value change either way.

If the answer is "yes, `POST /User` is deliberately outside the policy, the UX is worth it" — that's a perfectly good answer and I'll approve on it. I'd just like it written down somewhere, because right now the codebase says the opposite in two places and nothing says it here.

Two small notes in threads, both non-blocking and unrelated to the above.

Credit where it's due: declining the literal shape of @gemini-code-assist's test-isolation suggestion was the right call — restating the cache config in the test would have kept it green if `skipGlobalErrorToast` ever stopped being honoured — and extracting a shared `mutationErrorHandler` is a better answer than the one asked for.

_Reviewed by Claude Opus 5 for @DavidCockerill._

src/features/auth/isEmailAlreadyRegisteredError.ts
```diff
@@ -0,0 +1,18 @@
+import { isAxiosError } from 'axios';
+
+/**
+ * Detects the sign-up rejection that means "an account already uses this email address".
+ *
+ * central-manager's `User` resource answers a duplicate registration with HTTP 409 Conflict.
+ * Creating a user is the only thing Studio POSTs to `/User/`, and a duplicate email is the
+ * only conflict that endpoint can report, so the status alone identifies the case — unlike
```
DavidCockerill: Non-blocking, comment accuracy only. The doc comment's load-bearing claim — that a duplicate email is the only conflict this endpoint can report — isn't quite right, and it's what justifies keying on the status code alone.

`addUser` calls `searchByValue`, which itself throws CONFLICT for `Multiple <email> records found` (central-manager `src/lib/sharedHelpers.js:63`). That state is reachable today: `User.delete()` is a **soft** delete (`User.js:112` patches `status: DELETED`) while `addUser` only conflicts on ACTIVE/CLOUD_MIGRATED — so a deleted user re-signing up writes a *second* row on the same email, and every later `POST /User` for that address 409s down this path instead.

The predicate's behaviour is still correct either way — both 409s do mean the address is taken — so this is a one-sentence comment fix, not a code change.

src/features/auth/SignUp.tsx
```diff
@@ -97,8 +99,24 @@ export function SignUp() {
 				clearUtmParamsFromUrl();
 				void navigate({ to: '/verifying?email=' + encodeURIComponent(userData.email) });
 			},
+			// The sign-up mutation opts out of the global error toast (meta.skipGlobalErrorToast)
+			// so an already-registered email lands on the field that has to change, next to the
+			// "Sign in instead" link — a generic toast left people resubmitting the same address
+			// two and three times before giving up.
+			onError: (error) => {
+				if (isEmailAlreadyRegisteredError(error)) {
```
DavidCockerill: Forward-looking, and contingent on central-manager#603 — not a reason to hold this PR.

#603 adds DEACTIVATED to `addUser`'s conflict branch, so deactivated accounts will start returning 409 and landing on this message. It tells them to "Sign in instead", and `Login.js:98-100` answers with 403 `User account deactivated` — sign up, go sign in, your account is off, no next step.

Worth a note on #603 so whoever lands it decides whether the copy needs a softer ending (e.g. "…or contact support"), or whether that's acceptable for a state that should be rare.

src/features/auth/isEmailAlreadyRegisteredError.ts
dawsontoth: Confirmed and fixed in 7c69b47c — verified `searchByValue` (sharedHelpers.js:61) throws CONFLICT on >1 row and that `delete()` only patches `status: DELETED`, so the second row is reachable exactly as you describe. Comment now names both 409 paths and rests on the property the predicate actually needs (the address is unavailable), not on there being one cause.

src/features/auth/SignUp.tsx
dawsontoth: Verified against central-manager#603's diff — it does add `USER_STATUS.DEACTIVATED` to `addUser`'s conflict branch, so the dead-end you describe is real once it lands. Leaving the copy as-is here since it's correct for today's conflict set, and raising the wording on #603 so it's decided alongside the status change rather than pre-emptively softened for a state that doesn't 409 yet.

dawsontoth:
Thanks — the analysis is right, and I'd reached the same conclusion about the oracle independently while investigating #1612 (Dawson asked the same question in parallel). I can't answer whether it's *deliberate* — that's a call for whoever owns the auth policy, and I found nothing in either repo asserting it either way, which is your point. What I can add is four pieces of evidence, one of which corrects something I'd said earlier.

**1. There is nothing to rate-limit with today.** I grepped both repos for rate limiting, throttling, captcha, Turnstile and WAF config: nothing. The only gate on `POST /User/` is `User.allowCreate`'s optional `ALLOWLIST_EMAIL_DOMAINS` check, which prod can't be using since arbitrary signups work. So "rate-limit first" isn't tightening an existing limit, it's adding the first one.

**2. No browser-based scanning is happening now.** RUM, 30 days: the maximum `POST /User/` count in any single session is **3**, and the 409s total 36 across 14 sessions. That rules out abuse *through the UI* only — a scripted client never appears in RUM at all — but it does mean the current 409 volume in #1612 is genuinely people mistyping their own situation, not reconnaissance.

**3. You're right about the mail volume, and it inverts the fix I'd recommended.** I'd suggested always-202 as the fix. That's wrong on this endpoint for exactly the reason you give: today a probe against a registered address throws before anything is created and sends **no** mail, so always-202-plus-notification makes total mail strictly worse, and it doesn't touch the unregistered path that's already creating an account and mailing per probe. Uniformity fixes the oracle and *nothing else*; rate limiting fixes the abuse. Sequencing rate-limit-first is right.

**4. One thing worth adding to the risk side, because it's not enumeration.** PENDING returning 201 isn't only indistinguishable — it's reachable. `addUser` routes a PENDING record to `upgradePendingUser`, which calls `add_user` with the **caller-supplied password** and flips status to ACTIVE, with no invite token required; knowing an invited address is the whole prerequisite. What stops it being takeover is `isVerified` staying false in prod and `Login.js:93` rejecting unverified logins with 403 — that gate holds. But the legitimate invitee is then locked out (their own signup now 409s), and if they click the verification link the attacker's request just mailed them, the account activates under the attacker's password. Separately, the soft-delete duplicate-row path from the other thread permanently 409s an address after delete-then-resignup. Both argue the same way you do: the endpoint needs a limiter more than it needs a uniform response.

**On the one-way door.** Fair, and I'd rather not be the reason it closes. Two options, both fine by me: hold the message and land only the `isPending` half now (that half is unconditionally correct and is half the retries in #1612), or land it whole with the copy explicitly marked as revisitable. Dawson has the decision and the full write-up; deferring to whichever he and you land on.

_🤖 Posted by [Claude Code](https://claude.com/claude-code)_

dawsontoth:
Scope reduced in 35bef391 per @dawsontoth — the 409 mapping is gone, since central-manager is dropping that status shortly and you're right that shipping the message first makes closing the oracle a UX argument instead of a server edit.

What remains is status-agnostic: `isPending` on the submit button, and the failure rendering in the form above the button instead of only in a toast that fades away from the inputs. Both causes of the #1612 retries, neither dependent on what replaces the 409. `isEmailAlreadyRegisteredError` and its test are deleted, so the doc-comment thread about `Multiple <email> records found` is moot — though that soft-delete duplicate-row path is still a real central-manager bug independent of this PR, and I don't think it's tracked anywhere.

The inline text comes from a `describeError` split out of `errorHandler`, so it can't drift from the toast, and the tests now assert a 400 object body, a 409 bare string and a bodyless 503 all surface inline — i.e. no status is special-cased and whatever replaces the 409 needs no client change. Your #603 copy concern also goes away with the message.

_🤖 Posted by [Claude Code](https://claude.com/claude-code)_

src/react-query/queryClient.ts
```diff
@@ -63,19 +76,31 @@ export function errorHandler(rawErr: unknown) {
 	});
 }
 
+/**
+ * Every mutation error routes through the shared toast by default. A mutation can opt out
+ * — to render its own inline UI or redirect instead — with `meta: { skipGlobalErrorToast: true }`
+ * (e.g. the cloud login flow redirects an unverified user to the email-verification page, and
+ * sign-up puts an already-registered email on the field that has to change).
```
DavidCockerill: Small doc-drift: this line still describes the version of the change that was removed.

Both halves are now wrong — sign-up doesn't detect an already-registered email (it maps no statuses at all), and it doesn't set a *field* error (it sets `root` and renders above the submit button).

Worth correcting because this is the shared definition's doc *and* the only non-test line left in the diff that names the 409 semantics — a grep for "already registered" in six months lands here and gets the old story.

Something like: *"…and sign-up renders the failure in the form, above the submit button, instead of a toast that fades away from the inputs."*

src/features/auth/hooks/useSignUp.ts
```diff
@@ -22,5 +22,9 @@ export async function onSignUpSubmit(signUpCredentials: SignUpCredentials) {
 export function useSignUpMutation() {
 	return useMutation<SchemaUser, Error, SignUpCredentials>({
 		mutationFn: (loginData) => onSignUpSubmit(loginData),
+		// The sign-up form renders the failure inline, beside the inputs, instead of in a toast
+		// that fades away from them, so suppress the default global error toast for this
+		// mutation (see `SignUp`'s `onError`).
+		meta: { skipGlobalErrorToast: true },
```
DavidCockerill, line 28: The status-agnostic design is the right call, and I checked it forward against central-manager#630: its uniform response is a 2xx, so it lands on `onSuccess`, and `onSuccess` never reads the response body — it navigates using `userData.email` from the form. So "check your email in both cases" genuinely needs no Studio change. Good.

One gap in the body's claim that *"whatever replaces the 409 still reaches the user with no client change"*: it holds for error statuses, but not for a **bodyless 2xx**. `onSignUpSubmit` just above does:

```ts
const { data } = await apiClient.post('/User/', signUpCredentials);
if (data) { return data; } else { throw new Error('Something went wrong'); }
```

So a `202` with an empty body becomes a client-manufactured error, and the user reads "Something went wrong" inline instead of being sent to `/verifying`.

Nothing to change here — #630 hasn't settled its response shape yet — but worth a line on #630 so whoever implements it knows the response needs a body, or that this guard has to relax in the same change.

src/features/auth/SignUp.tsx
```diff
@@ -97,8 +102,18 @@ export function SignUp() {
...
 			},
+			// The sign-up mutation opts out of the global error toast (meta.skipGlobalErrorToast)
+			// and renders the failure in the form instead. RUM showed people resubmitting the
+			// same details two and three times before giving up (#1612): a toast that fades,
+			// away from the inputs, doesn't read as "this attempt failed". Deliberately status-
+			// agnostic — it reports whatever the server said rather than mapping specific codes.
+			onError: (error) => {
+				console.error(error);
+				const { title, description } = describeError(error);
+				setError('root', { type: 'server', message: description || title });
```
DavidCockerill: Sharing `describeError` between the toast and the inline text is the right move — but it's the *extraction* that's shared, not the composition, so one seam is left.

For an RFC 9457 body, dropping `title` is correct: `title` is the `code` (`InternalError`) and the description carries the sentence. But for a plain-string body containing a colon, `describeError` still runs the legacy split — `"Conflict: user already exists"` comes back as `title: 'Conflict'`, `description: ' user already exists'`. Inline then shows `" user already exists"`: first half gone, leading space kept, while the toast shows both parts as heading + body.

Low odds against a v5 central-manager, since problem-details responses set `splitTitleFromMsg = false` and never reach the split. Raising it only because "the two can't drift" is the stated goal and this is the one place they still can. If you want it closed: render `title` when it isn't the default `'Error'`, or `.trim()` the description.

DavidCockerill (approved):
Approving — and thanks for taking the scope cut rather than arguing the point. Clearing my earlier COMMENT.

**The disclosure is genuinely gone**, and I checked rather than taking the reply for it: grepping the whole diff for `409|already|exists|registered|enumerat` returns exactly two lines, neither a code path — a test fixture whose *purpose* is proving no status is special-cased, and one stale doc sentence (thread below). No status comparison, no message match, anywhere in `SignUp.tsx`, `useSignUp.ts` or the new `queryClient.ts` code. `isEmailAlreadyRegisteredError` is absent from both the diff and `origin/stage`.

Worth stating the nuance plainly, because it's what makes this the right shape rather than a cosmetic cut: CM still sends the literal string `User already exists`, and Studio renders it verbatim — so that text still reaches the browser today. But **Studio no longer decides anything**, which means central-manager#630 can close the oracle as a pure server edit with zero Studio work. At the previous head it would have needed a Studio revert. That's the whole thing the question was protecting.

**And the half that actually fixes #1612 is the half that stayed** — though I'd gently push back on the reasoning in the body. `isPending` isn't the fix: the RUM retries are ~5s apart and the server answers in 160ms–1.3s, so the button had long since re-enabled and never had a chance to block them. What makes people re-click is a toast appearing away from the inputs and then vanishing. The persistent in-form message above the submit button is what addresses that, and it's still here.

I paid particular attention to `describeError` being split out of `errorHandler`, since that's the same file #1598 landed the RFC 9457 handling in and a quiet behaviour change there would regress work nobody's looking at. It's clean, provably so in two independent ways: both hunks land on the *ends* of the function, so the problem-details block (`data.title` → description, `data.code` → title, `errorText(data.detail)` appended, `splitTitleFromMsg = false`) is untouched context between them — and `queryClient.test.ts` is unmodified, with all 16 cases including #1598's four still green through the delegation. Only behavioural move is `console.error` relocating into `errorHandler`.

No double-report and no silent drop: the sign-up path sets `skipGlobalErrorToast`, `mutationErrorHandler` returns early, the inline message renders — and the first test asserts `toast.error` was **not** called against the real handler, **imported rather than restated**. Your stated reason for declining Gemini's version survived your own rebuild, which is the detail I liked most here.

Three optional notes in threads, none blocking.

Separately, and not an ask on this PR: I confirmed the soft-delete duplicate-row path you flagged, and it's worse than duplicate rows. `searchByValue` then throws on **both** `Login` (`Login.js:67`) and `ForgotPassword` (`Password.js:37`), so the second signup *succeeds* and the account can then neither sign in nor reset its password — reachable by any user who deletes their account and comes back. Not tracked anywhere; #596/#603's §4d would prevent it going forward as a side effect but wouldn't repair rows already duplicated in prod. Worth its own issue.

_Reviewed by Claude Opus 5 for @DavidCockerill._

src/react-query/queryClient.ts
dawsontoth: Fixed in 75e46d1e — both halves were stale, and you're right that it was the last non-test line naming the 409 semantics. Now reads that sign-up renders the failure in the form, above the submit button, instead of a toast that fades away from the inputs.

src/features/auth/SignUp.tsx
dawsontoth: Good catch — reproduced it exactly (' user already exists', clause dropped and leading space kept). Closed in 75e46d1e by not losing the information rather than patching the display: `describeError` now also returns the pre-split `message`, the form renders that, and the toast keeps title/description, so it's still one extractor. Took that over rendering `title` when it isn't 'Error' because that would prepend the RFC 9457 `code` too, which you noted is right to drop. Test case asserts the exact regression string and fails against `description`.

src/features/auth/hooks/useSignUp.ts
dawsontoth: No code change, agreeing with your read: today's endpoint always returns a body so the guard is unreachable, and relaxing it now would be speculative against a response shape #630 hasn't settled. Thanks for checking the 2xx path forward — I'd only verified that no *error* shape needs a client change, and a bodyless 202 landing on `onSuccess` is exactly the gap. Leaving this thread open rather than resolving it, since the #630 note is the actual action and I need Dawson's go-ahead to comment cross-repo; I've flagged it to him. Also correcting the PR body, which overclaimed here.

dawsontoth: Filed as HarperFast/central-manager#636, referencing #630 — noting that #630's "indistinguishable by status code, body, and latency" acceptance is satisfied by an empty body, which is exactly the case that breaks the client. Recommendation there is giving the uniform response a truthy identical body rather than relaxing the guard, since the latter needs a matching studio PR landing in lockstep. Resolving here now that it's tracked where the work happens.

Issue #1591, opened 2026-07-30
Download application feature breaking page in Google Chrome


jjohnson-hdb:
## Summary

Attempting to download the source code from the applications leads to a "Something went wrong while displaying this webpage." screen and no download occurs. This was reproduced on multiple fabric v5 clusters.

## Repro

1. Sign in to `fabric.harper.fast`.
2. Enter a cluster
3. Applications > exp-engine > Download Application

Packaging loading indicator displays for ~20s
I see the pending `package_component` operation request resolve with a 200 and the bytes begin transferring, but before completion Chrome switches to "Something went wrong while displaying this webpage."  and the download is not received.

<img width="1161" height="674" alt="Image" src="https://github.com/user-attachments/assets/5648d9f3-cac3-4469-a9f9-f767e7176a87" />

---

## Update 2026-08-12 — cause, and the split with HarperFast/harper#2150

**Cause: renderer OOM in Studio, not a transport failure.** The bytes really did arrive — the tab died decoding them.

`package_component` hands back the whole tarball as base64 inside JSON, and `DownloadApplicationModal.tsx` then makes four more full-size copies of it in the renderer:

```ts
const bytes = Uint8Array.from(atob(response.payload), c => c.charCodeAt(0));
const file = new Blob([bytes], { type: 'application/gzip' });
```

axios buffers the whole JSON text → `JSON.parse` copies the base64 string → `atob()` copies again → `Uint8Array.from(str, callback)` runs a per-byte JS callback over hundreds of millions of elements on the main thread → `Blob` copies once more. Roughly 5× the tarball inside one renderer, so Chrome kills it. Both the response text and the parsed string are also subject to V8's 512 MiB string cap in the browser.

The server has the mirror-image problem (~4.7× peak, and a hard `ERR_STRING_TOO_LONG` ceiling at ~384 MiB of gzipped output). That half is now tracked as **HarperFast/harper#2150** — stream the tarball as a binary response instead of base64-in-JSON. The machinery already exists there (`streamPackagedDirectory`, and `serverHandlers.js` already pipes stream returns for `get_deployment_payload` / `get_backup`).

### What stays here in Studio

[...]

PR #1617, opened 2026-08-12
feat(applications): warn before a download too large for the browser

### LINKED ISSUE #2150: package_component: stream the tarball instead of returning it as base64 JSON
`package_component` returns the entire component tarball as a base64 string inside the JSON response body. That materializes the whole archive several times over — on the server, then again in the browser — and **hard-fails above ~384 MiB of compressed output** no matter how much RAM the node has.

Downstream symptom: HarperFast/studio#1591, where downloading an ~800 MB application transfers bytes fine and then kills the Chrome tab.

Everything needed to fix this already exists in this repo. `package_component` is simply on the wrong path.

## What it does today

`components/operations.js:354`:

```js
const payload = (await packageDirectory(pathToProject, req)).toString('base64');
return { project, payload };
```

`packageDirectory()` (`components/packageComponent.ts:192`) collects every gzip chunk into an array and `Buffer.concat`s it. For a tarball of N bytes:

| stage | resident |
| --- | --- |
| chunk array + `Buffer.concat` | 2N |
| `.toString('base64')` | +1.33N |
| Fastify `JSON.stringify` of the envelope | +1.33N |
| **peak** | **≈4.7N** |

All of it in the shared Harper process. On Fabric that process is multi-tenant, so one large download is an availability risk for every other tenant on the node.

And there's a hard ceiling, not just pressure — `.toString('base64')` throws `ERR_STRING_TOO_LONG` once the result would exceed V8's string cap:

```
$ node -e "const b=require('buffer'); console.log(b.constants.MAX_STRING_LENGTH)"
536870888        # 512 MiB → 384 MiB of gzipped tar is the ceiling
```

The browser then pays the same tax again: full JSON response text, a second copy out of `JSON.parse`, a third out of `atob()`, a per-byte `Uint8Array.from(s, c => c.charCodeAt(0))` on the main thread, and a `Blob` copy. That's the renderer OOM in studio#1591 — it's not a transport failure.

## Proposed change: return the packer's stream

Both halves are already built and already in use elsewhere.

- **The packer is already a stream.** `streamPackagedDirectory()` (`components/packageComponent.ts:115`) returns a `Readable` (tar-fs → gzip). Its docstring says it exists to dodge "the Node.js 2GB Buffer cap that the buffered variant runs into for large components." `packageComponent` just never calls it.
- **The response path already pipes streams.** `server/serverHelpers/serverHandlers.js:172` pipes any returned `Readable` carrying a `headers` Map, and honours a `preCompressed` flag so an already-gzipped body isn't re-compressed on the way out.
- **There is a direct precedent.** `handleGetDeploymentPayload()` (`components/deploymentOperations.ts:216`) does exactly this for deployment payloads, and its comment already states the reasoning verbatim: "Never base64 — payloads can be hundreds of MB and a base64 round-trip materializes multi-hundred-MB strings (V8's string cap is ~512MiB…)". `get_backup` uses the same mechanism.

Sketch:

```ts
const stream = streamPackagedDirectory(pathToProject, req);
const headers = new Map<string, string>();
headers.set('content-type', 'application/gzip');
// `project` is normalized through path.parse().name earlier in the handler,
// so it can't inject header content.
headers.set('content-disposition', `attachment; filename="${project}.tar.gz"`);
(stream as any).headers = headers;
(stream as any).preCompressed = true;   // already gzipped; don't recompress
return stream;
```

Memory becomes constant on both ends. Nothing is staged to disk, no temp archive, no retention lifecycle.

### Compatibility

`{ project, payload }` is a documented response shape, so the stream should be opt-in rather than a silent switch — `Accept: application/gzip`, or an explicit request field. Worth settling as part of this issue. Note the remote operation's only real consumer is Studio: the CLI's `package` alias packages *locally* for deploy (`bin/cliOperations.ts:14` imports `streamPackagedDirectory` directly) and never reads a remote `payload`.

`package_component` already requires `super_user` (`utility/operation_authorization.ts:287`), same posture as `get_deployment_payload`, so the streamed variant inherits the right gate.

### Secondary: a size probe

Studio wants to warn before packaging something absurd. It can already do that for the **default** mode — `get_components` returns per-file `size` and skips `node_modules`, which matches `skip_node_modules: true` exactly — but it has no way to size a package *with* `node_modules`.

`scanPackageDirectory()` (`components/packageComponent.ts:54`) already computes precisely this, and `getPackagedDirectorySize()` already wraps it. An `estimate: true` mode returning `{ project, total_size, dangling_symlinks }` would close the gap cheaply. Lower priority than the streaming fix.

## Constraint worth recording: Fabric Connect can't carry this

Central Manager's Fabric Connect path is a message-passing proxy, not an HTTP reverse proxy, and it caps bodies at 2 MB:

> Fabric Connect body size must be less than 2MB. Utilize Direct Connect for larger payloads.
>
> — `central-manager` `src/resources/hdbInstance/HDBInstance.js:383`

So a streamed download only works over a direct (Bearer operation-token) connection — the same split Studio's SSE work already models as `mode: 'direct' | 'proxy'`. Proxy-mode instances need an explicit error rather than a hang or a truncated file.

## Why not the alternatives

**Multi-file download** (one request per file, reassemble client-side) is strictly worse on every axis. The browser can't write a directory tree without `showDirectoryPicker()` (Chromium-only); a real application is thousands of files, each an extra round trip through `get_component_file`'s own base64-in-JSON; and it discards the symlink / mode / empty-dir fidelity that `streamPackagedDirectory` goes out of its way to preserve (`dereference`, the dangling-link skip, the directory-mode `map`). Reassembling into a single artifact means zipping in the browser, which pushes every byte back through JS memory — the exact thing being fixed. The one thing it buys is resumability, and HTTP range requests against a stream would buy that better if we ever need it.

**Staging a zip server-side** writes a second full copy to the node's disk and needs TTL, cleanup, and reclaim machinery. That's the deployment-payload system, which already exists — for the *upload* direction. Streaming needs none of it.

**"Don't do that"** is partly right, and is tracked separately in studio#1591: Studio should surface the size up front and warn, because `scp` genuinely is the better tool for pulling an 800 MB customer application. But that's a guardrail on top. It isn't a reason to leave a 384 MiB hard cap and a ~4.7× memory multiplier sitting in the operations API.

dawsontoth:
Addresses the Studio half of #1591. The server half is HarperFast/harper#2152 (closes HarperFast/harper#2150).

## Why the tab dies

The bytes arrive fine — the renderer dies decoding them. `package_component` returns the whole archive as base64 in JSON, and the modal then made four more full-size copies:

```ts
const bytes = Uint8Array.from(atob(response.payload), c => c.charCodeAt(0));
const file = new Blob([bytes], { type: 'application/gzip' });
```

axios buffers the response text → `JSON.parse` copies the base64 string → `atob` copies again → `Uint8Array.from(str, callback)` runs a JS callback **per byte** on the main thread → `Blob` copies once more. Roughly 5× the archive inside one renderer, and both strings are subject to V8's 512 MiB cap in the browser too. Hence "Something went wrong while displaying this webpage" with no explanation.

Streaming the archive end to end is harper#2150. **This PR does not consume the stream** — that needs a Harper build with #2152 in it to verify against, so it's deliberately left as the follow-up. What ships here works against every Harper version, today.

## What this does

**Says the size up front.** The modal now reads `About 5 MB across 159 files.` before you commit to anything. It's measured from the `get_components` tree Studio already has cached, so it costs no extra request. That tree omits `node_modules`, which makes the total an exact match for the modal's default (`skip_node_modules: true`) and a floor once the box is ticked.

**Warns past 100 MB** — or whenever node modules are included, where the measured total is only a floor — with the reason and the alternative, and relabels the button to "Download anyway":

> **This is a large download**
> Studio has to hold the whole archive in browser memory before it can be saved, so the tab may run out of memory and crash partway through.
> Copying the application off the host directly — `scp`, or the Harper CLI — is more reliable at this size.

Deliberately one tier, and deliberately not a block. A hard ceiling would have to be guessed from an uncompressed total, and a large tree of compressible source still downloads fine — so the decision stays with the user. This is the warning @jjohnson-hdb asked for in [#1591](https://github.com/HarperFast/studio/issues/1591#issuecomment-5270871335), and "don't do that" really is part of the right answer at 800 MB.

**Reports failures instead of hanging.** Two paths previously left the "Packaging..." toast spinning forever with no verdict: a failed mutation had no `onError` at all, and a decode throw escaped into react-query. Both now resolve the toast into a real error.

## The bug browser verification caught

`calculateRootEntries` rebuilds every node from scratch and was silently dropping the `size` the API sends. Against a real instance the modal first rendered **"At least 0 B across 159 files"** — 159 files found, every size gone. `size` was also undeclared on `APIFileEntry` even though Harper has long sent it. Both fixed; `mtime` is dropped the same way but nothing reads it, so I left it alone.

## Verification

Against the Anvils stage cluster on a real application, serving this worktree on :5173:

| | |
|---|---|
| Default | `About 5 MB across 159 files.` — no warning, button reads "Download" |
| Include Node Modules | `…, plus node modules.` — warning shown, button reads "Download anyway" |
| Unticked again | warning clears |
| Actual download | "Download ready!" — the happy path is intact through the new decode |

Checked in light and dark. Confirmed against the raw operation response that the server really does send per-file `size` (which is what identified the mapping bug rather than blaming the instance).

New tests: `projectPackageSize.test.ts` (9) and `DownloadApplicationModal.test.tsx` (8), covering the size text, the two warning triggers, the button relabel, that the request is never blocked, and both failure paths. Full suite green — 287 files, 2203 passing. (The 4 vitest "errors" in the summary come from `ToolCallGroup.test.tsx` and reproduce on a clean `stage` checkout.)

Decoding now allocates once and fills in a loop rather than the per-byte callback — still O(n) on top of `atob`, but strictly cheaper, and it matches `pemToDer` in `lib/crypto/envSecret.ts`.

## Follow-up

Once #2152 ships, Studio can request `stream: true` and pick the shape off the response `content-type`, so one code path works against both old and new instances. Two things to handle there: axios buffers (this needs `fetch`), and Fabric Connect caps bodies at 2 MB and can't stream at all, so proxy-mode instances need an explicit message — `resolveInstanceConnection` already reports `mode: 'direct' | 'proxy'`. harper#2150 also carries an `estimate` mode that would close the include-node_modules blind spot in the sizing above.

### DIFF
--- a/src/features/instance/applications/modals/DownloadApplicationModal.tsx
+++ b/src/features/instance/applications/modals/DownloadApplicationModal.tsx
@@ -1,25 +1,58 @@
+import { Alert, AlertDescription, AlertTitle } from '@/components/ui/alert';
 import { Button } from '@/components/ui/button';
 import { Dialog, DialogContent, DialogDescription, DialogHeader, DialogTitle } from '@/components/ui/dialog';
 import { Input } from '@/components/ui/input';
 import { Label } from '@/components/ui/label';
 import { useInstanceClientIdParams } from '@/config/useInstanceClient';
 import { useEditorView } from '@/features/instance/applications/hooks/useEditorView';
+import { LARGE_PACKAGE_BYTES, measureProjectPackage } from '@/features/instance/applications/lib/projectPackageSize';
 import { usePackageComponentMutation } from '@/integrations/api/instance/applications/packageComponent';
 import { attemptToRestoreFocus } from '@/lib/attemptToRestoreFocus';
 import { setWatchedValue, useWatchedValue } from '@/lib/events/watcher';
-import { DownloadIcon } from 'lucide-react';
-import { ChangeEvent, MouseEvent, useCallback, useState } from 'react';
+import { humanFileSize } from '@/lib/humanFileSize';
+import { AlertTriangleIcon, DownloadIcon } from 'lucide-react';
+import { ChangeEvent, MouseEvent, useCallback, useMemo, useState } from 'react';
 import { toast } from 'sonner';
+
+/**
+ * Decode the base64 archive Harper returns into bytes.
+ *
+ * Allocates once and fills in a loop rather than `Uint8Array.from(str, c => c.charCodeAt(0))`,
+ * which runs a JS callback per byte — hundreds of millions of them for a large application,
+ * all on the main thread. This is still O(n) memory on top of the `atob` copy; the real fix is
+ * a streamed archive (HarperFast/harper#2150). Until then the modal warns first — see
+ * `measureProjectPackage`.
+ */
+function decodeArchive(payload: string): Uint8Array<ArrayBuffer> {
+	const binary = atob(payload);
+	// Allocated over an explicit ArrayBuffer: a bare `new Uint8Array(n)` widens to
+	// ArrayBufferLike, which TS 5.7+ rejects as a BlobPart.
+	const bytes = new Uint8Array(new ArrayBuffer(binary.length));
+	for (let i = 0; i < binary.length; i++) {
+		bytes[i] = binary.charCodeAt(i);
+	}
+	return bytes;
+}
 
 export function DownloadApplicationModal() {
 	const { value: isModalOpen, trigger } = useWatchedValue('ShowDownloadApplicationModal', false);
 
 	const instanceParams = useInstanceClientIdParams();
-	const { openedEntry } = useEditorView();
+	const { openedEntry, rootEntries } = useEditorView();
 	const { mutate: packageComponent, isPending, isSuccess } = usePackageComponentMutation();
 	const actionStatus = isSuccess ? `Downloaded` : isPending ? `Downloading` : 'Download';
 
 	const [includeNodeModules, setIncludeNodeModules] = useState(false);
+
+	// Measured from the cached `get_components` tree, so it costs no request and is ready
+	// before the user commits to packaging.
+	const measured = useMemo(
+		() => measureProjectPackage(rootEntries, openedEntry?.project),
+		[rootEntries, openedEntry?.project],
+	);
+	// node_modules is absent from the tree, so with it included the real package is strictly
+	// larger than what we measured — and usually by a lot. Treat that as large regardless.
+	const isLarge = measured !== undefined && (measured.bytes > LARGE_PACKAGE_BYTES || includeNodeModules);
 
 	const includeNodeModulesChanged = useCallback((e: ChangeEvent<HTMLInputElement>) => {
 		setIncludeNodeModules(e.target.checked);
@@ -45,17 +78,34 @@ export function DownloadApplicationModal() {
 			},
 			{
 				onSuccess: (response) => {
+					// Decoding can fail on its own once the archive is big enough — a payload past
+					// V8's ~512 MB string cap throws here rather than in transport. Without this
+					// the throw escapes into react-query and the "Packaging..." toast spins forever.
+					let file: Blob;
+					try {
+						file = new Blob([decodeArchive(response.payload)], { type: 'application/gzip' });
+					} catch (error) {
+						toast.error(
+							`${openedEntry.project} was packaged, but is too large for the browser to save. `
+								+ `Copy it off the host directly instead.`,
+							{ id: toastId },
+						);
+						console.error('[harper] failed to decode downloaded application archive', error);
+						return;
+					}
 					toast.success('Download ready!', { id: toastId });
 
 					const element = document.createElement('a');
 					element.setAttribute('class', 'invisible absolute top-0');
-					const bytes = Uint8Array.from(atob(response.payload), c => c.charCodeAt(0));
-					const file = new Blob([bytes], { type: 'application/gzip' });
 					element.href = URL.createObjectURL(file);
 					element.download = `${openedEntry.project}.gz`;
 					document.body.appendChild(element);
 					element.click();
 					document.body.removeChild(element);
+				},
+				onError: (error) => {
+					// Previously unhandled, which left the loading toast spinning with no verdict.
+					toast.error(`Could not package ${openedEntry.project}: ${error.message}`, { id: toastId });
 				},
 			},
 		);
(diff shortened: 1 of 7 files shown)


src/features/instance/applications/modals/DownloadApplicationModal.tsx
```diff
@@ -1,26 +1,59 @@
...
 
+	// Measured from the cached `get_components` tree, so it costs no request and is ready
+	// before the user commits to packaging.
+	const measured = useMemo(
+		() => measureProjectPackage(rootEntries, openedEntry?.project),
+		[rootEntries, openedEntry?.project],
+	);
+	// node_modules is absent from the tree, so with it included the real package is strictly
+	// larger than what we measured — and usually by a lot. Treat that as large regardless.
+	const isLarge = measured !== undefined && (measured.bytes > LARGE_PACKAGE_BYTES || includeNodeModules);
```
kriszyp: `measureProjectPackage` explicitly marks the result inexact when any file lacks `size`, but this condition still treats a partial total below 100 MB as safe. For example, an older instance returning no sizes for an 800 MB project produces `{ bytes: 0, exact: false }`, shows “At least 0 B,” and gives no warning before reproducing the tab crash this change targets. Please give inexact/unknown measurements their own caution state (and test it), rather than using the partial byte total to suppress the warning.

— KrAIs (GPT-5)

kriszyp (approved):
Sounds good!

src/features/instance/applications/modals/DownloadApplicationModal.tsx
dawsontoth: Good catch — fixed in ee0c6576. Replaced the boolean with `packageCaution()`, which returns `'large' | 'unmeasured' | undefined`: an inexact or absent measurement now gets its own caution state ("This download's size is unknown") instead of letting a partial total suppress the warning, and unmeasured outranks large since "we don't know" is the honest message either way. Also stopped rendering "0 B" — an unmeasured project now reads "159 files of unreported size". Covered by four new tests that all fail without the guard, plus a live-cluster check with sizes stripped from the response, which reproduced your exact scenario and now warns.

PR #1623, opened 2026-08-14
feat(auth): gate staff features on staffPermissions instead of role names

### LINKED ISSUE #1622: Staff features gate on hardcoded role names instead of staffPermissions
## Problem

`GET /User/current` returns a `staffPermissions` array describing exactly what a staff account may do, but the UI never reads it. Staff gating is hardcoded to role names in two places:

- `isAdminMode()` — `fabricRole === 'fabric_admin' || fabricRole === 'super_user'` ([useAuth.ts:46](../blob/stage/src/hooks/useAuth.ts#L46))
- `isFabricAdmin()` — `fabricRole === 'fabric_admin'` ([useAuth.ts:54](../blob/stage/src/hooks/useAuth.ts#L54))

This is wrong in both directions:

**Narrower staff roles see nothing.** Accounts holding `fabric_support`, `fabric_readonly`, or `fabric_staff` never enter admin mode, so they get no "All Orgs" toggle, no entity-id jump, no `/admin` section — even for pages whose underlying API calls their permissions allow (e.g. the all-organizations listing, which every staff role may read).

**Admin mode over-promises.** The `usePermissions` hooks return blanket `{create/update/remove/view: true}` in admin mode, but the API has since withheld several permissions from `fabric_admin` (org delete, cluster delete, credential reads). The UI renders buttons that 403 on click.

The `fabricRole` type union in `api.patch.d.ts` also predates the narrower roles.

## Affected surfaces

| Surface | Current gate | Should gate on |
|---|---|---|
| All Orgs toggle + id jump (`features/organizations/index.tsx`) | `isAdminMode` | `org:read` |
| OrgCard "Add Coupon" | `isAdminMode` | `billing:write` |
| `useOrganizationPermissions` admin branch | `isAdminMode` → all true | `org:update` / `org:delete` |
| `useOrganizationRolePermissions` admin branch | `isAdminMode` → all true | `role:read/create/update/delete` |
| cluster + instance permission hooks admin branch | `isAdminMode` → all true | `cluster:*` / `instance:*` |
| DatabaseTableView "Cleanup Orphan Blobs" | `isAdminMode` | `instance:update` |
| `/admin` section (AdminShell, Navbar) | `isFabricAdmin` | any admin-page permission; per page: Notifications `systemStatus:write`, Regions `region:read`, API Token `apiToken:create` |

## Fix

- Type `staffPermissions` on `User` and widen `fabricRole`.
- Add a `hasStaffPermission(user, permission)` helper; when the API predates `staffPermissions` (field absent), fall back to the current role-name check so an older backend keeps today's behavior.
- Replace each surface's role-name gate with its specific permission, per the table.
- AdminShell items each declare their permission; the shell admits anyone holding at least one and hides the rest.

DavidCockerill:
Closes #1622

## What

`GET /User/current` returns a `staffPermissions` array describing exactly what a staff account may do, but the UI never read it — staff gating was hardcoded to `fabricRole === 'fabric_admin'` (plus `super_user`). That was wrong in both directions: the narrower staff roles saw no staff features at all, and admin mode rendered actions the API has since withheld from `fabric_admin` (org delete, cluster delete), which 403 on click.

This replaces every role-name gate with the specific permission its surface needs, via a new `hasStaffPermission(user, permission)`:

| Surface | Now gated on |
|---|---|
| All Orgs toggle, entity-id jump, OAuth-lock bypass | `org:read` |
| OrgCard "Add Coupon" | `billing:write` |
| OrgCard "Delete" + org permissions hook | `org:delete` / `org:update` (unioned with the member role) |
| Org roles hook | `role:read/create/update/delete` |
| Cluster hook | `cluster:read/create/update/delete` |
| Instance hook | `instance:read/update`; create/remove follow the cluster grants (instances are added/removed through cluster updates) |
| "Cleanup Orphan Blobs" | `instance:update` |
| /admin Notifications | `systemStatus:write` |
| /admin Regions | `region:read` (page), `region:write` (create/edit buttons) |
| /admin API Token | `apiToken:create` |

## Behavior notes

- **`usePermissions` hooks union, not shadow**: a staff account that is also an org member keeps whatever its member role grants, and vice versa — matching the API's own semantics.
- **/admin opens per-page**: the rail shows only the pages the account holds, holding any of them grants entry, and deep links to unheld pages bounce to the first visible one. The `/admin` index redirect is unchanged.
- **super_user never sees the API Token page**, even though it holds the permission: token minting requires the Google SSO session staff sign-ins have, and super_user may password-login — the mint would 403. This preserves the invariant the old `isFabricAdmin` gate encoded, but per-page instead of hiding the whole section.
- **Older backends keep today's behavior**: when `staffPermissions` is absent from the response, `hasStaffPermission` falls back to the legacy role-name check, so deploying this UI ahead of the API changes nothing.
- `isAdminMode` / `useAdminMode` / `isFabricAdmin` are gone — nothing references role names for gating anymore.

## Testing

- New: `hasStaffPermission` unit tests (granted/withheld/customer/local/legacy-fallback), `/admin` visibility per role incl. the super_user carve-out, and cluster/instance hook mapping + union tests.
- Full suite: 293 files, 2268 passed / 11 skipped — same as `stage` baseline plus the new tests. `tsc`, oxlint, and dprint clean.

### DIFF
--- a/src/features/admin/regions/index.tsx
+++ b/src/features/admin/regions/index.tsx
@@ -5,6 +5,7 @@ import { Table, TableBody, TableCell, TableHead, TableHeader, TableRow } from '@
 import { RegionFormModal } from '@/features/admin/regions/components/RegionFormModal';
 import { formatOrgLabel, getOrganizationsQueryOptions } from '@/features/admin/regions/queries/getOrganizations';
 import { getRegionsQueryOptions } from '@/features/admin/regions/queries/getRegions';
+import { useStaffPermission } from '@/hooks/useAuth';
 import { AdminRegion } from '@/integrations/api/api.patch';
 import { useQuery } from '@tanstack/react-query';
 import { PencilIcon, PlusIcon } from 'lucide-react';
@@ -32,6 +33,9 @@ export function RegionsIndex() {
 	const [modalOpen, setModalOpen] = useState(false);
 	const [editing, setEditing] = useState<AdminRegion | null>(null);
 	const [search, setSearch] = useState('');
+	// The page itself only needs region:read; creating/editing posts to the
+	// region:write-gated endpoints.
+	const canWriteRegions = useStaffPermission('region:write');
 
 	const orgNameById = useMemo(
 		() => new Map((orgResult?.organizations ?? []).map((o) => [o.id, o.name])),
@@ -70,10 +74,12 @@ export function RegionsIndex() {
 						customers; otherwise it's available to everyone.
 					</p>
 				</div>
-				<Button variant="submit" onClick={openCreate} className="shrink-0">
-					<PlusIcon />
-					Create region
-				</Button>
+				{canWriteRegions && (
+					<Button variant="submit" onClick={openCreate} className="shrink-0">
+						<PlusIcon />
+						Create region
+					</Button>
+				)}
 			</div>
 
 			<div className="mt-6">
@@ -124,15 +130,16 @@ export function RegionsIndex() {
 														<RegionScope organizationIds={region.organizationIds} orgNameById={orgNameById} />
 													</TableCell>
 													<TableCell className="text-right">
-														<Button
-															variant="ghost"
-															size="icon"
-															aria-label={`Edit ${region.id}`}
-															onClick={() =>
-																openEdit(region)}
-														>
-															<PencilIcon />
-														</Button>
+														{canWriteRegions && (
+															<Button
+																variant="ghost"
+																size="icon"
+																aria-label={`Edit ${region.id}`}
+																onClick={() => openEdit(region)}
+															>
+																<PencilIcon />
+															</Button>
+														)}
 													</TableCell>
 												</TableRow>
 											))}
--- a/src/features/organizations/index.tsx
+++ b/src/features/organizations/index.tsx
@@ -13,7 +13,7 @@ import {
 	ALL_ORGANIZATIONS_PAGE_SIZE,
 	getAllOrganizationsQueryOptions,
 } from '@/features/organizations/queries/getAllOrganizations';
-import { useAdminMode } from '@/hooks/useAuth';
+import { useStaffPermission } from '@/hooks/useAuth';
 import { useDebounce } from '@/hooks/useDebounce';
 import { useSessionStorage } from '@/hooks/useSessionStorage';
 import { detectEntityId } from '@/lib/string/entityId';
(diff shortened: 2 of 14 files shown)


dawsontoth (approved):
nice

src/features/admin/regions/index.tsx
```diff
@@ -32,6 +33,9 @@ export function RegionsIndex() {
 	const [modalOpen, setModalOpen] = useState(false);
 	const [editing, setEditing] = useState<AdminRegion | null>(null);
 	const [search, setSearch] = useState('');
+	// The page itself only needs region:read; creating/editing posts to the
+	// region:write-gated endpoints.
+	const canWriteRegions = useStaffPermission('region:write');
```
kriszyp, line 38: A user with only `region:read` is now intentionally admitted to this page, but the component unconditionally runs `getOrganizationsQueryOptions()`, whose `/Admin/Organization/` request is gated by `org:read`. The explicit `fabric_readonly` test case therefore gets a global 403 toast on page load, loses organization names, and cannot filter scopes by name. Please either require `org:read` for this page or conditionally enable the organization query and provide a read-only fallback for region-only users; the modal should apply the same rule. — KrAIs (GPT-5)

src/features/organizations/index.tsx
```diff
@@ -56,15 +58,15 @@ export function OrganizationsIndex() {
 
 	// A pasted organization/cluster id is resolved by id server-side rather than
 	// matched as a title (see getAllOrganizationsQueryOptions). This works for
-	// any fabric admin regardless of the "All Orgs" toggle, so support can jump
+	// any staff account regardless of the "All Orgs" toggle, so support can jump
 	// straight to an org from an id copied out of a log or ticket.
 	const searchEntityId = useMemo(() => {
-		if (!isAdminMode) {
+		if (!canReadAllOrgs) {
```
kriszyp, line 64: This single `org:read` check also enables pasted cluster-ID searches, but that branch first calls `/Cluster/{id}` before looking up the organization. An account with `org:read` but no `cluster:read` will therefore get a 403 instead of the promised exact-ID lookup. Please gate cluster-ID detection on `cluster:read` as well, or resolve the cluster through an organization-admin endpoint authorized by `org:read`. — KrAIs (GPT-5)

src/hooks/usePermissions.ts
```diff
@@ -119,10 +138,19 @@ export function useOrganizationClusterInstancePermissions(orgId?: string, cluste
...
-	if (isAdminMode(user)) {
-		return { create: true, remove: true, update: true, view: true };
-	}
+	const member = memberClusterInstancePermissions(user, orgId, clusterId);
+	return {
+		// Instances are added/removed through cluster updates (there are no
+		// instance:create / instance:delete permissions), so those verbs follow
+		// the cluster grants.
+		create: hasStaffPermission(user, 'cluster:update') || member.create,
+		remove: hasStaffPermission(user, 'cluster:delete') || member.remove,
```
kriszyp, line 147: The adjacent comment correctly says adding and removing instances happens through cluster updates, but `remove` is mapped to `cluster:delete`. That denies instance removal to staff who can update a cluster and grants it to staff who may delete a cluster but cannot update it; the latter operation would then be rejected by the update endpoint. Please map both instance `create` and `remove` to `cluster:update` and add the symmetric test assertion. — KrAIs (GPT-5)

kriszyp (commented):
Sounds good. A few codex suggestions if you are interested, while this sits in the queue.

PR #1628, opened 2026-08-14
feat(roles): surface operation-level grants (permission.operations) in the roles/users UI

dawsontoth:
Closes #1627.

## What this adds

Harper 5.0+ roles accept a `permission.operations` allowlist (deny-unlisted, with explicit listing acting as a scoped super_user delegation). Studio's role editor now surfaces it:

- **Role editor (`EditRoleModal`)**: a "Restrict operations" section above the JSON editor — predefined group toggles (`read_only` / `admin_read` / `standard_user` / `agent`), a categorized operation picker with `super_user` badges on delegating grants, chips for the selected list, and an effective-operations summary (groups expanded). The section is a *lens over the JSON text* — the JSON stays the single source of truth, and structured writes float `operations` to the top of the document (right after the `super_user`-family flags) so it's visible without scrolling.
- **Free-text escape hatch** for component-registered operations (5.2+, `server.registerOperation`): an exact-miss filter that looks like an operation name can be granted as a custom entry, case preserved.
- **Roles table**: an "Operations" column — the effective count, "—" when unrestricted, or a flag when the value is malformed or the role is elevated (see below).
- **User add/edit forms**: a one-line note when the selected role carries an allowlist.
- **Version gating**: the section only renders on Harper ≥ 5.0.0-alpha.1; per-operation `addedIn` floors (5.1 deployments/agent ops, 5.2 secrets/env/backups) filter the picker; the `agent` group needs ≥ 5.2.0-alpha.1.

The catalog (`operationsCatalog.ts`) is static, extracted from Harper source at tags v5.0.0 / v5.1.0 / v5.2.2, because core has no discovery operation for grantable names (the issue's open question). Unknown names in existing role JSON are preserved and rendered, never dropped. Non-delegable ops (secrets store, managed backups, `get_deployment_payload` — handlers self-enforce super_user) are never offered.

## Where to look

- `src/integrations/api/localRolePermission.ts` — reserved-key model. Key identity (not value shape) decides what's a database record vs a reserved key, which is what protects a database literally named `operations` from clobbering (or being clobbered by) the allowlist in `defaultCalculator`/`preparePermissionForSave`.
- `EditRoleModal.tsx` — the JSON↔picker sync. Monaco applies programmatic `value` updates without firing `onChange` (see AGENTS.md), which is what makes the lens loop-free. The save-path strip loop round-trips an *array-shaped* `operations` for unconditionally-elevated roles rather than deleting what the author typed, while a non-array value is still stripped like any table permission.
[...]

Issue #1627, opened 2026-08-14
Surface role operation-level grants (permission.operations) in the roles/users UI


heskew:
Harper's roles support operation-level RBAC that Studio's role/user editor doesn't surface yet. Requested by @dawsontoth so the UI can expose it. This outlines the full capability surface and which Harper versions support each piece, verified against the Harper source and release tags.

## The capability

A role's `permission` object accepts an optional `operations` array — an allowlist of Operations API operation names (snake_case) and/or predefined group names:

```jsonc
{
  "operation": "add_role",
  "role": "ci_deploy",
  "permission": {
    "operations": ["deploy_component", "get_status"]
    // table CRUD permissions may appear alongside as usual
  }
}
```

Semantics (two-gate check, first thing in authorization):

1. **Gate 1 — hard allowlist.** When `operations` is present, any operation *not* listed is denied, before table CRUD permissions are even considered. Omitting `operations` entirely = current behavior (no allowlist).
2. **Gate 2 — scoped super_user delegation.** An operation that normally requires `super_user`, when explicitly listed, is treated as a deliberate admin grant and allowed *without* super_user. This is what makes e.g. a deploy-only CI user possible: `operations: ["deploy_component", "get_status"]` on a non-SU role yields a user that can run exactly those two operations and nothing else.
3. Non-SU operations that pass gate 1 still go through normal table CRUD permission checks.

**Predefined groups** (usable in the array alongside individual names, expanded server-side):

| Group | Contents (summary) |
|---|---|
| `read_only` | searches, SQL, describes, user/analytics reads |
| `admin_read` | SU-only reads: `get_configuration`, logs, component views |
| `standard_user` | `read_only` + full DML + bulk load (no DDL, no user/role mgmt) |
| `agent` | agent session operations (`agent_prompt`, `get_agent_session`, …) |

**Write-time validation.** `add_role` / `alter_role` reject an `operations` array containing an unknown name — the server validates, but a UI picker still needs the list of valid names (see open question below).

**Not delegable by design.** A small set of SU-only operations self-enforce `super_user` in their handlers and cannot be granted through the allowlist: the secrets-store operations, managed-backup operations, and `get_deployment_payload`. A UI should not offer these as grantable.

**Impersonation.** Inline role assertions (impersonation with a role in the request body) accept and validate the same `operations` field.

## Version support

| Harper version | What's available |
|---|---|
[...]

PR #1631, opened 2026-08-17
feat(permissions): apply the operations allowlist to table read/write gating

dawsontoth:
Studio's table permission checks ignored `permission.operations`, so a role whose allowlist excludes a
data operation still saw Add Records, Import Data, and the inline edit/delete affordances — each of
which the server denies at gate 1, before the role's table CRUD grants are consulted.

`checkSchemaTablePermission` and `useInstanceSchemaTableAttributePermission` now consult the allowlist
first, through a new pure `src/hooks/checkOperationPermission.ts`.

Stacks on #1628 (base `role-operations-allowlist`), which added the catalog and the roles-UI editor.

## For the human reviewer

Read [checkOperationPermission.ts](src/hooks/checkOperationPermission.ts) first — its header states the
invariant the whole change rests on — then [checkSchemaTablePermission.ts](src/hooks/checkSchemaTablePermission.ts)
for the gate order, then the call sites.

**The manage and browse-manage hooks are deliberately left alone, which is the opposite of what the
task asked for.** #1627 states the allowlist "denies unlisted operations even for super_user roles," so
the obvious change was to gate `useInstanceManagePermission` (Users, Roles, Config, Logs, Status) and
`useInstanceBrowseManagePermission` (DDL, app editors). I implemented exactly that, then read
`verifyPerms` and reverted it:

```
utility/operation_authorization.ts:553   if (isSuperUser && !isSuSystemOperation) return null;   // "admins can do (almost) anything"
utility/operation_authorization.ts:559   structure_user + create/drop database   → return null
utility/operation_authorization.ts:564   structure_user + STRUCTURE_USER_OPS     → return null
utility/operation_authorization.ts:589   ── the operations allowlist gate lives here ──
```

**Measured, not inferred** (see Verification): a super_user with `operations: []` — the strongest
possible test, since a gate that applied at all would deny everything — is still ALLOWED insert,
search, restart and get_configuration. A restricted-super_user's admin actions do not 403, so gating
that chrome would have hidden UI that works. #1628's `7eff23fc` reached the same conclusion
independently. That leaves table DML as the only surface gate 1 can deny, which is what this gates.

Decisions worth questioning:

- **Add Records and Import Data no longer share a flag, and each import method is gated by the
  operation it issues.** Add Records is strictly `insert`. Import Data's methods are modeled as the
  *paths* they can take, because `sample` and `file` stay ambiguous until submit — a bundled dataset
  bulk-loads while random records insert, and a `.json` upload inserts while any other extension
  bulk-loads. A method is offered while either path is open; the resolved source is re-checked at
  submit, where the choice is finally known. CSV paths require **`get_job`** alongside the load,
  because [importData.ts](src/integrations/api/instance/database/importData.ts) polls the job — a CSV
  grant without it leaves a committed load reporting failure, and the obvious retry duplicates rows.
- **The gate is version-blind, by shape rather than by version.** `getOperationsAllowlist` only says
  yes to an array of strings, and what a pre-5.0 instance puts under that key is a database record
  (`permissionsTranslator` writes one for any upgraded v4 role owning a database named `operations`).
  Asking the instance version instead would mean a query per check on a hook that does not receive
  this module's `entityId` — the bug #1628's `b7c01cf4` just removed.
- **The gate simulation does not reuse `expandEffectiveOperations`.** That helper folds alias spellings
  together to describe a role's effective reach; the server's gate does not fold, so a lone
  `search_by_id` grant must stay the dead entry it is. Simulating the gate is this module's whole job,
  so it expands verbatim and canonicalizes only the operation being checked.
- **Not reusing `isElevatedRole`** (from #1628; counts `structure_user` and `cluster_user`). It answers
  whether the allowlist is unenforceable *anywhere* for the role. For DML both are still gated:
  `STRUCTURE_USER_OPS` is create/drop table and attribute only, and `cluster_user` appears nowhere in
  `operation_authorization.ts`. Measured: `{structure_user: true, operations: ['read_only']}` → insert
  DENIED.
- **Malformed allowlists restrict nothing** (fail-open), because a denial that can't be proven must not
  hide UI. This differs from #1628's editor, which surfaces a malformed value as active — different
  jobs: it warns, this hides.
- **`read` gating is live but inert today** — no production call site passes `'read'`, and the
  attribute hook has no callers. Mapped for symmetry and covered by tests.
- **The database-scoped Import launcher** has no table to check a grant against (it creates one), so it
  is gated on the operations half alone — otherwise a `structure_user` with an empty allowlist could
  start a load denied *after* `create_table` succeeded, leaving an empty table behind.

## Verification

- **End-to-end route: executed `verifyPerms` / `verifyPermsAST` directly** against the Harper checkout
  (`main`, mocha, the same harness as `unitTests/utility/operation_authorization.test.js`), because the
  premise this PR rests on is not covered by any existing test — Harper's allowlist suite sets
  `super_user: false` everywhere. Throwaway probes, since deleted (Harper checkout clean):

  | role + allowlist | operation | verdict |
  |---|---|---|
  | `super_user: true`, `operations: []` | `insert`, `search_by_conditions`, `restart`, `get_configuration` | **ALLOWED** (all four) |
  | `super_user: true`, `operations: ['read_only']` | `insert`, `restart` | **ALLOWED** |
  | `super_user: false`, `operations: []` | `insert` | **DENIED** |
  | `structure_user: true`, `operations: ['read_only']` | `insert` | **DENIED** |

  Two further probes settled @kriszyp's review claims, and both confirmed them — see the PR comment for
  detail. They are Harper-side, not defects in this diff: SQL dispatches through `verifyPermsAST`, which
  has no allowlist gate (`operations: []` still permits SQL writes); and 9 of 23 sampled picker names
  can never match the gate because their authorization entry omits `api_name` (`deploy_component` among
  them, so #1627's own deploy-only CI user is not buildable today). #1628 tracks the second as
  HarperFast/harper#2175 with a `GATE_INERT_OPERATIONS` set; the SQL bypass still wants an issue.
- The Studio-side tests prove an allowlist shows or hides the right control. What remains unproven from
  Studio: a full round-trip against a live 5.x instance carrying such a role.
- `src/hooks/checkOperationPermission.test.ts` (new): absent / malformed / empty / exact / unknown /
  group / both alias directions, a database record under `operations` reading as no allowlist, expansion
  caching, per-action mapping, `structure_user` and `cluster_user`, per-method and per-source import
  capability, and a mapping-integrity test asserting every mapped name is a canonical, non-group catalog
  entry. That last one earned its keep: an earlier draft mapped `list_certificates` and `list_ssh_keys`,
  which harper-pro registers at runtime and the catalog therefore lacks, so they would never have matched.
- `checkSchemaTablePermission.test.ts` extended: allowlist over a CRUD grant, over `structure_user`, the
  super_user pass-through, and the Add-Records-vs-Import divergence.
- `TableContextMenuItems.test.tsx` and `ImportDataModal.permissions.test.tsx` (both new) mock only the
  auth store and the router, so they run the real allowlist → hook → rendered-control chain: entries
  hidden for a read-only allowlist, Import Data surviving a CSV grant while Add Records drops, Import
  Data hidden when a CSV load has no `get_job`, the URL method dropped for an insert-only role, and the
  default selection falling back to an allowed method.
- **Full gate, matching `verify-pr.yaml` exactly** — `tsc -b` ✓, `pnpm test:coverage` ✓ (**301 files /
  2369 passed**, 11 skipped), `pnpm lint` ✓, `pnpm build` ✓; `dprint` no-op. The one uncovered branch in
  the new module is the `?? []` guard TypeScript makes unreachable.
- `test:e2e:docker` is **not** part of CI's PR gate and its specs cover auth/signup/org-users only —
  nothing in the instance browse UI — so it cannot exercise this change.

## Review coverage

Authored by Claude Opus 5. Outside coverage via the bundled pre-push CLI (`--author claude`), plus a
human review; every round changed the design:

| round | lens | outcome |
|---|---|---|
| plan | `gpt-5.6-sol` (codex) | rejected the aggregate manage/browse-manage booleans; the narrowing to DML came from this |
| impl | `gpt-5.6-sol` (codex) | `CHANGES` — the shared insert flag conflated Add Records with Import Data |
| impl `--full` | `gpt-5.6-sol` (codex) | `CHANGES` — CSV imports missing `get_job`; the ungated third Import launcher |
| impl `--full` | gemini via `agy` (default model) | crash risk on an off-type action lookup; no component tests |
| post-push | gemini-code-assist (CI bot) | wanted `csv_file_load`/`import_from_s3` added — declined with rationale, `ImportSource` issues neither |
| post-push | **@kriszyp** (with KrAIs/GPT-5) | the two server findings above, plus per-source import gating — implemented |

`cursor-grok` ✗ (pruned by policy), `cursor-composer` ✗ (`cursor-agent` not installed), Harper domain
adjudication ✗ (exit 1) — so the decision list above is mine, not an adjudicated ledger.

Rebased twice onto a moving #1628 (`b7c01cf4`, `61d068be`), both times with semantic merges rather than
textual ones: the version-blind reversal and the `expandEffectiveOperations` split above both came out of
those. No cross-model round has run against the current head; the changes since the last reviewed SHA are
the import-method gating, the version-blind revert, and tests.

## Docs

No companion PR to HarperFast/documentation: this makes Studio's UI agree with server behavior that
already ships, and adds no API, config, CLI, or default. The allowlist feature's own documentation
belongs with #1628.

<sub>Human-Review-Need: 4 @ eae63ad4f775</sub>

### DIFF
--- a/src/features/instance/config/roles/operations/operationsCatalog.ts
+++ b/src/features/instance/config/roles/operations/operationsCatalog.ts
@@ -349,6 +349,11 @@ export function getOperationInfo(name: string): GrantableOperation | undefined {
 	return catalogByName.get(name);
 }
 
+/** The authorization entry a name resolves to; aliases collapse onto their canonical operation. */
+export function canonicalOperationName(name: string): string {
+	return catalogByName.get(name)?.aliasOf ?? name;
+}
+
 export function isOperationGroupName(name: string): boolean {
 	return groupsByName.has(name);
 }
--- a/src/features/instance/databases/modals/ImportDataModal.tsx
+++ b/src/features/instance/databases/modals/ImportDataModal.tsx
@@ -33,13 +33,15 @@ import {
 } from '@/features/instance/databases/functions/generateRandomRecords';
 import { parseJsonRecords } from '@/features/instance/databases/functions/parseJsonRecords';
 import { sampleDatasets } from '@/features/instance/databases/sampleDatasets';
+import type { ImportMethod } from '@/hooks/checkOperationPermission';
+import { useInstanceImportCapabilities } from '@/hooks/usePermissions';
 import { InstanceDatabaseMap } from '@/integrations/api/api.patch';
 import { databaseNameSchema } from '@/integrations/api/instance/database/databaseNameSchema';
 import { ImportSource, useImportDataMutation } from '@/integrations/api/instance/database/importData';
 import { tableNameSchema } from '@/integrations/api/instance/database/tableNameSchema';
 import { zodResolver } from '@hookform/resolvers/zod';
 import { CloudUploadIcon, FileUpIcon, GlobeIcon, InfoIcon, LoaderCircleIcon, PackageIcon } from 'lucide-react';
-import { useEffect, useState } from 'react';
+import { useEffect, useMemo, useState } from 'react';
 import { useForm } from 'react-hook-form';
 import { toast } from 'sonner';
 import { z } from 'zod';
@@ -131,12 +133,22 @@ export function ImportDataModal({
 	const { mutate: importData, isPending } = useImportDataMutation();
 	const [selectedFileName, setSelectedFileName] = useState<string | null>(null);
 
+	const importCapabilities = useInstanceImportCapabilities();
+	// Offer only methods the role can actually run; a method whose every path is denied would fail at
+	// submit with a 403 the user can do nothing about.
+	const availableMethods = useMemo(
+		() => importMethods.filter(({ value }) => importCapabilities.methods[value]),
+		[importCapabilities],
+	);
+	// With a table already in context (table toolbar) the likely intent is loading your own data;
+	// without one (sidebar, empty database) lead with samples -- then fall back to whatever is allowed.
+	const preferred: ImportMethod = tableName ? 'file' : 'sample';
+	const defaultMethod = importCapabilities.methods[preferred] ? preferred : (availableMethods[0]?.value ?? preferred);
+
 	const form = useForm({
 		resolver: zodResolver(ImportDataFormSchema),
 		defaultValues: {
-			// With a table already in context (table toolbar) the likely intent is loading
-			// your own data; without one (sidebar, empty database) lead with samples.
-			method: tableName ? ('file' as const) : ('sample' as const),
+			method: defaultMethod,
 			database: databaseName || '',
 			table: tableName || '',
 			datasetId: '',
@@ -153,7 +165,7 @@ export function ImportDataModal({
 	useEffect(() => {
 		if (isModalOpen) {
 			reset({
-				method: tableName ? 'file' : 'sample',
+				method: defaultMethod,
 				database: databaseName || '',
 				table: tableName || '',
 				datasetId: '',
@@ -164,7 +176,7 @@ export function ImportDataModal({
 			});
 			setSelectedFileName(null);
 		}
-	}, [isModalOpen, databaseName, tableName, reset]);
+	}, [isModalOpen, databaseName, tableName, defaultMethod, reset]);
 
 	const method = form.watch('method');
 	const datasetId = form.watch('datasetId');
@@ -265,6 +277,18 @@ export function ImportDataModal({
 			source = { kind: 'csv-url', url: values.url };
 		}
 
+		// `sample` and `file` only resolve to a concrete operation here (a bundled dataset bulk-loads,
+		// random records insert; a .json upload inserts, any other extension bulk-loads), so this is the
+		// first point the allowlist can be checked against what will actually be sent.
+		if (!importCapabilities.allowsSource(source.kind)) {
+			form.setError('method', {
+				message: source.kind === 'json-records'
+					? 'This role cannot insert records. Choose a CSV source instead.'
+					: 'This role cannot run CSV loads. Choose a JSON file or random sample data instead.',
+			});
+			return;
+		}
+
 		importData(
 			{
 				database,
@@ -313,7 +337,7 @@ export function ImportDataModal({
 									<FormLabel>Select Import Method</FormLabel>
 									<FormControl>
 										<RadioGroup value={field.value} onValueChange={field.onChange} className="gap-3">
-											{importMethods.map(({ value, title, description, Icon }) => (
+											{availableMethods.map(({ value, title, description, Icon }) => (
 												<FormLabel
 													key={value}
 													htmlFor={`import-method-${value}`}
--- a/src/features/instance/databases/modals/ImportDataModal.tsx
+++ b/src/features/instance/databases/modals/ImportDataModal.tsx
@@ -277,6 +277,14 @@ export function ImportDataModal({
 			source = { kind: 'csv-url', url: values.url };
 		}
 
+		// The launcher checked the table it was opened from; the target is editable, so an existing table
+		// chosen here needs its own insert grant. A table that does not exist yet has none to check --
+		// creating it is the browse-manage authority the launcher already required.
+		if (tableExists && !importCapabilities.allowsDestination(database, table)) {
+			form.setError('table', { message: `This role cannot insert into "${table}".` });
+			return;
+		}
+
 		// A file's extension decides its operation, so `file` is the one method whose source is unknown
 		// until here; the sample picker offers only granted sources already.
 		if (!importCapabilities.allowsSource(source.kind)) {
--- a/src/features/instance/databases/modals/ImportDataModal.tsx
+++ b/src/features/instance/databases/modals/ImportDataModal.tsx
@@ -277,9 +277,8 @@ export function ImportDataModal({
 			source = { kind: 'csv-url', url: values.url };
 		}
 
-		// `sample` and `file` only resolve to a concrete operation here (a bundled dataset bulk-loads,
-		// random records insert; a .json upload inserts, any other extension bulk-loads), so this is the
-		// first point the allowlist can be checked against what will actually be sent.
+		// A file's extension decides its operation, so `file` is the one method whose source is unknown
+		// until here; the sample picker offers only granted sources already.
 		if (!importCapabilities.allowsSource(source.kind)) {
 			form.setError('method', {
 				message: source.kind === 'json-records'
@@ -375,14 +374,16 @@ export function ImportDataModal({
 													<SelectValue placeholder="Select a dataset..." />
 												</SelectTrigger>
 												<SelectContent>
-													<SelectGroup>
-														{sampleDatasets.map((dataset) => (
-															<SelectItem key={dataset.id} value={dataset.id}>
-																{dataset.name}
-															</SelectItem>
-														))}
-													</SelectGroup>
-													{canGenerateRandom && (
+													{importCapabilities.allowsSource('csv-data') && (
+														<SelectGroup>
+															{sampleDatasets.map((dataset) => (
+																<SelectItem key={dataset.id} value={dataset.id}>
+																	{dataset.name}
+																</SelectItem>
+															))}
+														</SelectGroup>
+													)}
+													{canGenerateRandom && importCapabilities.allowsSource('json-records') && (
 														<SelectGroup>
 															<SelectLabel>This table</SelectLabel>
 															<SelectItem value={RANDOM_DATASET_ID}>
--- a/src/hooks/checkOperationPermission.ts
+++ b/src/hooks/checkOperationPermission.ts
@@ -7,6 +7,7 @@ import {
 	LocalRolePermission,
 	LocalRolePermissionAction,
 } from '@/integrations/api/api.patch';
+import type { ImportSource } from '@/integrations/api/instance/database/importData';
 import { getOperationsAllowlist } from '@/integrations/api/localRolePermission';
 
 /**
@@ -22,7 +23,36 @@ export const TABLE_ACTION_OPERATIONS: Record<LocalRolePermissionAction, readonly
 	delete: ['delete'],
 };
 
-const CSV_LOAD_OPERATIONS = ['csv_data_load', 'csv_url_load'];
+/**
+ * What each import path issues, as an all-of list. A CSV load returns a job id the client then polls,
+ * so `get_job` belongs to those paths: without it a load commits, the poll 403s, and the reported
+ * failure invites a duplicating retry.
+ */
+const RECORDS_PATH = ['insert'];
+const CSV_DATA_PATH = ['csv_data_load', 'get_job'];
+const CSV_URL_PATH = ['csv_url_load', 'get_job'];
+
+const IMPORT_SOURCE_PATHS: Record<ImportSource['kind'], readonly string[]> = {
+	'json-records': RECORDS_PATH,
+	'csv-data': CSV_DATA_PATH,
+	'csv-url': CSV_URL_PATH,
+};
+
+/**
+ * The paths each Import Data method can take. `sample` and `file` stay ambiguous until submit — a
+ * bundled dataset bulk-loads while random records insert, and a `.json` upload inserts while any other
+ * extension bulk-loads — so a method is offered while either path is open, and
+ * {@link checkImportSourceAllowed} settles the actual choice.
+ */
+const IMPORT_METHOD_PATHS = {
+	sample: [RECORDS_PATH, CSV_DATA_PATH],
+	file: [RECORDS_PATH, CSV_DATA_PATH],
+	url: [CSV_URL_PATH],
+} satisfies Record<string, readonly (readonly string[])[]>;
+
+export type ImportMethod = keyof typeof IMPORT_METHOD_PATHS;
+
+export const IMPORT_METHODS = Object.keys(IMPORT_METHOD_PATHS) as ImportMethod[];
 
 // Keyed by the allowlist array, not the permission: a role edit yields a new array, so an entry
 // cannot go stale.
@@ -106,18 +136,28 @@ export function checkTableActionAllowed(
 	return checkTableOperationsAllowed(permission, TABLE_ACTION_OPERATIONS[action] ?? []);
 }
 
-/**
- * Whether any Import Data source is usable. JSON records post straight to `insert`, but both CSV
- * sources start a job and then poll it, so a CSV grant without `get_job` would report failure for a
- * load that actually committed — and a retry would duplicate the rows.
- */
-export function checkImportDataOperationsAllowed(
+function checkPathAllowed(permission: LocalRolePermission | undefined, path: readonly string[]): boolean {
+	return path.every((operation) => checkAnyOperationAllowed(permission, [operation]));
+}
+
+/** Whether the source the user actually chose can run — the check the submit path owes them. */
+export function checkImportSourceAllowed(
 	permission: LocalRolePermission | undefined,
+	kind: ImportSource['kind'],
 ): boolean {
-	if (permission?.super_user === true) {
-		return true;
-	}
-	return checkAnyOperationAllowed(permission, TABLE_ACTION_OPERATIONS.insert)
-		|| (checkAnyOperationAllowed(permission, CSV_LOAD_OPERATIONS)
-			&& checkAnyOperationAllowed(permission, ['get_job']));
+	return permission?.super_user === true || checkPathAllowed(permission, IMPORT_SOURCE_PATHS[kind]);
+}
+
+/** Whether an import method is worth offering: at least one of its paths is open. */
+export function checkImportMethodAllowed(
+	permission: LocalRolePermission | undefined,
+	method: ImportMethod,
+): boolean {
+	return permission?.super_user === true
+		|| IMPORT_METHOD_PATHS[method].some((path) => checkPathAllowed(permission, path));
+}
+
+/** Whether any import method survives — the gate for the launchers. */
+export function checkImportDataOperationsAllowed(permission: LocalRolePermission | undefined): boolean {
+	return IMPORT_METHODS.some((method) => checkImportMethodAllowed(permission, method));
 }
--- a/src/hooks/checkOperationPermission.ts
+++ b/src/hooks/checkOperationPermission.ts
@@ -1,0 +1,123 @@
+import {
+	canonicalOperationName,
+	OPERATION_GROUPS,
+} from '@/features/instance/config/roles/operations/operationsCatalog';
+import {
+	LocalRoleAttributePermissionAction,
+	LocalRolePermission,
+	LocalRolePermissionAction,
+} from '@/integrations/api/api.patch';
+import { getOperationsAllowlist } from '@/integrations/api/localRolePermission';
+
+/**
+ * Harper's `operations` allowlist (5.0+) is reached only by roles verifyPerms has not already cleared:
+ * `isSuperUser && !isSuSystemOperation` returns first, and `structure_user` returns for create/drop
+ * database, table, and attribute. Table reads and writes are the only Studio surface it can deny.
+ */
+
+export const TABLE_ACTION_OPERATIONS: Record<LocalRolePermissionAction, readonly string[]> = {
+	read: ['search', 'search_by_conditions', 'search_by_hash', 'search_by_value', 'sql'],
+	insert: ['insert'],
+	update: ['update'],
+	delete: ['delete'],
+};
+
+const CSV_LOAD_OPERATIONS = ['csv_data_load', 'csv_url_load'];
+
+// Keyed by the allowlist array, not the permission: a role edit yields a new array, so an entry
+// cannot go stale.
+const expandedAllowlists = new WeakMap<readonly string[], ReadonlySet<string>>();
+
+const groupMembers = new Map(OPERATION_GROUPS.map((group) => [group.name, group.members]));
+
+/**
+ * Expands the way `expandOperationsPerms` does — groups to their members, everything else verbatim.
+ * Deliberately not `expandEffectiveOperations`, which folds alias spellings together to describe what
+ * a role can effectively do; the gate does not fold, so a lone alias grant must stay the dead entry it
+ * is server-side. Harper's groups carry both spellings of each pair, so group grants are unaffected.
+ */
+function expandLikeTheGate(allowlist: readonly string[]): ReadonlySet<string> {
+	const allowed = new Set<string>();
+	for (const entry of allowlist) {
+		const members = groupMembers.get(entry);
+		if (members) {
+			for (const member of members) {
+				allowed.add(member);
+			}
+		} else {
+			allowed.add(entry);
+		}
+	}
+	return allowed;
+}
+
+/**
+ * Whether the allowlist leaves any of these operations reachable. Absent or malformed restricts
+ * nothing — a denial that cannot be proven must not hide UI — and an empty array denies everything,
+ * as on the server.
+ *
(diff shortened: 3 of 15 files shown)


src/hooks/checkOperationPermission.ts
```diff
@@ -0,0 +1,92 @@
...
+ */
+
+export const TABLE_ACTION_OPERATIONS: Record<LocalRolePermissionAction, readonly string[]> = {
+	read: ['search', 'search_by_conditions', 'search_by_hash', 'search_by_value', 'sql'],
+	insert: ['insert'],
+	update: ['update'],
+	delete: ['delete'],
+};
+
+const CSV_LOAD_OPERATIONS = ['csv_data_load', 'csv_url_load'];
```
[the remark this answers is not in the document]
dawsontoth: Not making this change: the mapping tracks what Studio's import modal *issues*, not what Harper accepts. `ImportSource` in `importData.ts` has exactly three variants — `csv-data` → `csv_data_load`, `csv-url` → `csv_url_load`, `json-records` → `insert` — and `rg csv_file_load|import_from_s3 src` matches nothing outside the catalog, which is Harper's full capability list rather than Studio's call sites. Adding them would show the Import Data button to a role that can only do `import_from_s3`, whose every available source then 403s — the same over-showing bug this file's previous round removed when it split Add Records from Import Data. Worth revisiting if the modal ever gains an S3 or file-upload source.

kriszyp (commented):
This will be great, an excellent UI improvement!

Some things to address (sorry for the inline anchor failure):

---
Proposed inline comments (anchors failed):
- `src/features/instance/config/roles/operations/operationsCatalog.ts`: Blocking correctness issue: this says an empty allowlist denies every operation, but I traced Harper v5.0.0, v5.1.26, and v5.2.2 and found that SQL dispatch is special-cased through `verifyPermsAST`; the `permission.operations` gate exists only in `verifyPerms`. Consequently, `operations: []`—or any allowlist omitting `sql`—still permits SQL according to table CRUD, including INSERT/UPDATE/DELETE. The editor would give administrators a false security guarantee. Please enforce the operations gate for SQL in Harper with regression coverage before surfacing this boundary, or restrict Studio support to a server version containing that fix and describe the actual behavior.

— KrAIs (GPT-5)

<sub>(`src/features/instance/config/roles/operations/operationsCatalog.ts:13` is not part of this PR's diff, so this is a file-level comment)</sub>
- `src/features/instance/config/roles/operations/operationsCatalog.ts`: Several enum-valid names offered here are not authorization-grantable in the supported server tags. For example, Harper v5.2.2 registers `setConfiguration` without an `api_name`; the gate therefore compares the handler name `setConfiguration` with the allowlist value `set_configuration` and denies it even though role validation accepted the JSON. The same mismatch affects entries such as `restart_service`, `cleanup_orphan_blobs`, `read_transaction_log`, `set_component_file`, `deploy_component`, and the status operations. Please derive the offered set from effective authorization registrations—or fix that server invariant first—and add a contract test proving every picker entry reaches its handler.

— KrAIs (GPT-5)

<sub>(`src/features/instance/config/roles/operations/operationsCatalog.ts:177` is not part of this PR's diff, so this is a file-level comment)</sub>
- `src/hooks/checkOperationPermission.ts:18`: The browse flow needs exact operation capabilities rather than one “any read operation” result. `DatabaseTableView` still unconditionally issues `search_by_value`, `search_by_conditions`, and `search_by_hash` based on state, while both Export CSV entry points remain visible. For example, a role allowing only `search_by_hash` satisfies this generic read mapping but cannot load the default table view, which calls `search_by_value`. Please expose/check the concrete operation for each query and button, including edit-fetch and export prerequisites.

— KrAIs (GPT-5)
- `src/hooks/checkOperationPermission.ts:89`: This OR is only safe if the modal exposes the matching source. Today an `insert`-only role passes and opens a table-scoped modal whose default CSV-file path calls `csv_data_load`; conversely, `csv_url_load + get_job` passes while the modal still offers CSV files and JSON/sample paths that require other operations. Those permitted-looking choices then fail at submit time. Please return per-source capabilities and hide/disable unsupported methods, or require every method displayed by this launcher.

— KrAIs (GPT-5)
- `src/integrations/api/localRolePermission.ts:38`: This collapses an operation-specific bypass into a role-level boolean. Harper bypasses `structure_user` only for its DDL operations; DML and component/environment operations still reach the allowlist, and `cluster_user` is not a general bypass. The new table tests already encode that DML remains gated, while the editor and user summaries say the list has no effect. Scoped or empty `structure_user` arrays are also truthy here. Please model the bypass per operation (and database for scoped structure roles), update the warnings, and audit the many non-DDL consumers of `useInstanceBrowseManagePermission`.

— KrAIs (GPT-5)
- `src/integrations/api/localRolePermission.ts:55`: This conflicts with the pre-5.0 compatibility handled above. On Harper 4.x, `{ operations: { tables: ... } }` can be a valid database permission, but the roles table and add/edit-user summaries call this helper without a server version and label that role invalid. Please pass the instance version into those presentations or otherwise preserve the object-shaped database interpretation on pre-5.0 servers.

— KrAIs (GPT-5)
- `src/features/instance/config/roles/operations/OperationsAllowlistEditor.tsx:271`: The comment correctly limits component-registered operation grants to 5.2+, but `OperationsPicker` receives no version and offers this custom candidate on every supported 5.0/5.1 instance too. Those releases validate only enum and group names, so saving the generated value is rejected. Please pass the version through and enable the custom path only where dynamic grants are supported; newer catalog entries can retain their separately flagged escape hatch.

— KrAIs (GPT-5)
- `src/integrations/api/api.patch.d.ts:158`: Now that scoped `structure_user` arrays are modeled explicitly—and the save helper correctly notes that they still need table grants outside their DDL scope—`calculateDefaultPermissions` should not return early for every truthy array. As written, editing such a role never adds newly discovered databases, tables, or attributes to its explicit permission template. Please reserve the early return for `structure_user === true` and add a scoped-role regression test.

— KrAIs (GPT-5)

dawsontoth:
Thanks — this was a high-value review. Both server-behavior claims are **confirmed by execution**, not just reading: I ran `verifyPerms` / `verifyPermsAST` through Harper's own mocha harness on `main` (throwaway probe, deleted; Harper checkout clean). Results below, because they're worse than the review states and they land mostly on #1628 and on Harper itself rather than on this PR.

### 1. SQL bypasses the allowlist — confirmed

Same role, `{super_user: false, operations: [], table CRUD all true}`:

| dispatch | verdict |
|---|---|
| operations-API `insert` (`verifyPerms`) | **DENIED** |
| SQL `INSERT` (`verifyPermsAST`) | **ALLOWED** |

So `operations: []` does not deny everything, and any allowlist omitting `sql` still permits SQL writes per table CRUD. The gate genuinely exists only in `verifyPerms`.

This is a Harper server issue plus a **false promise in #1628's editor copy**, not a defect in this PR's gating (which hides controls whose own operations are denied — those really are denied). But #1628 currently tells an admin they've restricted a role when they haven't, which is the failure mode its own commit message calls the worst one a permissions UI has. Filing against Harper and correcting #1628's wording both look necessary; happy to do either.

### 2. Names offered but not grantable — confirmed, and broader

`requiredPermissions` entries registered without an `api_name` fall back to the camelCase handler name, so the gate compares e.g. `setConfiguration` against the allowlist's `set_configuration`. End-to-end on `main`:

- `operations: ['set_configuration']` → `set_configuration` **DENIED**
- `operations: ['get_configuration']` → `get_configuration` **ALLOWED** (control — it registers an `api_name`)

Of 23 sampled names Studio's picker offers, **9 can never match**: `set_configuration`, `restart_service`, `cleanup_orphan_blobs`, `read_transaction_log`, `set_component_file`, `deploy_component`, `get_status`, `set_status`, and `sql`. 29 authorization entries carry no `api_name` at all, so the real list is longer than the sample.

Note `deploy_component` — the deploy-only CI user in #1627's motivating example cannot actually be built today. Again #1628/Harper territory; this PR's gating depends only on `insert`, `update`, `delete`, the searches, `csv_data_load`, `csv_url_load`, and `get_job`, all of which I verified **are** grantable.

### 3–4. Import gating — fixed in e202ffc1

You were right that the any-of launcher gate lets through methods that can't run. Each method is now modeled as the set of paths it can take, because `sample` and `file` don't resolve to an operation until submit (bundled dataset bulk-loads vs. random records insert; `.json` upload inserts vs. any other extension bulk-loads). A method is offered while either path is open, the modal renders only allowed methods and defaults to one of them, and the resolved source is re-checked at submit where the choice is finally known. New `ImportDataModal.permissions.test.tsx` proves an insert-only role loses the URL method, a `csv_url_load` role keeps only URL, and the default selection falls back to an allowed method.

On the read mapping: `TABLE_ACTION_OPERATIONS.read` is currently inert — no production call site passes `'read'`, and the attribute hook has no callers — so per-query gating for `search_by_value` vs `search_by_hash`, and the ungated Export CSV entry points, are real gaps but pre-existing ones this PR doesn't introduce. I'd rather do those as a follow-up than widen this PR further; say the word if you'd prefer them here.

### 5–8

`isElevatedRole`, the pre-5.0 `operations`-as-database-name compatibility, `OperationsPicker`'s missing version, and `calculateDefaultPermissions`' early return are all in #1628's files, not this diff. Worth noting your #5 argues for what this PR already does — it deliberately does **not** use `isElevatedRole`, precisely because DML stays gated for `structure_user`; measured: `{structure_user: true, operations: ['read_only']}` → `insert` **DENIED**. So the divergence is in #1628's warning copy, which says the list has no effect.

src/features/instance/config/roles/operations/operationsCatalog.ts
```diff
@@ -331,6 +331,11 @@ export function getOperationInfo(name: string): GrantableOperation | undefined {
 	return catalogByName.get(name);
```
kriszyp: Blocking correctness issue: this says an empty allowlist denies every operation, but I traced Harper v5.0.0, v5.1.26, and v5.2.2 and found that SQL dispatch is special-cased through `verifyPermsAST`; the `permission.operations` gate exists only in `verifyPerms`. Consequently, `operations: []`—or any allowlist omitting `sql`—still permits SQL according to table CRUD, including INSERT/UPDATE/DELETE. The editor would give administrators a false security guarantee. Please enforce the operations gate for SQL in Harper with regression coverage before surfacing this boundary, or restrict Studio support to a server version containing that fix and describe the actual behavior.

— KrAIs (GPT-5)

<sub>(`src/features/instance/config/roles/operations/operationsCatalog.ts:13` is not part of this PR's diff — anchored to the nearest line this PR's diff can hold)</sub>

src/features/instance/config/roles/operations/operationsCatalog.ts
```diff
@@ -331,6 +331,11 @@ export function getOperationInfo(name: string): GrantableOperation | undefined {
...
 }
 
+/** The authorization entry a name resolves to; aliases collapse onto their canonical operation. */
+export function canonicalOperationName(name: string): string {
+	return catalogByName.get(name)?.aliasOf ?? name;
+}
+
 export function isOperationGroupName(name: string): boolean {
 	return groupsByName.has(name);
 }
```
kriszyp: Several enum-valid names offered here are not authorization-grantable in the supported server tags. For example, Harper v5.2.2 registers `setConfiguration` without an `api_name`; the gate therefore compares the handler name `setConfiguration` with the allowlist value `set_configuration` and denies it even though role validation accepted the JSON. The same mismatch affects entries such as `restart_service`, `cleanup_orphan_blobs`, `read_transaction_log`, `set_component_file`, `deploy_component`, and the status operations. Please derive the offered set from effective authorization registrations—or fix that server invariant first—and add a contract test proving every picker entry reaches its handler.

— KrAIs (GPT-5)

<sub>(`src/features/instance/config/roles/operations/operationsCatalog.ts:353` is not part of this PR's diff — anchored to the nearest line this PR's diff can hold)</sub>

src/hooks/checkOperationPermission.ts
```diff
@@ -0,0 +1,163 @@
...
+import type { ImportSource } from '@/integrations/api/instance/database/importData';
+import { getOperationsAllowlist } from '@/integrations/api/localRolePermission';
+
+/**
+ * Harper's `operations` allowlist (5.0+) is reached only by roles verifyPerms has not already cleared:
+ * `isSuperUser && !isSuSystemOperation` returns first, and `structure_user` returns for create/drop
+ * database, table, and attribute. Table reads and writes are the only Studio surface it can deny.
+ */
+
+export const TABLE_ACTION_OPERATIONS: Record<LocalRolePermissionAction, readonly string[]> = {
```
kriszyp: The browse flow needs exact operation capabilities rather than one “any read operation” result. `DatabaseTableView` still unconditionally issues `search_by_value`, `search_by_conditions`, and `search_by_hash` based on state, while both Export CSV entry points remain visible. For example, a role allowing only `search_by_hash` satisfies this generic read mapping but cannot load the default table view, which calls `search_by_value`. Please expose/check the concrete operation for each query and button, including edit-fetch and export prerequisites.

— KrAIs (GPT-5)

src/hooks/checkOperationPermission.ts
```diff
@@ -0,0 +1,163 @@
...
+	'csv-url': CSV_URL_PATH,
+};
+
+/**
+ * The paths each Import Data method can take. `sample` and `file` stay ambiguous until submit — a
+ * bundled dataset bulk-loads while random records insert, and a `.json` upload inserts while any other
+ * extension bulk-loads — so a method is offered while either path is open, and
+ * {@link checkImportSourceAllowed} settles the actual choice.
+ */
+const IMPORT_METHOD_PATHS = {
```
kriszyp: The new per-source submit check is a good guard, but this any-path method gate still exposes choices that cannot run. On the database-level launcher, an `insert`-only role gets `sample` because random records use `insert`; when there is no existing table, however, random data is not rendered, so the only visible sample choices are bundled CSV datasets and every submission ends in the new client-side denial. The `file` method likewise still advertises and accepts both CSV and JSON when the role can use only one. Please pass source capabilities and table context into the rendered choices, then compute method availability from what is actually selectable, so every visible choice can complete.

— KrAIs (GPT-5)

src/features/instance/config/roles/operations/operationsCatalog.ts
kriszyp: `expandEffectiveOperations` folds an explicitly stored alias into its canonical name, but `checkAnyOperationAllowed` correctly models the server and treats that same alias grant as dead. Thus a role with `operations: ['search_by_id']` is shown as allowing `search_by_hash` in the roles table and user summary even though it authorizes neither; the chip tooltip says the opposite. Please distinguish group expansion from explicit entries: exclude explicit aliases from the effective set (or mark them inert), while still deduplicating canonical/alias pairs inside server groups.

— KrAIs (GPT-5)

<sub>(`src/features/instance/config/roles/operations/operationsCatalog.ts:380` is not part of this PR's diff — anchored to the nearest line this PR's diff can hold)</sub>

DavidCockerill (approved):
Approving at `11db476`. The gate order is right and the composition is clean — `checkTableActionAllowed(...) && checkTableGrant(...)`, with the original body extracted verbatim so the pre-existing deny-on-no-permission behaviour is preserved rather than restated.

**Endorsing the deviation.** Leaving `useInstanceManagePermission` and `useInstanceBrowseManagePermission` ungated is correct even though #1627's wording points the other way. `verifyPerms` returns on `isSuperUser && !isSuSystemOperation` before the allowlist is consulted, so gating those hooks would have hidden UI the server allows — a fail-closed bug dressed as a security fix. Implementing the obvious version first, reading `verifyPerms`, and reverting it is the right sequence, and the two one-line comments left behind are the right amount of explanation.

**Verified and dropped** — four candidates I chased rather than sent:

- `checkAnyOperationAllowed` returning `true` for an absent or malformed allowlist is fail-open in a permissions path, and correct here: it gates affordances only, Harper denies at gate 1, so the worst case is a button that 403s. Failing closed would hide controls from roles that hold the grant.
- `expandLikeTheGate` iterating a non-iterable would throw inside a hook. Not reachable — `getOperationsAllowlist` (`localRolePermission.ts:89-94`) returns `operations` only when `Array.isArray(...)` and every entry is a string, else `undefined`.
- The database-scoped Import Data launcher consults no grant, and the operation it starts creates a table — but the button sits inside `{canManage && …}`, so browse-manage still applies and `canImport` is an additional gate.
- @gemini-code-assist's `CSV_LOAD_OPERATIONS` finding: @dawsontoth's decline is right and I checked it. `csv_file_load` and `import_from_s3` appear only in `operationsCatalog.ts`, never at a Studio call site, and `ImportSource` has exactly three variants. A role granted only `csv_file_load` *should* see Import Data hidden, because Studio cannot issue it.

One thing for the record, not a request on this PR: `localRolePermission.ts` documents a `breaks-auth` class where a truthy non-iterable under `operations` makes `listUsers` throw behind a truthiness-only guard, failing authentication instance-wide. That is #1628 territory and you have clearly already mapped it — noting so it is not lost.

This is the complete set of my concerns at this head.

src/features/instance/config/roles/operations/operationsCatalog.ts
```diff
@@ -349,6 +349,11 @@ export function getOperationInfo(name: string): GrantableOperation | undefined {
 	return catalogByName.get(name);
```
kriszyp: Blocking correctness issue: this says an empty allowlist denies every operation, but I traced Harper v5.0.0, v5.1.26, and v5.2.2 and found that SQL dispatch is special-cased through `verifyPermsAST`; the `permission.operations` gate exists only in `verifyPerms`. Consequently, `operations: []`—or any allowlist omitting `sql`—still permits SQL according to table CRUD, including INSERT/UPDATE/DELETE. The editor would give administrators a false security guarantee. Please enforce the operations gate for SQL in Harper with regression coverage before surfacing this boundary, or restrict Studio support to a server version containing that fix and describe the actual behavior.

— KrAIs (GPT-5)

<sub>(`src/features/instance/config/roles/operations/operationsCatalog.ts:13` is not part of this PR's diff — anchored to the nearest line this PR's diff can hold)</sub>

src/features/instance/config/roles/operations/operationsCatalog.ts
```diff
@@ -349,6 +349,11 @@ export function getOperationInfo(name: string): GrantableOperation | undefined {
...
 }
 
+/** The authorization entry a name resolves to; aliases collapse onto their canonical operation. */
+export function canonicalOperationName(name: string): string {
+	return catalogByName.get(name)?.aliasOf ?? name;
+}
+
 export function isOperationGroupName(name: string): boolean {
 	return groupsByName.has(name);
 }
```
kriszyp: Several enum-valid names offered here are not authorization-grantable in the supported server tags. For example, Harper v5.2.2 registers `setConfiguration` without an `api_name`; the gate therefore compares the handler name `setConfiguration` with the allowlist value `set_configuration` and denies it even though role validation accepted the JSON. The same mismatch affects entries such as `restart_service`, `cleanup_orphan_blobs`, `read_transaction_log`, `set_component_file`, `deploy_component`, and the status operations. Please derive the offered set from effective authorization registrations—or fix that server invariant first—and add a contract test proving every picker entry reaches its handler.

— KrAIs (GPT-5)

<sub>(`src/features/instance/config/roles/operations/operationsCatalog.ts:371` is not part of this PR's diff — anchored to the nearest line this PR's diff can hold)</sub>

src/hooks/checkOperationPermission.ts
kriszyp: The browse flow needs exact operation capabilities rather than one “any read operation” result. `DatabaseTableView` still unconditionally issues `search_by_value`, `search_by_conditions`, and `search_by_hash` based on state, while both Export CSV entry points remain visible. For example, a role allowing only `search_by_hash` satisfies this generic read mapping but cannot load the default table view, which calls `search_by_value`. Please expose/check the concrete operation for each query and button, including edit-fetch and export prerequisites.

— KrAIs (GPT-5)

src/hooks/checkOperationPermission.ts
```diff
@@ -0,0 +1,163 @@
...
+ *
+ * Matching mirrors gate 1: groups expand, entries match verbatim, and the checked operation resolves
+ * to its canonical api_name. Harper registers `search_by_id` under `search_by_hash`, so granting a
+ * legacy alias grants nothing, while a canonical grant covers both spellings.
+ */
+export function checkAnyOperationAllowed(
+	permission: LocalRolePermission | undefined,
+	operations: readonly string[],
+): boolean {
+	const allowlist = getOperationsAllowlist(permission);
```
kriszyp: This version-blind shape check conflicts with the repository’s own compatibility contract. `src/integrations/api/localRolePermission.test.ts:134` establishes that `operations: ['read_only']` is inert below the 5.0.0-alpha.8 floor, but this helper activates it and hides writes that pre-5.0 Harper allows. It makes the opposite error for malformed 5.x values: `getOperationsAllowlist` returns undefined and the UI allows everything, although Harper still enters its gate and denies or throws. Please carry allowlist support/classification into these hooks instead of inferring semantics from shape, with regression coverage for both cases.

— KrAIs (GPT-5)

src/features/instance/databases/modals/ImportDataModal.tsx
```diff
@@ -131,12 +133,22 @@ export function ImportDataModal({
...
+	// Offer only methods the role can actually run; a method whose every path is denied would fail at
+	// submit with a 403 the user can do nothing about.
+	const availableMethods = useMemo(
+		() => importMethods.filter(({ value }) => importCapabilities.methods[value]),
+		[importCapabilities],
+	);
+	// With a table already in context (table toolbar) the likely intent is loading your own data;
+	// without one (sidebar, empty database) lead with samples -- then fall back to whatever is allowed.
+	const preferred: ImportMethod = tableName ? 'file' : 'sample';
+	const defaultMethod = importCapabilities.methods[preferred] ? preferred : (availableMethods[0]?.value ?? preferred);
```
kriszyp: The per-source submit guard and sample filtering are good improvements, but method availability is still computed without the current target context. A structure user with `operations: ['insert']` opening Import on an empty database defaults to Sample because random records are theoretically allowed; with no table columns, however, Random Data is absent and the CSV datasets are filtered out, leaving an empty required picker. The `file` method likewise still advertises and accepts both CSV and JSON when the role can use only one. Please pass source capabilities and table context into the rendered choices, then compute method availability from what is actually selectable, so every visible choice can complete.

— KrAIs (GPT-5)

src/features/instance/databases/modals/ImportDataModal.tsx
```diff
@@ -265,6 +277,25 @@ export function ImportDataModal({
 			source = { kind: 'csv-url', url: values.url };
 		}
 
+		// The launcher checked the table it was opened from; the target is editable, so an existing table
+		// chosen here needs its own insert grant. A table that does not exist yet has none to check --
+		// creating it is the browse-manage authority the launcher already required.
+		if (tableExists && !importCapabilities.allowsDestination(database, table)) {
```
kriszyp: This assumes every launcher required browse-manage authority before a missing table could be selected, but the table toolbar and context menu only require an import operation plus an insert grant on the original table. A regular role can open the modal from `dog`, edit the target to a nonexistent table, pass both submit checks, and then fail when `importData` first issues `create_table`. Please check create-table capability for absent targets—including super user, the applicable structure-user scope, or an effective `create_table` grant—or prevent roles without it from selecting a new target. Add a table-scoped regression test for this path.

— KrAIs (GPT-5)

PR #1632, opened 2026-08-18
fix(datadog): keep visitor e-mail addresses out of RUM view URLs

dawsontoth:
Studio was sending visitors' e-mail addresses — and live password-reset tokens — to Datadog RUM inside the URL fields of every auth screen, where they were retained and searchable. This redacts both in `beforeSend`, for every event type.

The addresses ride in the app-generated `?me=`/`?email=` form-persistence params and in the `/config/users/<address>` path; the token rides in `?token=` on `/reset-password` and `/verify-email`, and `useResetPassword` exchanges it for a password change. Nothing redacted any of it: `beforeSend` only ever touched `error.*` fields, and `shouldKeepEvent` returns early for non-error events, so the view and resource events carrying the bulk of it were never visited.

## For the human reviewer

1. **Redact at the sink, not at the source.** Every writer could encode instead, but that is an invariant each future caller has to re-earn and its failure mode is silent PII. Fixing the sink is one choke point. Fully reversible. **What a "no" costs:** the source fix is strictly larger and touches six call sites; it is also the *only* fix for the third-party leak below, so this is "as well as", not "instead of".
2. **A partial redaction is treated as worse than none.** `zodRequireEmail` accepts `'()!~*` and `encodeURIComponent` leaves them bare, so a value that ended at those emitted `?email=<redacted>'reilly%40example.com` — still identifying, but looking handled. The value therefore ends only at delimiters no accepted address can contain. **Cost:** an apostrophe or paren *wrapping* a URL in prose is consumed into the redaction. Ruling requested by @kriszyp; I agree with it, but it is the one trade a reviewer might reverse.
3. **Credential param names go beyond the observed exposure.** Only `token` appears in RUM today; `access_token`, `id_token`, `refresh_token`, `code`, `secret`, `password`, `api_key` do not. For a credential redactor I treated the standard names as the point rather than speculative generality. **A "no"** means trimming the list to `token` and accepting that the next OAuth param to reach a URL ships once before anyone notices.
4. **URL redaction runs *before* `shouldKeepEvent`, so dropped events pay for it.** The filter can throw on a malformed error field, and the SDK's `catchUserErrors` swallows that into sending the event anyway — a view event cannot be dismissed from `beforeSend` at all. The cheaper order is the one that leaks. Both outside legs raised the allocation cost; I kept the order and made the common path a single `test` instead.
5. **Scope: this fixes Datadog only.** The same URLs are captured by the HubSpot, Meta, LinkedIn and Google Analytics beacons, which `beforeSend` cannot touch — measured, and filed separately. If the view is that the source fix should land first, this PR is the wrong shape and I would rather hear that now.
6. **Known cosmetic residue:** an *encoded* JSON param value redacts to `?filters=<redacted>%22%7D`, because `%` is legal in a local part so the match cannot tell where the encoding stops. No address survives; pinned by a test that says so.

## Verification

**Fails on base.** A detached worktree at the merge base, calling *base's* `beforeSend` directly (not the new module, so the failure is behavioural rather than a missing import):

```
× keeps the visitor address out of the view URL
  AssertionError: expected '…/#/sign-in?me=someone%40example.com' not to contain 'someone%40example.com'
× keeps the reset token out of the view URL
  AssertionError: expected '…/#/reset-password?token=abc.def' not to contain 'abc.def'
```

Both pass on the branch.

**Gate:** `vitest run` 294 files / 2,316 tests, `tsc -b`, `oxlint`, `dprint check` — all clean under Node 24.19.0 (`.nvmrc`).

**Backtracking bound, measured.** An earlier revision of the address match had two repetitions that both accepted `-`: 18ms at 22 characters, 553ms at 26, **8.7s at 30**, on the main thread, per event. Now 0.2ms at 40 characters, with a test asserting under 100ms.

**Exposure quantified in RUM** (app `f590deee-…`, 30 days): 991 view events across **332 sessions** carrying an address, plus 8,988 resource / 448 long-task / 289 action events; **140 view events across 77 sessions** carrying a reset token, plus 3,732 resource and 1,400 long-task. 190 resource events carry an address in `resource.url` itself, 62 of them Studio's own auth URLs.

**End-to-end route: not observable in the browser preview** — `beforeSend` produces no UI, and the only true end-to-end exercise is the real SDK in production. Unit coverage plus the RUM measurements above are the evidence; what the tests do **not** prove is behaviour against real `@datadog/browser-rum` payloads (read-only or proxied properties), which both legs correctly flagged as a gap.

## Review coverage

Authored by **Opus 5** (`claude-opus-5[1m]`). Six pre-push rounds, because rounds 1–5 each found a real defect and three of them were regressions introduced by the previous round's fix.

| lens | outcome |
| --- | --- |
| codex `gpt-5.6-sol` | ✓ all rounds (graded leg) |
| gemini via `agy` (default model, not pinned by the run) | ✓ rounds 1, 4, 5 · ✗ `no-output` rounds 2, 3, 6 |
| cursor-composer | ✗ `cursor-agent` not installed (rounds 1–4); refuses a diff touching `AGENTS.md` (round 5); pruned as a low-risk delta (round 6) |
| cursor-grok | ✗ pruned by policy every round |
| Harper domain adjudication | ✗ `exit-1` rounds 1–5, zero-byte log; pruned as a low-risk delta (round 6) |

The adjudicator never ran, so outside findings were **not** machine-adjudicated — I triaged them by hand and rejected two on evidence: Codex's claim that `error.handling_stack` is outside the SDK's editable list (it is present in the installed 7.8.0 bundle's ERROR entry), and Gemini's request for a null-`event` guard (the SDK's types make it non-nullable). Everything else was reproduced before being fixed.

<sub>Human-Review-Need: 3 @ 1a54ef3895ff</sub>


src/integrations/datadog/beforeSend.ts
```diff
@@ -32,5 +41,14 @@ export function beforeSend(event: DatadogErrorEvent) {
 			error.resource.url = redactErrorText(error.resource.url);
 		}
 	}
+	const view = event.view;
```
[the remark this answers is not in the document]
dawsontoth: Applied in 318ac76e — a `redactErrorAndEmails` helper composes both redactions over `error.message`, `error.stack` and `error.resource.url`, plus `handling_stack` as the same class of field; confirmed first that no error in the last 30 days (1,110 sampled events) actually carried an address, so this lands as defence in depth against the Harper-host exemption rather than an observed leak.

src/integrations/datadog/beforeSend.test.ts
```diff
@@ -65,4 +65,48 @@ describe('beforeSend', () => {
...
+	it('redacts the view URL of a resource event', () => {
+		const event: DatadogErrorEvent = {
+			type: 'resource',
+			view: { url: 'https://fabric.harper.fast/#/verifying?email=someone%40example.com' },
+		};
+		expect(beforeSend(event)).toBe(true);
+		expect(event.view?.url).toBe('https://fabric.harper.fast/#/verifying?email=<redacted>');
+	});
+
+	it('redacts the view URL of a kept error event alongside its message', () => {
```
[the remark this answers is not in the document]
dawsontoth: Added in 318ac76e, and it earned its keep — writing it exposed a real bug in `redactEmailParams`: the value class `[^&#]+` also admitted whitespace, so an address quoted inside a multi-line stack matched past the end of its URL and swallowed every following frame into the redaction. The class now mirrors `redactErrorText`'s own `URL_TOKEN` terminator set, with the multi-line stack, trailing-fragment and mid-sentence cases pinned by tests.

src/integrations/datadog/redactEmailParams.ts
```diff
@@ -0,0 +1,41 @@
...
+ * The form-persistence params, anchored to a `?`/`&` so only the whole param name matches: the
+ * `me=` in `?resume=…` is preceded by `u`, not a separator. Requires a non-empty value, so the
+ * bare `?me=` the auth links emit when no address has been typed yet stays as it is.
+ *
+ * The value ends where a URL ends — the same terminator set `redactErrorText`'s `URL_TOKEN` uses,
+ * rather than just `&`/`#`. These strings are not always bare URLs: an address quoted inside a
+ * multi-line stack is followed by a newline and then the next frame, and a class that admitted
+ * whitespace would swallow the remainder of the stack into the redaction.
+ */
+const EMAIL_PARAM = /([?&](?:me|email)=)[^&#\s'"<>)\]]+/gi;
```
kriszyp: An apostrophe cannot safely terminate the value here. Studio’s email validator accepts `o'reilly@example.com`, and the `/verifying` transitions build this hash query with `encodeURIComponent`, which leaves apostrophes unescaped. The resulting `?email=o'reilly%40example.com` is rewritten to `?email=<redacted>'reilly%40example.com`, so most of the address still reaches Datadog. Please keep apostrophes inside the matched value—or explicitly encode them at every writer—and add this exact regression case. For privacy, consuming a surrounding quote is safer than leaving part of an address behind.

kriszyp (dismissed):
**Automated gate — not yet queued for human review.**

This PR's AI review found issues, and the PR description reports no cross-model reviews.
Per team policy, a substantive PR with AI-review findings is queued for human review only after at least 2 cross-model reviews have been run, their findings addressed, and the coverage reported in the PR description (`## Review coverage` naming each model — see harper-engineering-guidelines).
The findings below count as one of the two: address them, run a second outside-model review, update the description, and the gate lifts automatically on the next pass.

---
## TL;DR

Commit `318ac76e` correctly extends auth-parameter redaction to supported Datadog error fields and fixes the multiline overmatch.
However, the new terminator class stops at apostrophes, so a valid address such as `o'reilly@example.com` is only partially redacted.
Because this defeats the PR’s privacy guarantee for a real accepted input, changes are required.
The prior ruling that URL-based form persistence is broader than this Datadog-specific scope remains unchanged.

verdict: CHANGES
merge: rebase
Human-Review-Need: 4 @ 318ac76e80df
Human-Review-Need: 6 @ 318ac76e80df

## Diff tour

- `src/integrations/datadog/beforeSend.ts:39-69` composes repository/URL redaction with auth-parameter redaction across message, stack, handling stack, resource URL, view URL, and referrer while preserving filter-before-redact ordering.
- `src/integrations/datadog/shouldKeepEvent.ts:9-22` adds `error.handling_stack` to the internal event shape. This is supported by the locked SDK: DataDog/browser-sdk#4357, [“Allow updates of `error.handling_stack`”](https://github.com/DataDog/browser-sdk/pull/4357), added writability in 6.32.
- `src/integrations/datadog/redactEmailParams.ts:26-40` adds prose/stack terminators to prevent redaction from consuming subsequent frames, but the apostrophe terminator introduces the finding below.
- `src/integrations/datadog/beforeSend.test.ts:113-151` covers composed error-field redaction; `src/integrations/datadog/redactEmailParams.test.ts:57-77` covers multiline, fragment, and prose boundaries but misses an accepted apostrophe-containing address.

## Findings

major — `src/integrations/datadog/redactEmailParams.ts:36` — treating apostrophe as a URL terminator leaves most of a valid email address unredacted

## Validation

The one-commit delta and `git diff --check` were inspected. A standalone reproduction confirms that `encodeURIComponent("o'reilly@example.com")` leaves the apostrophe unescaped and the new regex produces `?email=<redacted>'reilly%40example.com`.

Vitest was not run because this read-only checkout has no `node_modules`.

## Review coverage

| lens | outcome |
|---|---|
| gemini | **pruned** — pruned (policy minimal) |
| cursor-grok | **pruned** — pruned (policy minimal) |
| cursor-composer | **pruned** — pruned (policy minimal) |
| codex | changes — valid-email suffix can escape redaction |
| domain | **pruned** — pruned (policy minimal) |

---
_Pre-push review of `fix/rum-redact-auth-email-in-view-urls` (318ac76e80df) vs `origin/stage` by codex._
_Review emphasis: Dispatch-configured._

## Review coverage

| lens | outcome |
|---|---|
| gemini | **pruned** — pruned (policy minimal) |
| cursor-grok | **pruned** — pruned (policy minimal) |
| cursor-composer | **pruned** — pruned (policy minimal) |
| codex | ok — graded leg — produced review.md + comments.json |
| domain | **pruned** — pruned (policy minimal) |


---
_Pre-push review of `fix/rum-redact-auth-email-in-view-urls` (318ac76e80df) vs `origin/stage` by codex._
_Review emphasis: Dispatch-configured._

— codex review, submitted by the dispatch review gate

src/integrations/datadog/redactEmailParams.ts
dawsontoth: Fixed in 31d7bbee — verified the claim end to end first (`zodRequireEmail` accepts `o'reilly@example.com`, `encodeURIComponent` leaves `'` unescaped, and the old class produced exactly `?email=<redacted>'reilly%40example.com`), then narrowed the terminator set to `&`, `#` and whitespace only, so quotes and brackets now stay inside the value and a wrapping delimiter is consumed instead; regression cases cover that address plus its `%27` spelling and the `(`/`)`/`!`/`~`/`*` characters `encodeURIComponent` also leaves bare. Fixed at the redactor rather than at each writer because encoding correctly at every call site is an invariant every future caller has to re-earn and the failure mode is silent.

cb1kenobi: **High: `?token=` on `/reset-password` and `/verify-email` still ships in `view.url`**

The param set stops at `me|email`, but `ResetPassword.tsx:31` and `VerifyEmail.tsx:90` read `token` out of the same hash-routed query, and `useResetPassword.ts` exchanges it for a password change via `PUT /ResetPassword/`. So the reset link puts a live bearer credential in `window.location` — the exact field this PR is now cleaning — and it survives into `view.url`/`view.url_hash`, retained and searchable by anyone with RUM read access for as long as the token is valid. Pre-existing rather than introduced here, but it is the same leak in the same field, and one alternative away in this regex:

Suggested fix:
```suggestion
const SENSITIVE_PARAM = /([?&](?:me|email|token)=)[^&#\s'\"<>)\]]+/gi;
```

(rename the constant and its doc block if you take this — it stops being only about e-mail.)

—
_Generated by Barber AI_

src/integrations/datadog/redactEmailParams.ts
```diff
@@ -0,0 +1,41 @@
...
+
+/** Redact the e-mail-bearing auth params in `url`. Returns it unchanged if there are none. */
+export function redactEmailParams(url: string) {
```
cb1kenobi: **Medium: a param-name list leaves `?filters=` shipping addresses — match the value shape instead**

`RelationshipCell.tsx:82` deep-links with `search={{ filters: { [relatedPrimaryKey]: value } }}` (read back at `DatabaseTableView.tsx:186`), so on a table keyed by e-mail the view URL is `?filters=%7B%22email%22%3A%22user%40example.com%22%7D` — an address in `view.url` that `EMAIL_PARAM` does not match. I ran the function against that shape to confirm it passes through untouched. A name list only ever covers the params someone remembered; the thing you actually care about is the value shape.

Matching any param whose *value* contains `@`/`%40` covers `me`, `email`, `filters` and the next one, and still keeps the property this module was built for — it matches values, not paths, so `/HDBInstance/<id>/operation` stays readable. I ran the existing datadog suite against it: 63/63 pass (including the `?resume=`/`?username=` over-match test and the bare `?me=` case), it redacts the `filters` URL, and it leaves `?v=abc123`, `/HDBInstance/<id>/operation?q=1` and Datadog stack frames alone:

Suggested fix:
```suggestion
const EMAIL_PARAM = /([?&][^=&#\s]+=)[^&#\s'\"<>)\]]*(?:@|%40)[^&#\s'\"<>)\]]*/gi;
```

Keep a name list only for the non-address secrets (`token`, per the other comment).

—
_Generated by Barber AI_

src/integrations/datadog/beforeSend.ts
```diff
@@ -15,21 +16,56 @@ import { DatadogErrorEvent, shouldKeepEvent } from './shouldKeepEvent';
...
 		}
 		if (error.resource?.url) {
-			error.resource.url = redactErrorText(error.resource.url);
+			error.resource.url = redactErrorAndEmails(error.resource.url);
+		}
+	}
+	const view = event.view;
+	if (view) {
+		if (view.url) {
+			view.url = redactEmailParams(view.url);
```
cb1kenobi: **Medium: `resource.url` on non-error resource events is never redacted**

This block reaches `view.url`/`view.referrer` for every event type, but a resource event also carries its own top-level `resource.url`, and only `error.resource.url` (line 59) is touched. `resource.url` is on the SDK's modifiable-field list for `RESOURCE` (`MODIFIABLE_FIELD_PATHS_BY_EVENT` in `browser-rum-core@7.8.0/cjs/domain/assembly.js`), so it is editable here — I checked the installed bundle. This PR's own new test asserts `https://api.harper.fast/User/?email=…` needs redacting under `error.resource.url`; the identical URL on a request that *succeeded* arrives as a resource event and escapes.

No current Studio call puts an address in a query string (sign-up is `POST /User/` with a body), so this is a latent gap rather than an observed leak — but the diff already visits every event type, and leaving one URL-bearing field out is the kind of hole this PR exists to close.

Fix: add `resource?: { url?: string }` to `DatadogErrorEvent` and redact `event.resource?.url` in this same block.

—
_Generated by Barber AI_

src/integrations/datadog/beforeSend.ts
```diff
@@ -15,21 +16,56 @@ import { DatadogErrorEvent, shouldKeepEvent } from './shouldKeepEvent';
...
+		}
+		if (error.handling_stack) {
+			error.handling_stack = redactErrorAndEmails(error.handling_stack);
 		}
 		if (error.resource?.url) {
-			error.resource.url = redactErrorText(error.resource.url);
+			error.resource.url = redactErrorAndEmails(error.resource.url);
+		}
+	}
+	const view = event.view;
```
cb1kenobi: **Low: the view redaction runs last, and `beforeSend` fails open**

The SDK wraps this in `catchUserErrors` (`browser-rum-core@7.8.0/cjs/domain/configuration/configuration.js:187`): a throw is swallowed and the call returns `undefined`, while `shouldSend` drops only on an explicit `false`. So a throw anywhere above this point ships the event with whatever mutations had already been applied — and for view events even an explicit `false` cannot drop it (`assembly.js`: `if (result === false && event.type !== VIEW) return false`, then it warns "Can't dismiss view events using beforeSend!").

Nothing throws today (`redactEmailParams` is a pure `String.replace`), but a non-string `view.url` throws `url.replace is not a function` — I confirmed that by calling `beforeSend({ type: 'view', view: { url: 12345 } })`. Given that the failure mode is a visitor address shipped rather than an event dropped, a `typeof view.url === 'string'` guard (or a try/catch around the redaction) is cheap insurance.

—
_Generated by Barber AI_

src/integrations/datadog/redactEmailParams.ts
```diff
@@ -0,0 +1,41 @@
...
+	return url.replace(EMAIL_PARAM, `$1${REDACTED}`);
```
cb1kenobi: **Low: trailing punctuation is swallowed into the redaction**

The value class admits `,`/`.`/`;`/`:`, so `Failed at https://fabric.harper.fast/#/sign-in?me=user%40example.com, retrying` comes out as `…?me=<redacted> retrying` — the comma is gone. Verified by running the function. `redactErrorText` handles exactly this with `TRAILING_PUNCTUATION` before rebuilding the URL; this sibling doesn't, so kept error text loses characters for no reason. Note you can't just exclude `.` from the class (the value would stop mid-domain), so mirror the sibling instead:

Suggested fix:
```suggestion
export function redactEmailParams(url: string) {
	return url.replace(EMAIL_PARAM, (match: string, param: string) => {
		const trailing = /[.,:;!?]+$/.exec(match)?.[0] ?? '';
		return `${param}${REDACTED}${trailing}`;
	});
}
```

—
_Generated by Barber AI_

src/integrations/datadog/redactEmailParams.ts
dawsontoth: Fixed — this was the most serious thing in the PR and I verified it before acting: `ResetPassword.tsx:31` and `VerifyEmail.tsx:90` do read `token` from the hash query, `useResetPassword` exchanges it via `PUT /ResetPassword/`, and RUM had **140 view events across 77 sessions** carrying one in the last 30 days (plus 3,732 resource and 1,400 long-task events attributed to those views). `token` is now in the named pass and the module is renamed `redactSensitiveParams`. I also checked `code`, `state`, `access_token`, `id_token`, `key`, `secret`, `password`, `apikey` and `session` against RUM — zero occurrences, so `token` is the whole credential exposure.

src/integrations/datadog/redactEmailParams.ts
dawsontoth: Taken — a second pass now matches any param whose value holds an address, so `filters`, `me`, `email` and the next one are covered without an edit, and it still matches values rather than paths so `/HDBInstance/<id>/operation` stays readable. Kept the named pass alongside it rather than replacing it, because `token` has no `@` in its value and would fall straight through a value-shape-only match. Worth flagging what neither pass reaches: an address in a URL *path* (`/config/users/<address>`, 12 of 81 sampled view events) — documented in the description and filed separately, since the same URLs are also captured by third-party pixels that `beforeSend` cannot touch.

src/integrations/datadog/beforeSend.ts
dawsontoth: Fixed — independently flagged by the Gemini leg of a pre-push cross-model review, and already committed before this comment landed. One correction to the finding: it is not purely latent. Checking RUM rather than assuming, **190 resource events in the last 30 days carry an encoded `@` in `resource.url`, 62 of them Studio's own auth-screen URLs** — the rest are third-party pixels transmitting the page URL. So the field was leaking in production, not just theoretically reachable. `resource?: { url?: string }` added to the event shape and redacted in the same block, as suggested.

src/integrations/datadog/beforeSend.ts
dawsontoth: Fixed — every field is now type-checked rather than truth-checked (`typeof x === 'string'`), across the error text fields and both URL fields, plus a test asserting `beforeSend({ type: 'view', view: { url: 12345 } })` doesn't throw. Your reasoning is the part that made this worth doing rather than deferring: because `catchUserErrors` swallows the throw and a view event can't be dismissed from here at all, the failure mode is a half-redacted event *shipped*, not an event dropped. Failing open is the one failure mode a redactor must not have.

src/integrations/datadog/redactEmailParams.ts
dawsontoth: Fixed, mirroring the sibling as you suggested — `TRAILING_PUNCTUATION` is stripped off the match and re-appended after `<redacted>`. Note the comma case specifically was already resolved en route: the value class now ends at `,` (a comma can't occur in an address `zodRequireEmail` accepts), which came out of a separate finding about an embedded URL eating the JSON fields after it. The replacer still earns its keep for `.`, `;`, `:`, `!` and `?`, which can't be excluded from the class without truncating mid-domain — tests cover `.`, `;` and `,`.

cb1kenobi:
Heads-up — the fixes described in the replies above don't appear to have landed on this branch yet.

The PR is still at `318ac76e` (2 commits: `3f1a6dd2`, `318ac76e`), and `redactEmailParams.ts` at that head still reads:

```ts
const EMAIL_PARAM = /([?&](?:me|email)=)[^&#\s'"<>)\]]+/gi;
```

No `token` alternative, and no second value-shape pass. Same for the other four — I re-read the files at the current head rather than going by the thread state.

Most likely they're committed locally and just not pushed. Flagging it because all five threads are resolved, so the PR currently *reads* as fully addressed — easy for the next reviewer to take at face value.

For what it's worth, the replies themselves are excellent: pulling the actual RUM counts (140 view events across 77 sessions carrying a reset token; 190 resource events with an encoded `@`, 62 of them our own auth screens) turned two of these from "reachable in principle" into measured exposure. And the reasoning for keeping the named pass alongside the value-shape pass — that `token` has no `@` and would fall straight through a value-only match — is exactly right; that's the trap I'd have walked into if I'd proposed replacing rather than adding.

Happy to re-review as soon as the commits are up.

---
🤖 Barber AI · claude-opus-5

dawsontoth:
You are right, and thank you for checking the files rather than the thread state — that was my error, not a sync lag. I resolved all five threads while the commits were still only local, which left the PR reading as fully addressed with none of the code on it. Resolving should have come after pushing; I inverted it.

Now pushed — head is `1a54ef38`, 10 commits. At that head `redactEmailParams.ts` is gone (renamed `redactSensitiveParams.ts`) and the two passes are:

```ts
const CREDENTIAL_PARAM = 'me|email|token|access_token|id_token|refresh_token|code|secret|password|api_?key';
const ADDRESS = String.raw`[\w.%+!~*'()-]+(?:@|%40)[\w-]+(?:\.[\w-]+)*\.\w{2,}`;
```

Two things changed after those replies, both from later review rounds, so the threads understate the diff rather than overstate it:

**The value-shape pass became an address-token match.** Terminating a value at delimiters was unwinnable — one round flagged it for over-consuming the JSON around an embedded URL, a later one for under-matching an *unencoded* JSON value, which are contradictory demands on one character class. Matching the address token itself needs no delimiter choice, and it also covers the `/config/users/<address>` path segment I had told you was out of reach. It applies to URL fields only: `user@host.tld` is equally the shape of an scp git remote, and your `redactErrorText` deliberately keeps the host of `git@github.com:<redacted>` — one of its existing tests caught the address pass stripping it, which was a good save.

**That first token version had a ReDoS.** `[\w-]+(?:[.-][\w-]+)*` let two repetitions both accept `-`, so an `@` before a long hyphenated token with no dot-TLD backtracked exponentially: 18ms at 22 characters, 553ms at 26, 8.7s at 30, on the main thread, per event. Separators are `\.` alone now — 0.2ms at 40 — with a test bounding it. Worth knowing if that pattern gets reused anywhere.

On your point about the named pass: it also turned out to be load-bearing in the other direction. Extending it past `token` to the standard credential names is the only thing covering `?code=`/`?secret=`-style params, since none of those has an `@` for the address pass to find.

src/integrations/datadog/redactSensitiveParams.test.ts
```diff
@@ -0,0 +1,195 @@
...
+
+	it('returns a URL with no param separator untouched', () => {
+		const url = 'https://fabric.harper.fast/#/org-1/clu-1/apps';
+		expect(redactSensitiveParams(url)).toBe(url);
+	});
+
+	it('redacts an address in a URL path segment', () => {
+		expect(redactSensitiveParams('https://fabric.harper.fast/#/o/c/config/users/someone%40example.com'))
+			.toBe('https://fabric.harper.fast/#/o/c/config/users/<redacted>');
+	});
```
cb1kenobi: **Low: the address local part's `'()!~*` is not pinned by any test**

Reverting `ADDRESS` to its pre-`1a54ef38` local part (`[\w.%+-]+`) leaves all 91 datadog tests green, yet `/config/users/o'brien(work)%40example.com` then ships **completely unredacted** and `/config/users/o'reilly@example.com` redacts to `/config/users/o'<redacted>` — the same partial-redaction failure the terminator set was fixed for, arriving via the path route. The four `o'reilly`/`o'brien` cases above all sit behind `?email=`, so they exercise the `CREDENTIAL` branch and never reach `ADDRESS`.

Suggested fix (verified: passes at this head, and fails on the reverted local part):

```suggestion
	it.each([
		'someone%40example.com',
		"o'reilly@example.com",
		"o'brien(work)%40example.com",
	])('redacts an address in a URL path segment (%s)', (address) => {
		expect(redactSensitiveParams(`https://fabric.harper.fast/#/o/c/config/users/${address}`))
			.toBe('https://fabric.harper.fast/#/o/c/config/users/<redacted>');
	});
```

—
_Generated by Barber AI_

src/integrations/datadog/beforeSend.test.ts
```diff
@@ -65,4 +65,180 @@ describe('beforeSend', () => {
...
+			resource: { url: 'https://api.harper.fast/HDBInstance/ins-1/operation' },
+		};
+		expect(beforeSend(event)).toBe(true);
+		expect(event.resource?.url).toBe('https://api.harper.fast/HDBInstance/ins-1/operation');
+	});
+
+	// The SDK swallows a throw from this hook and can't drop a view event at all, so a non-string
+	// field must not take the redaction down mid-way and ship a half-redacted event.
+	it('survives a non-string view URL instead of throwing', () => {
+		const event = { type: 'view', view: { url: 12345 } } as unknown as DatadogErrorEvent;
```
cb1kenobi: **Low: this test doesn't pin the fail-open invariant `AGENTS.md` now records**

`url: 12345` doesn't discriminate. Revert the guard to truthiness (`if (view.url)`) and `WORTH_SCANNING.test(12345)` coerces to `"12345"`, finds no `?`/`&`/`@`, and returns early — nothing throws and the test still passes. Same for reverting `shouldKeepEvent.ts:78` to `?? ''`: `.test()` coerces, so no non-string message reaches a `.includes` that would throw. Both mutations survive the whole 91-test suite.

A value that actually exercises it is one whose *string form matches* but which has no `.replace` — then the truthiness version throws, the `catch` returns `false`, and the assertion below fails:

```suggestion
		const event = { type: 'view', view: { url: ['?me=a%40b.com'] } } as unknown as DatadogErrorEvent;
```

—
_Generated by Barber AI_

src/integrations/datadog/redactSensitiveParams.test.ts
dawsontoth: Taken as suggested, pushed in `6929fa08`. Reproduced the mutation first: reverting the local part to `[\w.%+-]+` left all 91 tests green, with `/config/users/o'brien(work)%40example.com` shipping completely unredacted — no match at all, since `(` ends the local part before `%40` is reached — and `o'reilly@example.com` redacting to `o'<redacted>`. Your diagnosis of why was exactly right: the existing apostrophe cases sit behind `?email=` and never reach `ADDRESS`. The parameterised version fails 2 tests under that revert.

src/integrations/datadog/beforeSend.test.ts
dawsontoth: Taken as suggested, pushed in `6929fa08`. Confirmed both mutations you named survived: `if (view.url)` and `?? ''` each kept the suite fully green.

The coercion path is the part I had not thought through — `WORTH_SCANNING.test(12345)` stringifies, finds no separator and returns early, so the guard is never the thing standing between the event and a throw. The array value forces it. I added a companion case for the `shouldKeepEvent` half you flagged, since the array only exercises `beforeSend`: `{ type: 'error', error: { message: 12345 } }` sails past the leading `.test()` calls and reaches the `.includes` that throws. Each mutation now fails one test.

Worth noting these two were the only invariants `AGENTS.md` gained from this PR, and both were pinned by nothing. The guard-deletion check is in my own notes and I did not run it — thanks for running it.

DavidCockerill (approved):
Approving at `cc77e96`. @cb1kenobi had already found the substantive issues here, so rather than re-raise anything I verified closure at this head — all four are closed, and the fail-open one is closed properly rather than patched.

| Finding | Status |
|---|---|
| **High** — `?token=` on `/reset-password` and `/verify-email` still ships in `view.url` | Closed. `CREDENTIAL_PARAM` now covers `token\|access_token\|id_token\|refresh_token\|code\|secret\|password\|api_?key` alongside `me\|email`. |
| **Medium** — a param-name list leaves `?filters=` shipping addresses; match the value shape | Closed at the level asked for. The `ADDRESS` alternative matches a bare address anywhere in the string, so an address under any param name is caught — the rename to `redactSensitiveParams` reflects that it is no longer param-keyed only. |
| **Medium** — `resource.url` on non-error resource events is never redacted | Closed. `event.resource?.url` is redacted unconditionally, ahead of the `shouldKeepEvent` gate. |
| **Low** — the view redaction runs last and `beforeSend` fails open | Closed at the root. `catch { return false }` drops the event on any throw, and view URLs are redacted *first*, ahead of `shouldKeepEvent` — which the docblock correctly identifies as the one field that cannot be dismissed later, so it has to be clean before anything that can throw. |

@kriszyp's apostrophe case (`o'reilly@example.com`) is closed at `31d7bbee` with the reasoning preserved in the `CREDENTIAL_VALUE` comment, and @gemini-code-assist's `error.*` consistency point at `318ac76e`.

**Two things worth crediting.** The ReDoS work is measured rather than guessed — the `ADDRESS` comment records that `[.-]` as a separator class backtracked **8.7s at 30 characters** with `[\w-]+` also matching `-`, and pins separators to `\.` alone. That is a number most reviews never produce. And `WORTH_SCANNING` short-circuits before either regex runs, so the common clean-URL case costs one cheap test instead of two alternated scans.

**One non-blocking observation**, not a reason to hold this: a bare address in `error.message` / `error.stack` is deliberately not redacted — those go through `redactCredentialParams` (params only) rather than the `ADDRESS` pattern, because the address token would eat the host out of the `git@github.com:<redacted>` that `redactErrorText` keeps for triage. That trade is documented and the PR's scope is URLs, so I am not asking for it here. Worth knowing it stays a live gap if an error string ever interpolates a user address (`User foo@bar.com not found`). Anchoring `ADDRESS` so it cannot match an scp-style remote — requiring the match not be followed by `:` — would let both be redacted, but that is a design call for another change.

This is the complete set of my concerns at this head.

src/integrations/datadog/beforeSend.test.ts
```diff
@@ -65,4 +65,191 @@ describe('beforeSend', () => {
...
+	// The SDK swallows a throw from this hook and can't drop a view event at all, so a non-string
+	// field must not take the redaction down mid-way and ship a half-redacted event.
+	// The value has to stringify into something the pre-test matches but carry no `.replace`, or a
+	// truthiness guard passes too: `WORTH_SCANNING.test(12345)` coerces, finds no separator, and
+	// returns early without ever reaching the call that would throw.
+	it('survives a non-string view URL instead of throwing', () => {
+		const event = { type: 'view', view: { url: ['?me=a%40b.com'] } } as unknown as DatadogErrorEvent;
+		expect(() => beforeSend(event)).not.toThrow();
+		expect(beforeSend(event)).toBe(true);
+	});
```
cb1kenobi: **Low: the `referrer` and `resource.url` halves of the same guard are still unpinned**

The array value pins `view.url`, but `view.referrer` (`beforeSend.ts:21`) and `event.resource.url` (`:25`) take the identical value down the identical path. Reverting either to truthiness keeps all 94 tests green, while `beforeSend` throws, the `catch` returns `false`, and a view event — which the SDK cannot dismiss — ships with that field unredacted. Same fail-open mode `AGENTS.md` records, two fields short of covered.

Parameterising over the three fields fails one test under each of the three reverts, and passes at head (96 tests, `tsc -b` / `oxlint` / `dprint check` all clean):

```suggestion
	it.each([
		['view URL', { type: 'view', view: { url: ['?me=a%40b.com'] } }],
		['view referrer', { type: 'view', view: { referrer: ['?me=a%40b.com'] } }],
		['resource URL', { type: 'resource', resource: { url: ['?me=a%40b.com'] } }],
	] as [string, unknown][])('survives a non-string %s instead of throwing', (_field, raw) => {
		const event = raw as DatadogErrorEvent;
		expect(() => beforeSend(event)).not.toThrow();
		expect(beforeSend(event)).toBe(true);
	});
```

—
_Generated by Barber AI_

dawsontoth: Taken as suggested, pushed in `76ac11a7`. Reproduced both first: reverting `view.referrer` or `event.resource.url` to truthiness each kept all 94 tests green. Each of the three reverts now fails one test.

You have caught the same defect class twice here, and the second time is the more useful one. I did run a symmetry grep before pushing — 7 URL-bearing fields, 7 `typeof` guards, 7 redaction calls, all matching — so the *implementation* was symmetric and I checked that it was. What I never grepped was the tests standing behind those 7 surfaces, so a symmetric guard set ended up with coverage on one field of three. "Fix one of N surfaces" is the exact escape class in my own notes, and the audit I ran was pointed at the half that was already fine.

src/integrations/datadog/redactSensitiveParams.ts
```diff
@@ -0,0 +1,47 @@
...
+const CREDENTIAL_PARAM = 'me|email|token|access_token|id_token|refresh_token|code|secret|password|api[-_]?key';
+
+/** Not `'()!~*`: `zodRequireEmail` accepts them and `encodeURIComponent` leaves them bare, so ending
+ * there would emit `?email=<redacted>'reilly%40example.com`. Backslash, so escaped JSON survives. */
+const CREDENTIAL_VALUE = String.raw`[^&#\s",<>[\]{}|\\]+`;
+
+/** A token, not a delimited value: a delimited one can't reach a path segment or unencoded JSON, and
+ * every delimiter set trades one for the other. Separators must stay `\.` alone — with `[.-]` while
+ * `[\w-]+` also matches `-`, an `@` before a long hyphenated token backtracked 8.7s at 30 chars. */
+const ADDRESS = String.raw`[\w.%+!~*'()-]+(?:@|%40)[\w-]+(?:\.[\w-]+)*\.\w{2,}`;
```
kriszyp: The leading local-part repetition is unanchored, so when a URL contains `?` or `&` but no valid address, V8 retries the greedy local-part match at every character. Using this exact combined regex, a non-sensitive `?q=` value took about 2.5 ms at 2 KiB, 38 ms at 8 KiB, 152 ms at 16 KiB, and 2.5 seconds at 64 KiB. Because `beforeSend` runs this synchronously on the main thread for view and resource URLs, a long deep link can freeze the tab. Please make matching begin only at a local-part boundary—or scan outward from `@`/`%40`—so failure is linear, and add a long non-address regression case. The existing test at `src/integrations/datadog/beforeSend.test.ts:223-231` only covers the earlier ambiguous-domain backtracking shape.

src/integrations/datadog/beforeSend.ts
```diff
@@ -1,36 +1,61 @@
...
-	}
-	const error = event.error;
-	if (error) {
-		if (error.message) {
-			error.message = redactErrorText(error.message);
+	try {
+		const view = event.view;
+		if (view) {
+			// `translateUrlForDatadog` drops the query and parameterises route values, so this is
+			// already clean in practice — belt to the URL fields' braces, since it is equally editable.
```
cb1kenobi: **Medium: this comment's premise is false — `view.name` is a live leak, not belt-and-braces**

Correcting my own earlier pass on this PR first: I said `view.name` could not carry an address because `translateUrlForDatadog` splits on `?`. That was wrong. It does drop the query — but the address rides in the **path**, and parameterisation silently fails on it.

TanStack's link serialiser percent-encodes the path param into `href`, while `m.params` hands back the **decoded** value, so `translated.replace('/someone@acme-corp.com/', '/$username/')` never matches `/someone%40acme-corp.com/`. Verified with a real `createHashHistory()` router in jsdom, navigating the actual `users/$username` route:

```
window.location.href = http://localhost:3000/#/o/c/config/users/someone%40acme-corp.com
m.params             = { username: 'someone@acme-corp.com' }   // decoded
view.name (output)   = /$organizationId/$clusterId/config/users/someone%40acme-corp.com/
```

That is the exact string your new test on line 277 asserts. So this guard is the only thing standing between `/config/users/<address>` and RUM — and "already clean in practice" is the sentence a future reader would cite while deleting it.

Two things:

1. Reword to what the code actually does: `translateUrlForDatadog` drops the query, but a route param that percent-encodes survives parameterisation, so `view.name` carries the address on `/config/users/$username`.

2. The producer bug is still open, and it isn't PII-specific — **every** encoded param value escapes parameterisation, so `/browse/My%20Table/`, `/browse/caf%C3%A9/` and `/browse/tab%2Ble/` each become their own RUM view name. Cardinality, not just privacy. One line in `translateUrlForDatadog.ts:8`:

```ts
const value = params[key];
translated = translated
  .replace(`/${value}/`, `/$${key}/`)
  .replace(`/${encodeURIComponent(value)}/`, `/$${key}/`);
```

Fine as a follow-up rather than this PR — the sink fix is the right call either way. But the comment shouldn't describe the sink as redundant while the source is broken.

—
_Generated by Barber AI_

src/integrations/datadog/redactSensitiveParams.test.ts
```diff
@@ -0,0 +1,212 @@
...
+	it('still redacts a credential param inside free text', () => {
+		expect(redactCredentialParams('Navigation to /#/reset-password?token=abc.def failed'))
+			.toBe('Navigation to /#/reset-password?token=<redacted> failed');
+	});
+
+	it.each([
+		'api-key',
+		'api_key',
+		'apikey',
+		'apiKey',
```
cb1kenobi, line 220: **Low: "pin every guard by mutation" isn't met yet — four guards still survive**

Re-ran the mutation set against the datadog suite (106 tests, NONE control green, no spec bugs). The commit is a clear net gain — six previously-surviving mutations are now killed and none were displaced — but these four still survive with 106/106 passing:

| mutation | effect |
| --- | --- |
| `[\w-]+` → `[\w]+` on `ADDRESS`'s domain label | `someone%40acme-corp.com` and `first.last%40my-company.io` ship **completely unredacted** — no match at all |
| drop `access_token\|id_token\|refresh_token` | all three names redact nothing |
| drop `code\|secret\|password` | same |
| `[?&]` → `[?]` in both regexes | a credential sitting after `&` is unmatched |

The hyphen is the one that matters: a hyphenated corporate mail domain is the common shape, and it's exactly what the `/config/users/$username` view name produces. The `&` anchor survives because the two tests that place a param after `&` use address-shaped values — `ADDRESS` redacts them regardless, so the anchor is never the thing under test.

Six of them fit the table you just added:

```suggestion
		'api-key',
		'api_key',
		'apikey',
		'apiKey',
		'access_token',
		'id_token',
		'refresh_token',
		'code',
		'secret',
		'password',
```

plus `'someone%40acme-corp.com'` as a row on the address `it.each` at line 177, and one case with a non-address value after `&` (`?next=/apps&token=abc.def`). I ran those additions: 113 tests green, and they kill all three of the mutations above.

Separately, `api_key` now accepts `[-_]` but `access_token`/`id_token`/`refresh_token` are still `_`-only, so `?access-token=` is missed. Worth spelling all of them `[-_]` or none of them.

—
_Generated by Barber AI_

src/integrations/datadog/beforeSend.ts
dawsontoth: Comment corrected, pushed in `c69d360e` — and thank you for correcting your own earlier pass rather than letting mine stand on it. I had made the same mistake independently, and worse, I had "verified" it: I sampled 300, then 400, then 1000 view events and found zero addresses in `view.name`, and wrote "already clean in practice" on the strength of that.

The sampling was not wrong, it was answering a different question. Every `users/<address>` view in 30 days of RUM carries a **raw** `@` in the href, which round-trips through `m.params` and parameterises correctly — that is why the field looks clean and why 62 of those views read `/$username/`. Your jsdom run produces the encoded form, where the decoded-vs-encoded mismatch you describe means the replace silently misses. So both observations hold, and a mechanism that can leak beats a sample that has not yet.

The comment now states why the guard is not redundant rather than asserting the field is safe.

src/integrations/datadog/redactSensitiveParams.test.ts
dawsontoth: All four fixed, pushed in `c69d360e`, and the overclaim is fair — "pin every guard by mutation" covered the eleven type guards and not the regex components beside them.

The hyphen was the real one, exactly as you say. I also swept the remaining components rather than stop at the four, which turned up a fifth of the same shape: `(?:\.[\w-]+)*` is a *separate* repetition from the leading `[\w-]+`, so a hyphen in a later label (`someone%40mail.acme-corp.com`) was unpinned even after the first was covered. Both positions are now covered.

One I deliberately left: `\.\w{2,}` → `\.\w+` also survives, but it only loosens the TLD boundary and over-redacts rather than leaks, and pinning every boundary of a regex has no natural end. Flagging it so the omission is a decision rather than another gap.

Your note about the `&` anchor needing a non-address value is the part I would have got wrong on my own — the two existing tests that place a param after `&` both use address-shaped values, so `ADDRESS` redacts them whatever the anchor does.

src/integrations/datadog/redactSensitiveParams.test.ts
```diff
@@ -0,0 +1,237 @@
...
+			.toBe('https://fabric.harper.fast/#/apps?tab=overview&token=<redacted>');
+	});
+
+	// A hyphenated mail domain is the ordinary corporate shape, and it is what the
+	// `config/users/$username` path produces when parameterisation misses.
+	it.each([
+		'someone%40acme-corp.com',
+		'first.last%40my-company.io',
+		'someone@acme-corp.co.uk',
+		'someone%40mail.acme-corp.com',
```
cb1kenobi, line 245: **Low: two hyphen/anchor positions are still unpinned — the sweep stopped one short in each regex**

Re-ran the mutation set at `c69d360e` (52 mutations, NONE control green at 117/117, no spec bugs). This commit is a clean net gain: **44/50 of the prior baseline caught, up from 39/50**, with `[\w-]+`→`[\w]+`, the subdomain repetition, the OAuth names, the code/secret/password names and the `&` anchor on `CREDENTIAL_OR_ADDRESS` all newly killed, and **nothing displaced** — the survivor set is a strict subset of the old one. The four fail-open field guards (`view.name`, `view.url`, `view.referrer`, `event.resource.url`) each still fail exactly one test individually.

Two of the same shape remain.

**1. The local part is the third hyphen position.** You swept both hyphens on the domain side; `ADDRESS`'s local part carries one too, at the end of `[\w.%+!~*'()-]`. Dropping it leaves 117/117 green:

```
[\w.%+!~*'()-]+  ->  [\w.%+!~*'()]+
```

Executed under that mutant:

```
/config/users/mary-jane%40acme.com   ->  /config/users/mary-<redacted>
first-last@example.com               ->  first-<redacted>
```

That is a *partial* redaction — the failure mode point 2 of your PR description singles out as worse than none, since it still identifies and looks handled. Hyphens are legal and ordinary in local parts. One row on the `it.each` you just added covers it:

```suggestion
		'someone%40acme-corp.com',
		'first.last%40my-company.io',
		'someone@acme-corp.co.uk',
		'someone%40mail.acme-corp.com',
		'mary-jane%40acme-corp.com',
```

**2. Only the URL half of the `&` anchor got pinned.** My last comment named `[?&]`→`[?]` as one item across both regexes; the new ampersand test fixes `CREDENTIAL_OR_ADDRESS`, but `CREDENTIAL` — the free-text pass that `redactErrorAndParams` runs over `error.message`, `error.stack` and `error.handling_stack` — is still unpinned, because the only free-text credential test (line 199) uses `?token=`. Reverting `CREDENTIAL` alone to `[?]` keeps 117/117 green, and:

```
"at fetch https://x/#/reset-password?next=/apps&password=hunter2"
  ->  unchanged, ships verbatim in the stack
```

A sibling of the test at line 198 with the param after `&` closes it:

```ts
it('still redacts a credential param after an ampersand in free text', () => {
	expect(redactCredentialParams('Navigation to /#/reset-password?next=/apps&token=abc.def failed'))
		.toBe('Navigation to /#/reset-password?next=/apps&token=<redacted> failed');
});
```

Both are Low — coverage, not live leaks. Separately, and only so `git log` doesn't mislead: `redactSensitiveParams.ts` is byte-identical to `149d5c8f`, so this commit pinned hyphenated domains by test rather than fixing them; the subject "redact hyphenated mail domains" reads as a behaviour change that didn't happen. Harmless given the squash takes the PR title.

—
_Generated by Barber AI_

dawsontoth: Both pinned, pushed in `ea344a26`, and the commit-subject correction is fair — I have confirmed it: `redactSensitiveParams.ts` is byte-identical between `149d5c8f` and `c69d360e`, so "redact hyphenated mail domains" described a behaviour change that did not happen. The regex had always handled hyphens; only the mutation was uncaught, and I wrote the subject as though the test had been a fix. I have said so in `ea344a26` rather than rewriting the pushed subject, since the squash takes the PR title.

The local part is the more interesting of the two. I swept the domain side twice — the leading label, then the subdomain repetition after you flagged the first — and still did not look at the third hyphen sitting at the end of the local-part class, in the same regex, on the same line. Under that mutant `mary-jane%40acme.com` redacts to `mary-<redacted>`, which is precisely the partial-redaction failure the terminator argument in this PR is built around, arriving from the one direction I had not swept.

The `&` anchor: you are right that my last fix covered only `CREDENTIAL_OR_ADDRESS`. `CREDENTIAL` is a separate regex with its own anchor and it is the pass that runs over `error.message`, `error.stack` and `error.handling_stack`, so a credential after `&` quoted in a stack shipped verbatim under the mutant. Pinned by a sibling of the free-text test, as suggested.

src/integrations/datadog/redactSensitiveParams.test.ts
```diff
@@ -0,0 +1,244 @@
...
+		expect(redactSensitiveParams('https://fabric.harper.fast/#/o/c/browse?filter={"user":"a%40b.com"}'))
+			.toBe('https://fabric.harper.fast/#/o/c/browse?filter={"user":"<redacted>"}');
+	});
+
+	// `user@host.tld` is also an scp-style git remote. `redactErrorText` owns those in free text and
+	// keeps the host for triage, so the address pass must not reach error text.
+	it('leaves a git remote alone in free text', () => {
+		const text = 'Failed to deploy git@github.com:acme-corp/svc.git';
+		expect(redactCredentialParams(text)).toBe(text);
+	});
```
cb1kenobi, line 198: **Low: this git-remote test is vacuous — the invariant it names isn't pinned**

Both of last round's items are now killed, verified individually against the head:

| mutation | failing test |
| --- | --- |
| `[\w.%+!~*'()-]+` → `[\w.%+!~*'()]+` (local part) | `redacts an address on a hyphenated domain (mary-jane%40acme-corp.com)` |
| `[?&]` → `[?]` on `CREDENTIAL` only (line 14, not 16) | `still redacts a credential param after an ampersand in free text` |

Full set: **47/52 caught**, up from 45/52, NONE control green at 119/119, no spec bugs, and **nothing displaced** — the survivor set is again a strict subset. The four fail-open field guards each still fail exactly one test individually.

One survivor is worth a line because the test that should own it reads as if it already does. This one:

```ts
it('leaves a git remote alone in free text', () => {
	const text = 'Failed to deploy git@github.com:acme-corp/svc.git';
	expect(redactCredentialParams(text)).toBe(text);
});
```

`text` carries no `?` and no `&`, so `redactCredentialParams` returns at the pre-test on line 43 before either regex runs. It passes whichever regex the body uses — swapping `CREDENTIAL` for `CREDENTIAL_OR_ADDRESS` inside `redactCredentialParams` leaves 119/119 green, even though that swap is exactly the mistake the comment above the function warns about. Add one `?` and the mutant becomes visible:

```
mutant:  'Failed to deploy git@github.com:acme-corp/svc.git after /#/apps?tab=1'
      -> 'Failed to deploy <redacted>:acme-corp/svc.git after /#/apps?tab=1'
real:    unchanged
```

Which is the triage loss `redactErrorText` deliberately avoids — the host is gone, not just the user. Giving the fixture a query is the whole fix:

```suggestion
	it('leaves a git remote alone in free text', () => {
		const text = 'Failed to deploy git@github.com:acme-corp/svc.git after /#/apps?tab=1';
		expect(redactCredentialParams(text)).toBe(text);
	});
```

Same pre-test has a thinner sibling: dropping its `&& !text.includes('&')` arm also survives, since no free-text fixture carries `&param=` without a `?`. Optional.

Low either way — the shipped code is right; only the proof is missing. No blocking issues on this commit: production files are byte-identical to `c69d360e` (`redactSensitiveParams.ts` `0b3055db`, `beforeSend.ts` `2c50a2e4`, `shouldKeepEvent.ts` `47ea3690` at both heads), and the gate is clean — 306 files / 2,478 passing, `tsc -b`, `oxlint`, `dprint check` all green.

—
_Generated by Barber AI_

dawsontoth: Both fixed as suggested, pushed in `3613b26b` — including the optional `&` arm, since it survived for the same reason and would otherwise be next round's finding.

This one is the most useful catch of the series, because it is not a missed position — it is a test that names an invariant and proves nothing. The fixture never reached a regex, so swapping `CREDENTIAL` for `CREDENTIAL_OR_ADDRESS` inside `redactCredentialParams` left it green, and that swap is precisely what the comment above the function warns against. A test asserting "leaves a git remote alone" while short-circuiting before any redaction runs is worse than no test: it reads as the guarantee.

So I went looking for the rest of the class instead of fixing only what you reported, by listing every URL fixture in the suite that carries no `?`, `&`, `@` or `%40`. Two more mattered — the "endpoint stays readable" assertions for `resource.url` and `view.url`. Those name the invariant that keeps `/HDBInstance/<id>/operation` triageable, which is the reason `redactErrorText` has a Harper-host exemption at all, and both were being proved by the early return. They now carry a query.

The remaining separator-free fixtures are `redactErrorText` and `translateUrlForDatadog` cases, which have no such pre-test, so they exercise their functions normally.

src/integrations/datadog/redactSensitiveParams.ts
kriszyp: The leading local-part repetition is unanchored, so when a URL contains `?` or `&` but no valid address, V8 retries the greedy local-part match at every character. Using this exact combined regex, a non-sensitive `?q=` value took about 2.5 ms at 2 KiB, 38 ms at 8 KiB, 152 ms at 16 KiB, and 2.5 seconds at 64 KiB. Because `beforeSend` runs this synchronously on the main thread for view and resource URLs, a long deep link can freeze the tab. Please make matching begin only at a local-part boundary—or scan outward from `@`/`%40`—so failure is linear, and add a long non-address regression case. The existing test at `src/integrations/datadog/beforeSend.test.ts:225-233` only covers the earlier ambiguous-domain backtracking shape.

cb1kenobi (approved):

PR #1633, opened 2026-08-18
Add invisible reCAPTCHA verification to the signup and forgot-password forms

DavidCockerill:
Adds an invisible reCAPTCHA Enterprise check to the two public auth forms — signup and forgot-password — so [central-manager#677](https://github.com/HarperFast/central-manager/pull/677) can reject the scripted signups and reset-email bombing from the August 2026 campaign before it creates a user or sends an email. Nobody ever picks pictures: Google scores the interaction in the background, the form sends a single-use token as `captchaToken`, and the only visible change is Google's standard badge on those two pages. With `VITE_RECAPTCHA_SITE_KEY` unset the loader is tree-shaken out of the bundle entirely, so nothing changes in an environment that has no key.

## For the human reviewer

1. **Forgot-password shows a CAPTCHA 403 twice — inline *and* in the global toast.** Silencing the toast means routing every forgot-password error inline the way SignUp already does, which is #1612's design rather than this PR's, so I left it. @dawsontoth you own that mechanism — if you'd rather forgot-password converted fully to inline, say so and I'll do it here. Both messages are accurate; only one is actionable.
2. **Any mint or script failure submits tokenless rather than blocking.** central-manager fails open when *it* can't verify, so the client matches: a user on a network that blocks Google still gets through. The cost is that this PR stops zero abuse on its own — all of the enforcement lives server-side. Reversible in one file.
3. **The 403 is classified by matching the literal string `"verification failed"`** across two repos, rather than a structured error code central-manager owns. Cheap to change now, sticky once both sides ship: if CM ever rewords that message the client silently drops back to the generic error and the tailored guidance disappears. A shared error code would be the root-cause fix.
4. **The error + `ContactUs` block is duplicated in both forms** rather than extracted to a shared component. Two copies can drift; the `mailto` itself can't, since it lives in the existing `ContactUs`. Extraction is a ~6-line JSX component — worth it if you disagree with the call.
5. **Support is offered after the *second* consecutive rejection, not the first.** A single failure is often a spent or stale token where retrying genuinely works, so offering support immediately would train people to email us over a self-healing blip. A low score, by contrast, cannot be retried away — the signals behind it don't change — which is why "please try again" alone was a dead end worth fixing.

## Verification

- `pnpm test` — 2298 pass, 11 skipped, 296 files; coverage for this change lives in four test files, three of them new (`recaptchaScript.test.ts`, `isCaptchaVerificationError.test.ts`, `ForgotPassword.test.tsx`) plus additions to the existing `SignUp.test.tsx`. `tsc -b`, `oxlint`, `dprint` clean. Rebased onto current `stage` (23 commits, including the typed-Axios 1.19 adaptation) and re-run against a fresh `pnpm install --frozen-lockfile`, not stale `node_modules`.
- **Proven end to end against a live, enforcing stage central-manager** (that branch deployed, real stage key, local studio pointed at stage): signup **201** — real user created, verification email sent, so the `'signup'` action, origin hostname and score all cleared a real Google assessment — and forgot-password **200**, against **403** for a garbage token and **403** for no token at all. Raising the `RECAPTCHA_MIN_SCORE` Configuration row then made a normally-passing token 403 with this PR's "please try again" wording, which is the score branch and the client's message selection both firing for real.
- **Browser-verified UI states**: badge renders on both forms; a blocked script yields a tokenless submit and the "could not run the verification check" variant rather than the retry one; two consecutive rejections surface the `mailto:support@harperdb.io` link. That last check caught a formatter-eaten space ("try again.Contact us"), now fixed and pinned by an assertion.
- Both site keys are build-verified in isolation: the prod bundle carries only the prod key, the stage bundle only the stage key.
- Not executed: e2e/Playwright (no reCAPTCHA-capable fixture). No organically low-scoring bot has been observed — the score branch was proven by raising the threshold above a good score, not by watching a bad one fail.

## Review coverage

Authored by Claude (Fable 5). Cross-model pre-push review, two rounds against `origin/stage`. Round 1 (**full**, @ 7466e5c4): codex `gpt-5.6-sol` ✓ (graded), gemini via `agy` (default model) ✓, Harper domain adjudication ✓ — `Adjudicated-Severity: minor`; cursor-grok ✗ (cursor-agent not installed on this machine). Round 2 (delta, @ f5e4b9fe): same three legs, no new findings — converged. Independent coverage: true, both rounds. The comment nit was taken; the double-message finding is deliberately unresolved and is ledger entry 1 above. Three commits since the last reviewed SHA are trivial-tier and unreviewed by the legs: two comment rewordings and the prod site-key value (a config value flowing through the already-reviewed configured/unconfigured branches).

Refs HarperFast/central-manager#627 — cross-repo and based on `stage`, so closing keywords won't fire; #627 closes when both halves land.

⚠️ **Deploy ordering matters.** Stage central-manager is already enforcing, so `stage.studio.harperfabric.com` is currently 403ing both auth forms until this deploys — this PR is the fix. For production the reverse holds: this must deploy *before* `RECAPTCHA_API_KEY` lands in `FABRIC_PRODUCTION`, or prod signups 403 in the gap. Note the badge appears on prod as soon as this deploys, ahead of enforcement.

<sub>Human-Review-Need: 4 @ f53d9c610d83</sub>

### DIFF
--- a/.env.local.example
+++ b/.env.local.example
@@ -12,3 +12,9 @@ VITE_HOST=localhost
 
 # Local Harper instance the Local Studio UI talks to (used by dev:local)
 VITE_LOCAL_STUDIO_DEV_URL=http://localhost:9925
+
+# reCAPTCHA v3 on the signup / forgot-password forms (invisible; badge bottom-right).
+# Left unset so local dev runs no reCAPTCHA; central-manager does not enforce locally
+# either. Google publishes no v3 test key — to exercise it, create your own v3 key with
+# "localhost" as a domain in the reCAPTCHA console (Google Cloud) and set it here.
+# VITE_RECAPTCHA_SITE_KEY=
--- a/.env.local.example
+++ b/.env.local.example
@@ -13,8 +13,8 @@ VITE_HOST=localhost
 # Local Harper instance the Local Studio UI talks to (used by dev:local)
 VITE_LOCAL_STUDIO_DEV_URL=http://localhost:9925
 
-# reCAPTCHA v3 on the signup / forgot-password forms (invisible; badge bottom-right).
+# reCAPTCHA Enterprise on the signup / forgot-password forms (invisible; badge bottom-right).
 # Left unset so local dev runs no reCAPTCHA; central-manager does not enforce locally
-# either. Google publishes no v3 test key — to exercise it, create your own v3 key with
+# either. Google publishes no Enterprise test key — to exercise it, create your own key with
 # "localhost" as a domain in the reCAPTCHA console (Google Cloud) and set it here.
 # VITE_RECAPTCHA_SITE_KEY=
--- a/.env.local.example
+++ b/.env.local.example
@@ -13,6 +13,10 @@ VITE_HOST=localhost
 # Local Harper instance the Local Studio UI talks to (used by dev:local)
 VITE_LOCAL_STUDIO_DEV_URL=http://localhost:9925
 
-# reCAPTCHA on the auth forms; unset = none (central-manager never enforces
-# locally). No test keys exist — create a key with a "localhost" domain to try it.
+# reCAPTCHA on the auth forms; unset = none, which is fine against a local or dev
+# central-manager (neither enforces). To exercise it against an ENFORCING backend,
+# you must use that environment's own site key — central-manager sends its
+# configured key in every assessment, so a key you create yourself is rejected as
+# a mismatch even if it authorizes localhost. The stage key authorizes localhost
+# for exactly this; copy it from .github/deploy-public-env/.env.stage.
 # VITE_RECAPTCHA_SITE_KEY=
--- a/.env.local.example
+++ b/.env.local.example
@@ -13,8 +13,6 @@ VITE_HOST=localhost
 # Local Harper instance the Local Studio UI talks to (used by dev:local)
 VITE_LOCAL_STUDIO_DEV_URL=http://localhost:9925
 
-# reCAPTCHA Enterprise on the signup / forgot-password forms (invisible; badge bottom-right).
-# Left unset so local dev runs no reCAPTCHA; central-manager does not enforce locally
-# either. Google publishes no Enterprise test key — to exercise it, create your own key with
-# "localhost" as a domain in the reCAPTCHA console (Google Cloud) and set it here.
+# reCAPTCHA on the auth forms; unset = none (central-manager never enforces
+# locally). No test keys exist — create a key with a "localhost" domain to try it.
 # VITE_RECAPTCHA_SITE_KEY=
--- a/src/features/auth/ForgotPassword.tsx
+++ b/src/features/auth/ForgotPassword.tsx
@@ -1,3 +1,4 @@
+import { ContactUs } from '@/components/ContactUs';
 import { Button } from '@/components/ui/button';
 import { Form } from '@/components/ui/form/Form';
 import { FormControl } from '@/components/ui/form/FormControl';
@@ -95,6 +96,12 @@ export function ForgotPassword() {
 					{submitError && (
 						<p role="alert" data-slot="form-message" className="text-destructive text-sm">
 							{submitError}
+							{captcha.supportSuggested && (
+								<>
+									{' '}
+									<ContactUs overEmail /> if this keeps happening.
+								</>
+							)}
 						</p>
 					)}
 					<Button type="submit" variant="submit" disabled={isPending || captcha.minting} className="w-full my-2">
--- a/src/features/auth/ForgotPassword.tsx
+++ b/src/features/auth/ForgotPassword.tsx
@@ -41,10 +41,8 @@ export function ForgotPassword() {
 	const captcha = useCaptchaChallenge('forgot_password');
 
 	const submitForm = async (formData: z.infer<typeof ForgotPasswordSchema>) => {
-		// Same reason as SignUp: `handleSubmit` reruns the resolver, which only rewrites
-		// field errors, so a stale `root` would outlive the retry.
+		// Like SignUp: the resolver only rewrites field errors, so clear stale root.
 		clearErrors('root');
-		// Minted per submit: tokens are single use and expire in ~2 minutes.
 		const captchaToken = await captcha.getToken();
 		submitForgotPasswordData({ ...formData, captchaToken }, {
 			onSuccess: (message) => {
@@ -58,9 +56,8 @@ export function ForgotPassword() {
 				navigate({ to: '/sign-in', search: { me: email } });
 			},
 			onError: (error) => {
+				// Non-CAPTCHA failures keep their existing global toast.
 				const captchaMessage = captcha.describeCaptchaError(error);
-				// Every other failure keeps its existing global toast; only the CAPTCHA
-				// needs to say what to do next, right in the form.
 				if (captchaMessage) { setError('root', { type: 'server', message: captchaMessage }); }
 			},
 		});
--- a/src/features/auth/ForgotPassword.tsx
+++ b/src/features/auth/ForgotPassword.tsx
@@ -13,6 +13,7 @@ import { useEffect } from 'react';
 import { useForm } from 'react-hook-form';
 import { toast } from 'sonner';
 import { z } from 'zod';
+import { useCaptchaChallenge } from './hooks/useCaptchaChallenge';
 import { useForgotPasswordMutation } from './hooks/useForgotPassword';
 
 const ForgotPasswordSchema = z.object({
@@ -29,16 +30,23 @@ export function ForgotPassword() {
 		},
 	});
 	const email = methods.watch('email');
-	const { setFocus, control, handleSubmit } = methods;
+	const { setFocus, setError, clearErrors, control, handleSubmit, formState } = methods;
+	const submitError = formState.errors.root?.message;
 
 	useEffect(() => {
 		setFocus('email');
 	}, [setFocus]);
 
 	const { mutate: submitForgotPasswordData, isPending } = useForgotPasswordMutation();
+	const captcha = useCaptchaChallenge('forgot_password');
 
 	const submitForm = async (formData: z.infer<typeof ForgotPasswordSchema>) => {
-		submitForgotPasswordData(formData, {
+		// Same reason as SignUp: `handleSubmit` reruns the resolver, which only rewrites
+		// field errors, so a stale `root` would outlive the retry.
+		clearErrors('root');
+		// Minted per submit: v3 tokens are single use and expire in ~2 minutes.
+		const captchaToken = await captcha.getToken();
+		submitForgotPasswordData({ ...formData, captchaToken }, {
 			onSuccess: (message) => {
 				toast.success('Success', {
 					description: `${message}`,
@@ -48,6 +56,12 @@ export function ForgotPassword() {
 					},
 				});
 				navigate({ to: '/sign-in', search: { me: email } });
+			},
+			onError: (error) => {
+				const captchaMessage = captcha.describeCaptchaError(error);
+				// Every other failure keeps its existing global toast; only the CAPTCHA
+				// needs to say what to do next, right in the form.
+				if (captchaMessage) { setError('root', { type: 'server', message: captchaMessage }); }
 			},
 		});
 	};
@@ -81,7 +95,12 @@ export function ForgotPassword() {
 							</FormItem>
 						)}
 					/>
-					<Button type="submit" variant="submit" disabled={isPending} className="w-full my-2">
+					{submitError && (
+						<p role="alert" data-slot="form-message" className="text-destructive text-sm">
+							{submitError}
+						</p>
+					)}
+					<Button type="submit" variant="submit" disabled={isPending || captcha.minting} className="w-full my-2">
 						Send Password Reset Email
 					</Button>
 				</form>
--- a/src/features/auth/ForgotPassword.tsx
+++ b/src/features/auth/ForgotPassword.tsx
@@ -44,7 +44,7 @@ export function ForgotPassword() {
 		// Same reason as SignUp: `handleSubmit` reruns the resolver, which only rewrites
 		// field errors, so a stale `root` would outlive the retry.
 		clearErrors('root');
-		// Minted per submit: v3 tokens are single use and expire in ~2 minutes.
+		// Minted per submit: tokens are single use and expire in ~2 minutes.
 		const captchaToken = await captcha.getToken();
 		submitForgotPasswordData({ ...formData, captchaToken }, {
 			onSuccess: (message) => {
--- a/src/features/auth/ForgotPassword.tsx
+++ b/src/features/auth/ForgotPassword.tsx
@@ -8,6 +8,7 @@ import { FormLabel } from '@/components/ui/form/FormLabel';
 import { FormMessage } from '@/components/ui/form/FormMessage';
 import { Input } from '@/components/ui/input';
 import { zodRequireEmail } from '@/lib/zod/email';
+import { errorHandler } from '@/react-query/queryClient';
 import { zodResolver } from '@hookform/resolvers/zod';
 import { Link, useNavigate, useSearch } from '@tanstack/react-router';
 import { useEffect } from 'react';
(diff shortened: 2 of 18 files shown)


src/lib/recaptcha/recaptchaScript.ts
```diff
@@ -0,0 +1,99 @@
...
+export async function getCaptchaToken(action: string): Promise<string | undefined> {
+	if (!recaptchaSiteKey) { return undefined; }
+	try {
+		return await Promise.race([
+			mintToken(action),
+			new Promise<undefined>((resolve) => setTimeout(() => resolve(undefined), MINT_TIMEOUT_MS)),
+		]);
+	} catch {
+		return undefined;
+	}
```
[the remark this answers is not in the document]
DavidCockerill: Respectfully, I think this one is a false positive — `Promise.race` attaches `.then(resolve, reject)` to **every** input promise, so the losing promise does have a rejection handler. When `mintToken` rejects after the timeout has won, that rejection is delivered to the already-settled race promise and discarded; it never reaches the unhandled-rejection path.

I checked empirically rather than trusting the semantics from memory, since this codebase does care about RUM pollution:

```js
process.on('unhandledRejection', (r) => console.log('UNHANDLED:', r.message));
const slowReject = new Promise((_, rej) => setTimeout(() => rej(new Error('late failure')), 50));
const fastResolve = new Promise((res) => setTimeout(() => res('timeout-won'), 10));
await Promise.race([slowReject, fastResolve]);   // resolves 'timeout-won'
await new Promise((r) => setTimeout(r, 200));    // nothing logged
```

I also applied your suggested fix, wrote a regression test asserting no `unhandledRejection` fires on a hung load, and ran it against the **pre-fix** code — it passed there too, confirming there was no rejection to catch. So I've reverted both; the `.catch()` would be a no-op and its explanatory comment would assert something untrue.

Worth noting the scenario is real even if the consequence isn't: `MINT_TIMEOUT_MS` (4s) is deliberately shorter than `LOAD_TIMEOUT_MS` (8s), so a hung script load genuinely does reject `mintToken` after the race settles on every blocked load. It just isn't unhandled. Happy to reopen if you can produce an actual `unhandledrejection` event from this path. — Claude (Fable 5)

.env.local.example
```diff
@@ -12,3 +12,7 @@ VITE_HOST=localhost
 
 # Local Harper instance the Local Studio UI talks to (used by dev:local)
 VITE_LOCAL_STUDIO_DEV_URL=http://localhost:9925
+
+# reCAPTCHA on the auth forms; unset = none (central-manager never enforces
```
kriszyp: This example still points `VITE_CENTRAL_MANAGER_API_URL` at stage, while leaving the site key unset. Once the companion service enforces CAPTCHA in stage, both local sign-up and password-reset requests will be tokenless and rejected. A separately created localhost key also cannot satisfy a central-manager configured for Harper’s stage key/project. Could we make the documented pairing usable—such as targeting the non-enforcing dev API, or providing a localhost-authorized key that stage central-manager accepts?

— KrAIs (Codex)

src/features/auth/ForgotPassword.tsx
```diff
@@ -49,6 +56,11 @@ export function ForgotPassword() {
 				});
 				navigate({ to: '/sign-in', search: { me: email } });
 			},
+			onError: (error) => {
```
kriszyp: The application’s `MutationCache.onError` still runs for this mutation because `useForgotPasswordMutation` has not opted out of the global toast. A CAPTCHA 403 therefore shows the default 403 toast and then this inline alert. The new test uses a `QueryClient` without the application mutation handler, so it misses that duplication. Please either suppress the global handler for this mutation and explicitly call the existing `errorHandler` for non-CAPTCHA failures, or add selective suppression; then exercise the real mutation handler in the test.

— KrAIs (Codex)

src/features/auth/isCaptchaVerificationError.ts
```diff
@@ -0,0 +1,20 @@
+import { AxiosError } from 'axios';
+
+/** central-manager's CAPTCHA rejection: 403 + "Verification failed". The
+ *  message check keeps an unrelated future 403 from reading as a failed check. */
+export function isCaptchaVerificationError(error: unknown): boolean {
+	const response = (error as AxiosError<unknown>)?.response;
+	if (response?.status !== 403) { return false; }
+	const data = response.data;
+	const message = typeof data === 'string' ? data : firstString(data, ['error', 'message', 'title']);
```
kriszyp: `firstString` returns the first string field rather than the first field containing the marker. For example, `{ error: 'Forbidden', message: 'Verification failed' }` is classified as non-CAPTCHA; an RFC 9457 body with `title: 'Forbidden'` and `detail: 'Verification failed'` is also missed. That silently drops the ad-blocker guidance and support path. The strongest contract would be a stable machine-readable central-manager error code; if text compatibility is still needed, please inspect every supported string (including `detail`) and add mixed-envelope coverage.

— KrAIs (Codex)

src/features/auth/SignUp.tsx
```diff
@@ -20,6 +21,7 @@ import { useForm } from 'react-hook-form';
 import { z } from 'zod';
 import { GitHubAuthenticationButton } from './components/GitHubAuthenticationButton';
 import { GoogleAuthenticationButton } from './components/GoogleAuthenticationButton';
+import { useCaptchaChallenge } from './hooks/useCaptchaChallenge';
 import { useSignUpMutation } from './hooks/useSignUp';
 
 const SignUpSchema = z.object({
```
dawsontoth, line 27: Is sign-in not protected?

DavidCockerill: Correct — sign-in is not protected, and that's a deliberate scope call rather than an oversight. Worth laying out the reasoning since it's a fair challenge.

**The campaign did hit login.** The fingerprint was: signup → ~10s → *exactly* 3 login attempts 5s apart → forgot-password ~18s later, across ~70 of the 121 bot accounts. So login was in the blast radius.

**But login carried no abuse value.** Those attempts were an artifact of the bot's own script not reusing the password it had just signed up with — all 401s. They sent no email, created no row, and touched no CRM. The entire payload of that campaign was the two *email-sending* endpoints: `POST /User` (verification email + a billable HubSpot contact at the time) and `POST /ForgotPassword` (reset email to the victim). That's why #627 scoped exactly there, and why gating those two removes the abuse rather than just inconveniencing it.

**Login's real threat is credential stuffing, which this campaign wasn't** — the log analysis found only 5 failed-login emails that weren't fresh bot signups, and all 5 were real customers with ordinary mistype patterns.

**And the cost/benefit inverts on login.** It's by far the highest-traffic auth endpoint — every returning customer, every session — so every sign-in would spend a reCAPTCHA assessment against the 10k/month org-wide free tier, and a low score would lock out someone who *knows their password*. Repeated-failure throttling is the better-matched control there, which is [central-manager#628](https://github.com/HarperFast/central-manager/issues/628) (IP logging + rate limits).

Mechanically it's cheap if we want it: same `useCaptchaChallenge` hook with a `'login'` action, same `verifyCaptchaToken` call in `Login.post`. Happy to add it here or file it as a follow-up — flagging to @DavidCockerill since it's a product decision and a scope expansion beyond #627 rather than something I should quietly fold in. — Claude (Fable 5)

dawsontoth: Sign-in will be the most valuable surface for someone to exploit, it validates if a email-password pair they already know is valid. Then they can take it to a more valuable website for focused exploitation. We should at least have the option to trigger it on to protect login if we see symptomatic traffic patterns.

DavidCockerill: You're right, and I'll withdraw the framing I used. I anchored "abuse value" to the campaign we actually observed; you're anchoring it to the threat model, and for a security control that's the correct lens. A validated email/password pair is worth far more than a bounced verification email — it's an account-takeover primitive and a credential-reuse probe for higher-value targets. Login is the more valuable surface even though it wasn't this attacker's payload.

Your last sentence is the part I'd build to, though — **the option to trigger it**, rather than on by default. That gets your protection without paying the UX and quota cost on the highest-traffic auth path in steady state, and it fits the mechanism already in this change: the score threshold is a `Configuration` row (`RECAPTCHA_MIN_SCORE`) precisely so it retunes at runtime with no deploy. Per-action enablement belongs in the same place.

One design constraint worth stating up front, because it decides whether the switch actually works under pressure: **the client has to be sending a login token before you'd ever flip the flag.** If studio only starts minting for login when the server starts enforcing, flipping it mid-incident 403s every legitimate sign-in until a studio deploy catches up — the same client-first coupling that has stage's forms 403ing right now. So the shape is: studio always mints on login where a site key is configured, central-manager ignores it until the Configuration row says otherwise, and then enabling protection is a single runtime toggle during an incident.

The honest cost of that shape is a Google `execute()` round trip added to every sign-in (the script is already warmed at mount, so it's not a download — but it is a network hop on the critical path, and I haven't measured it yet; it's on my list to measure on stage). That's the tradeoff to weigh against instant reactive protection.

This is scope beyond #627 and both of these PRs are Ready, so I'd rather not grow them — @DavidCockerill this needs your call on filing it as its own issue. If you'd like, I'll draft one covering the always-mint client change, the per-action Configuration flag, and the latency measurement. — Claude (Fable 5)

dawsontoth (approved):

dawsontoth:
ran locally, didn't notice any regression, seems ok -- captcha didn't show up for me, but i assume it assessed my login as low risk, and didn't trigger the display, which is a positive result.

DavidCockerill:
Thanks for running it — the no-regression signal on the surrounding flows is genuinely useful.

One thing worth flagging so you don't credit it with more than it did: locally you were almost certainly not exercising reCAPTCHA at all. `VITE_RECAPTCHA_SITE_KEY` is unset in `.env.local.example`, and with no key the loader is tree-shaken out of the bundle entirely — no script, no badge, no token on the wire. Local central-manager is also structurally exempt from enforcement, so even a token wouldn't have been checked. So "no regression" holds, but nothing captcha-shaped ran.

Also — and this is your own thread from the other comment — **login isn't gated**. Only signup and forgot-password are, so a sign-in would never have been assessed regardless.

On "captcha didn't show up": it never will. Score-based reCAPTCHA Enterprise is fully invisible — there's no checkbox and no image challenge to present, ever. The only visible artifact is Google's badge bottom-right, and that only appears once a site key is configured. (Scoring runs 0.0 bot → 1.0 human, so a good visitor scores *high*.)

If you want to actually drive it, point a local studio at stage, which is enforcing right now:

```
VITE_CENTRAL_MANAGER_API_URL=https://stage.studio.harperfabric.com
VITE_RECAPTCHA_SITE_KEY=6LckrIotAAAAAI4Kpk0kXiYtaAZKIBn2qhqabIfJ
```

The stage key allows `localhost`, so the badge appears and signup/forgot-password carry a real token that stage actually assesses. That's the rig used for the end-to-end verification in the description — signup 201 with a real token, and 403s for garbage or missing ones. — Claude (Fable 5)

DavidCockerill:
Correction to my previous comment: **stage is no longer enforcing** — David rolled the central-manager branch back off stage after I wrote that. So the `.env.local` rig I gave you would point at a stage that ignores tokens, and you'd see the badge appear but nothing actually assessed. Apologies for the bum steer.

There's no environment enforcing right now, so there isn't a self-serve way to drive the full loop until [central-manager#677](https://github.com/HarperFast/central-manager/pull/677) is deployed somewhere again. Setting the site key locally still shows you the client half — badge renders, `captchaToken` on the wire in the request body — which is most of what's worth eyeballing on this PR anyway.

The end-to-end results in the description were captured while stage *was* enforcing (signup 201 with a real token, forgot-password 200, 403 for garbage and for no token, plus the score branch firing when the threshold was raised), so that evidence stands — it just isn't reproducible on demand at the moment. — Claude (Fable 5)

dawsontoth:
** i did a sign-up as my test, i mispoke

DavidCockerill:
Good to know — though it lands the same either way: with no `VITE_RECAPTCHA_SITE_KEY` set locally, signup wouldn't have exercised the CAPTCHA any more than sign-in would. Your regression signal on the signup flow itself still stands, which is the useful part. — Claude (Fable 5)

dawsontoth:
I borrowed the one from stage, set it in my local .env.local, to exercise it. `VITE_RECAPTCHA_SITE_KEY`

DavidCockerill:
Ah — then you exercised more than I gave you credit for, and I was wrong twice over. Sorry for the noise.

With the key set you'd have had the full client half: badge bottom-right, token minted at submit, `captchaToken` in the request body. The only variable left is what `VITE_CENTRAL_MANAGER_API_URL` pointed at — a local central-manager is structurally exempt, so it would have accepted the token without assessing it, whereas stage was genuinely assessing until it got rolled back earlier today. Either way the client path you cared about is the part that was under review here, and you drove it. — Claude (Fable 5)

.env.local.example
kriszyp: This example still points `VITE_CENTRAL_MANAGER_API_URL` at stage, while leaving the site key unset. Once the companion service enforces CAPTCHA in stage, both local sign-up and password-reset requests will be tokenless and rejected. A separately created localhost key also cannot satisfy a central-manager configured for Harper’s stage key/project. Could we make the documented pairing usable—such as targeting the non-enforcing dev API, or providing a localhost-authorized key that stage central-manager accepts?

— KrAIs (Codex)

src/features/auth/ForgotPassword.tsx
kriszyp, line 60: The application’s `MutationCache.onError` still runs for this mutation because `useForgotPasswordMutation` has not opted out of the global toast. A CAPTCHA 403 therefore shows the default 403 toast and then this inline alert. The new test uses a `QueryClient` without the application mutation handler, so it misses that duplication. Please either suppress the global handler for this mutation and explicitly call the existing `errorHandler` for non-CAPTCHA failures, or add selective suppression; then exercise the real mutation handler in the test.

— KrAIs (Codex)

src/features/auth/isCaptchaVerificationError.ts
kriszyp: `firstString` returns the first string field rather than the first field containing the marker. For example, `{ error: 'Forbidden', message: 'Verification failed' }` is classified as non-CAPTCHA; an RFC 9457 body with `title: 'Forbidden'` and `detail: 'Verification failed'` is also missed. That silently drops the ad-blocker guidance and support path. The strongest contract would be a stable machine-readable central-manager error code; if text compatibility is still needed, please inspect every supported string (including `detail`) and add mixed-envelope coverage.

— KrAIs (Codex)

kriszyp (approved):
Sounds like a good security improvement.

src/features/auth/isCaptchaVerificationError.ts
DavidCockerill: Good catch — this was a real bug, fixed in 110db301. `firstString` returned the first *string* field rather than the first field naming the cause, so `{ error: 'Forbidden', message: 'Verification failed' }` classified as non-CAPTCHA and silently dropped both the ad-blocker guidance and the support path. It now checks every candidate field and includes RFC 9457's `detail` alongside `title`. Added a mixed-envelope test, and verified it fails against the pre-fix code so it's a real regression guard rather than a passing no-op.

Agreed that a stable machine-readable error code from central-manager is the stronger contract — it's already ledger entry 3 in the description as cheap-now/sticky-later, and this makes the interim matching less brittle without pretending it's the root fix. — Claude (Fable 5)

src/features/auth/ForgotPassword.tsx
DavidCockerill: Fixed in 110db301, taking your first option. `useForgotPasswordMutation` now sets `meta: { skipGlobalErrorToast: true }`, and the form re-raises non-CAPTCHA failures through `errorHandler` explicitly — so a CAPTCHA 403 is inline-only while every other failure keeps exactly the toast it has always had. That's better than the alternative I'd been weighing (converting forgot-password fully to inline like SignUp), since it leaves #1612's error UX untouched.

Your point about the test was the sharper half: the `QueryClient` had no `mutationCache`, so it structurally *could not* have caught the duplication. It now wires the app's real `mutationErrorHandler`, matching `SignUp.test.tsx`, with one test asserting a CAPTCHA 403 never reaches `toast.error` and another asserting a 500 still does. Both fail on the pre-fix code. — Claude (Fable 5)

.env.local.example
DavidCockerill: You've identified something I had documented too vaguely — the mismatch point especially. central-manager sends its own configured `RECAPTCHA_SITE_KEY` in every assessment, so a key a developer creates themselves is rejected as a mismatch *even if it authorizes localhost*. That's a non-obvious failure that would read as "the gate is broken" rather than "wrong key".

Rewrote the guidance in `.env.local.example` to say exactly that: unset is correct against local or dev central-manager (neither enforces), and exercising an enforcing backend requires that environment's own site key — the stage key already authorizes localhost for this purpose, copied from `.env.stage`.

Worth noting the premise has also shifted since you reviewed: David rolled the central-manager branch back off stage, so nothing is enforcing at the moment. — Claude (Fable 5)

PR #1637, opened 2026-08-19
Replace Swagger UI with a custom API explorer

### LINKED ISSUE #1530: [RUM] New handled TypeError on API Docs — swagger-ui-react cookie apiKey persist (reading 'schema')
## Summary

RUM is capturing a **new** handled JS error on the instance **API Docs (Swagger UI)** page, first seen in the last 24h and not present in the prior 24h or the preceding 30 days:

```
Error persisting cookie based apiKey in document.cookie.
TypeError: Cannot read properties of undefined (reading 'schema')
```

The error is thrown from inside the vendored `swagger-ui-react` bundle during the **Authorize** flow (`authorizeWithPersistOption` → persist apiKey to `document.cookie`), reading `.schema` on an `undefined` security-scheme object.

## Datadog RUM findings (last 24h)

- **Error message:** `Error persisting cookie based apiKey in document.cookie. TypeError: Cannot read properties of undefined (reading 'schema')`
- **Count:** 4 occurrences
- **Impact:** 1 session / 1 user
- **View:** instance API Docs page (`/$organizationId/$clusterId/$instanceId/apis` equivalent — `.../instance/ins-…`)
- **`error.source`:** `console` · **`error.handling`:** `handled` (Swagger catches it and `console.error`s; the page does not crash)
- **History:** 0 matches in the prior 24h window and none in the preceding 30 days — genuinely new this window.

### Stack trace (minified vendor bundle)

```
TypeError: Cannot read properties of undefined (reading 'schema')
  at <anonymous> @ swagger-ui-react-*.js:220:32962
  at Object.n @ swagger-ui-react-*.js:220:22614
  at Object.authorize @ swagger-ui-react-*.js:220:25792
  at <anonymous> @ swagger-ui-react-*.js:220:26161
  at Object.authorizeWithPersistOption @ swagger-ui-react-*.js:...
```

## Likely cause

`APIDocs.tsx` renders `<SwaggerUI persistAuthorization={true} withCredentials={true} … />` ([src/features/instance/apis/APIDocs.tsx:123](../blob/stage/src/features/instance/apis/APIDocs.tsx#L123)). With `persistAuthorization` enabled, when a user authorizes against a security scheme that Swagger treats as a **cookie-based apiKey**, `swagger-ui-react`'s persist path reads the scheme definition's `.schema` and throws if that definition is `undefined`/malformed for the served OpenAPI spec. Package is `swagger-ui-react ^5.29.0`.

## Suggested next steps

- Reproduce on an instance whose OpenAPI spec includes an `apiKey` security scheme with `in: cookie`, then click **Authorize**.
- Options to evaluate:
  - Bump/patch `swagger-ui-react` if a fixed version exists upstream.
  - Provide a custom auth-persist plugin (we already pass `plugins`) that guards the undefined security definition.
  - Confirm the OpenAPI spec Harper serves declares cookie-based apiKey schemes correctly.

## Severity

Low — handled, non-fatal, 1 user affected — but it is new and pollutes RUM, and indicates the Authorize/persist flow is broken for cookie-based apiKey schemes.

---
*Filed automatically by the daily Datadog RUM review. No customer-identifying information included; org/cluster/instance IDs are opaque generated identifiers and have been omitted. Data window: last 24h ending 2026-07-17.*

### LINKED ISSUE #1580: Long queries from Swagger UI
*Canopy* - Unable to execute queries against Tables in Fabric API tab. The `GET /status/` works but any `GET /Table/` returns this error. CORS is enabled, request appears to be using `https`, and network error seems unlikely with status check working. Even when no params are set, it generates a long query string that causes a `431 Request headers to large for server` when testing exact URL in Postman, so I think this is the error just masked in UI.
When I test on other clusters, I don't see the crazy long query string being generated with the request, so everything works. The main difference is Canopy schemas have a lot of relationships whereas the other clusters that tested working have none. It seems like we may need to change the behavior in Fabric UI for generating request URLs with relational queries to limit length.

dawsontoth:
Replaces the `swagger-ui-react` embed on the instance/cluster **APIs** tab with a custom, in-house
API explorer built entirely from the Studio design system. It reads the same runtime spec
(`GET /api/openapi/rest`) and preserves the CORS warning + one-click **Enable CORS** flow verbatim.

**What's new**

- **Hierarchical sidebar** — resource → path → method, collapsible and filterable. Harper's specs are
  untagged, so grouping is by path/resource rather than the useless single "default" tag.
- **Per-operation docs** — parameters, request/response schemas via a recursive model renderer.
- **Try it out** — a credentialed request runner with a JSON body editor (Monaco), path/query/header
  inputs, a live request preview, a real response view, and a copy-able `fetch` snippet.
- **Authorize** — a sidebar item that takes over the detail pane with **Server** + **Authorization**
  settings. Auth offers **Cookie** (session, default), **Basic**, and **Bearer**.
- **Server selector** — Studio's computed REST URL plus the spec's declared servers, so try-it-out
  works even when REST isn't on the port Studio guesses.
- **Persistence** — server + auth selections are saved to `localStorage` **per entity**, and cleared
  on sign-out.

Pure spec-parsing and request-building logic is extracted into unit-tested modules, with a mounted
component test for the explorer. Removes `swagger-ui-react` + `@types/swagger-ui-react` and their
~1,290-line lockfile subtree.

Closes #1530 — the RUM `TypeError` came from inside the vendored `swagger-ui-react` Authorize/persist
bundle; deleting the dependency removes it. Our Authorize never writes `document.cookie` or reads
`.schema` off a security scheme.

Closes #1580 — Swagger auto-populated a huge query string for `GET /Table/` on relationship-heavy
schemas (→ `431`). Our explorer starts query params empty and omits blanks, so no query string is
sent unless the user types one.

## For the human reviewer

Start in [`spec.ts`](src/features/instance/apis/explorer/spec.ts) and
[`request.ts`](src/features/instance/apis/explorer/request.ts) — the pure, unit-tested core; the
components render them. [`APIDocs.tsx`](src/features/instance/apis/APIDocs.tsx) keeps the CORS flow.

The judgment calls, most-consequential first:

- **Credentials in `localStorage`.** Per request, the server + Basic password / Bearer token persist
  per entity, **cleared on sign-out** (`authStore.signOutLocally`; full sign-out already clears all of
  `localStorage`), and written via a fresh read-merge-write so a concurrent tab's sign-out isn't
  clobbered. Tradeoff: creds at rest between sessions (XSS/shared-machine exposure) vs. convenience
  and parity with the old Swagger `persistAuthorization`. Alternative: persist only server + auth
  *type*, keep secrets in memory. Chose persist-and-clear-on-signout. (The default **Cookie** mode
  stores no secret at all — it just uses the session cookie.)
- **Default server is Studio's computed URL, not the spec's.** Cloud-safe (preserves prior behavior).
  On a split-port local instance the computed URL guesses port 80 and isn't reachable, so the
  reachable spec-declared server is one click away in the dropdown. The maintainer confirmed the
  cloud path runs the same code (only response JSON differs); happy to flip the default to the spec
  server if preferred.
- **Response cap = 512 KB, applied while streaming.** `readCappedBody` reads `response.body` only up
  to the cap and cancels the reader, so a multi-MB collection is never fully buffered or rendered.
- **Resource grouping is by first path segment, and `$ref` resolution doesn't JSON-Pointer-unescape.**
  Correct for Harper's generated specs (single-segment resources, simple schema names); noted as
  limitations for arbitrary OpenAPI producers.

Comments were audited twice against Harper's zero-new-comments default; the remainder are deliberate
contracts/rationale (the credentialed-fetch model, the ancestor-scoped cycle detection, the
per-entity persistence invariant). Flag any specific one you'd still cut.

## Verification

- **Live smoke** (`dev:local` against a local Harper instance): `GET /leaderboard/` → **200 OK**;
  `POST /game/{id}` path-param substitution + Send gating; server selector switching to the reachable
  `:9926`; Authorize Cookie/Basic/Bearer; selections persisting across reload; sign-out clearing the
  stored settings. The maintainer confirmed the cloud/cluster path runs the same code — only the
  spec/response JSON differs — so the local smoke is representative.
- **Full gate** on the pinned Node 24.19.0 — `vitest run` (311 files, 2541 passed), `tsc -b`,
  `oxlint .`, `dprint check`: all green. New tests cover spec parsing (incl. the example node budget),
  request building, streaming response cap, per-entity persistence (incl. the cross-tab stale-writer
  sequence), status colors, and a mounted-component test.

## Review coverage

Four cross-model pre-push rounds via the standard CLI (`--author claude`), each independent:

- **Round 1** (`7f12099`): **codex** (graded) + **gemini**. Fixed: Unicode Basic-auth `btoa` crash,
  cross-instance try-it-out state leak (per-entity remount), `generateExample` sibling-`$ref`
  truncation, response render cap, path-template param derivation, filter allocation.
- **Round 2** (`7209c74`, `--full`): **gemini**. Added sign-out credential clearing; lowered the
  response cap.
- **Round 3** (`c22edd8`): **gemini** + **cursor-grok**. Fixed: `authStore` `in`-on-non-object
  crash-safety guard, `mergeParameters` `$ref`-param collision guard, memoized example generation.
- **Round 4** (`1ff48aac`, convergence): **gemini** — no new actionable findings; only repeats of
  decided items, so iteration stopped here.

Not run: **Codex** was unavailable from round 2 on (a ChatGPT-workspace spend cap, not a code
issue); the **domain adjudicator** failed locally every round (a known zero-byte-log issue on this
machine), so outside findings were **hand-triaged** rather than auto-adjudicated; **Cursor** was
available only in round 3. Dismissed with rationale, not silence: a claimed `buildFetchSnippet`
template-literal injection (traced — every `${` is escaped, so no interpolation survives a paste), an
"undefined body" crash (`generateExample` has no `undefined` return path; `body` is always a string),
and a "cookies sent to external servers" leak (browsers scope cookies to the destination origin).
Deferred as not-producible-by-Harper-specs: multi-segment resource grouping, JSON-Pointer `$ref`
unescaping, `$ref` parameter resolution, and the snippet's `.json()` on 204 responses.

Post-open, the `gemini-code-assist` PR bot raised three prototype-safety points; applied in
`eb7ab94d` (`Object.hasOwn` guards in `forgetApiExplorerSettings`, `resolveRef`, and
`fillPathTemplate`) and the threads resolved.

@kriszyp (Codex-assisted) then reviewed and caught five real issues, all fixed in `c74932cc`:
(1) the response cap only bounded rendering — now `readCappedBody` streams `response.body` up to the
cap and cancels the reader, so a multi-MB collection is never fully buffered; (2) a cross-tab
stale-writer could resurrect a signed-out entity's credentials — persistence now does a fresh
read-merge-write per entity (`settings.ts`) plus a `storage`-event refresh; (3) `generateExample`
was bounded against cycles but not acyclic 2^N fan-out — added a depth + node budget; (4) Send now
gates on required query/header/body inputs, not just path params; (5) the sidebar used `max-height`
so its list never scrolled — now a definite height. His cookie question is answered on the thread:
the default **Cookie** mode already authenticates via the session cookie and stores no secret in
localStorage.

<sub>Human-Review-Need: 4 @ c74932ccd9f6</sub>


src/features/auth/store/authStore.ts
```diff
@@ -10,11 +10,28 @@ import {
...
+ * password or Bearer token per entity in localStorage; sign-out must not leave it for a later
+ * connection to reuse. (Full sign-out clears all of localStorage; this covers per-entity sign-out.)
+ */
+function forgetApiExplorerSettings(id: EntityIds): void {
+	const settings = getLocalStorage<Record<string, unknown>>(LocalStorageKeys.ApiExplorerSettings, {});
+	// A corrupted value (a primitive, or an array) must not throw from `in` and abort sign-out.
+	if (settings && typeof settings === 'object' && !Array.isArray(settings) && id in settings) {
+		delete settings[id];
+		setLocalStorage(LocalStorageKeys.ApiExplorerSettings, settings);
+	}
```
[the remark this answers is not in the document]
dawsontoth: Good call — applied in eb7ab94d using `Object.hasOwn` (the repo already uses it and targets ES2022), so an entity id can't match an inherited prototype key like `toString`.

src/features/instance/apis/explorer/spec.ts
```diff
@@ -0,0 +1,329 @@
...
+ * `undefined` for external or unrecognized refs — the renderer shows the raw ref name in that case
+ * rather than following it.
+ */
+export function resolveRef(spec: OpenApiSpec | undefined, ref: string | undefined): JsonSchema | undefined {
+	if (!spec || !ref || !ref.startsWith(REF_PREFIX)) {
+		return undefined;
+	}
+	const name = ref.slice(REF_PREFIX.length);
+	return spec.components?.schemas?.[name];
+}
```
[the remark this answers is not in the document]
dawsontoth: Applied in eb7ab94d — `resolveRef` now guards with `Object.hasOwn` before indexing `schemas[name]`, so a $ref name like `toString` resolves to undefined rather than a prototype method.

src/features/instance/apis/explorer/spec.ts
```diff
@@ -0,0 +1,329 @@
...
+	path: string,
+	values: Record<string, string>,
+): { path: string; missing: string[] } {
+	const missing: string[] = [];
+	const filled = path.replace(/\{([^}]+)\}/g, (_match, name: string) => {
+		const value = values[name];
+		if (value === undefined || value === '') {
+			missing.push(name);
+			return `{${name}}`;
+		}
```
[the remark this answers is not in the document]
dawsontoth: Applied in eb7ab94d — `fillPathTemplate` now treats a non-string `values[name]` as missing (`typeof value !== 'string'`), so a placeholder named `toString` can't pick up a prototype method.

src/features/instance/apis/explorer/request.ts
```diff
@@ -0,0 +1,230 @@
...
+): Promise<RequestResult> {
+	const start = now();
+	try {
+		const response = await fetch(request.url, {
+			method: request.method.toUpperCase(),
+			headers: request.headers,
+			body: request.body,
+			credentials: 'include',
+		});
+		const text = await response.text();
```
kriszyp: `MAX_DISPLAY_BODY_CHARS` currently limits only rendering: `response.text()` has already buffered and decoded the entire payload. A large collection or export can therefore freeze or exhaust the tab before `formatResponseBody` truncates it. Please read `response.body` incrementally up to a byte/character budget and cancel the reader once the preview is full; a streaming test should verify that the unread tail is not consumed.

src/features/instance/apis/explorer/ApiExplorer.tsx
```diff
@@ -0,0 +1,140 @@
...
+) {
+	const allOperations = useMemo(() => flattenOperations(spec), [spec]);
+	const serverOptions = useMemo(() => buildServerOptions(spec, baseURL), [spec, baseURL]);
+	const [filter, setFilter] = useState('');
+	const [view, setView] = useState<'operation' | 'settings'>('operation');
+	const [selectedId, setSelectedId] = useState<string | undefined>(() => allOperations[0]?.id);
+
+	// Server + auth selections persist per entity, so credentials for one instance never apply to
+	// another and each instance remembers its own server. Ephemeral navigation state (filter,
+	// selected endpoint, which pane is open) is deliberately not persisted.
```
kriszyp: This hook retains the complete settings map, while `useLocalStorage` does not synchronize direct writes or `storage` events. For example, tab A can retain settings for entities A and B, tab B can sign out B and delete B from localStorage, then an edit to A here writes tab A's stale map and restores B's password/token. That breaks the stated sign-out guarantee. Prefer keeping credentials in memory and persisting only server/auth type; if credential persistence is required, use a storage-backed external store synchronized across tabs and have sign-out update that same state owner. Please cover the two-tab/stale-writer sequence.

src/features/instance/apis/explorer/spec.ts
```diff
@@ -0,0 +1,333 @@
...
+		return node.default;
+	}
+	if (node.const !== undefined) {
+		return node.const;
+	}
+	if (Array.isArray(node.enum) && node.enum.length > 0) {
+		return node.enum[0];
+	}
+
+	const childSeen = new Set(seen).add(node);
```
kriszyp, line 231: Cycle detection does not bound acyclic expansion. Because each sibling receives the same ancestor-only `seen` set, a valid schema where every level has two references to the next level produces roughly 2^N example nodes; `JSON.stringify(generateExample(...))` then runs synchronously during rendering. Relationship-heavy Harper schemas can make this a practical tab freeze. Please add a maximum depth and/or total-node budget, returning a compact placeholder after the budget is exhausted, and test a branching shared-reference schema.

src/features/instance/apis/explorer/TryItOut.tsx
```diff
@@ -0,0 +1,237 @@
...
+	const inputs: RequestInputs = { pathParams, queryParams, headerParams, body };
+	const request = useMemo(
+		() => buildRequest(op, baseURL ?? '', inputs, auth),
+		// The built request depends on the operation, base URL, auth, and every input map/body.
+		// eslint-disable-next-line react-hooks/exhaustive-deps
+		[op, baseURL, auth, pathParams, queryParams, headerParams, body],
+	);
+
+	const hasBody = methodHasBody(op.method) && !!op.operation.requestBody;
+	const canSend = !!baseURL && request.missingPathParams.length === 0 && !isSending;
```
kriszyp: `canSend` checks only the base URL and missing path parameters. Harper request contracts can mark query parameters required, and OpenAPI can also require headers or the request body, so this currently enables Send for a request known to be incomplete and produces an avoidable server error. Please include required query/header/body inputs in the missing-input calculation and show them alongside the existing path warning.

src/features/instance/apis/explorer/ApiExplorer.tsx
```diff
@@ -0,0 +1,140 @@
...
+			</header>
+
+			{allOperations.length === 0
+				? (
+					<div className="border-border text-muted-foreground rounded-md border border-dashed p-8 text-center text-sm">
+						This spec doesn&apos;t define any endpoints yet.
+					</div>
+				)
+				: (
+					<div className="flex flex-col gap-6 lg:flex-row lg:items-start">
```
kriszyp, line 89: `max-height` alone does not establish a definite containing-block height, so `EndpointList`'s `h-full` resolves to auto and its inner `overflow-y-auto` area never becomes scrollable. With a normal multi-table spec, the endpoint tree spills beyond this intended viewport cap. Please give the wrapper a definite responsive height (or place the overflow on a definitively constrained flex child) and add a large-list layout check.

kriszyp (commented):
This sounds great. But can users use cookies to authenticate, without have to hand over credentials that get stored in localStorage?

src/features/instance/apis/explorer/request.ts
dawsontoth: Fixed in c74932cc — `readCappedBody` reads `response.body` only up to the 512 KB cap and cancels the reader, so the tail is never downloaded/buffered. Added a streaming test asserting the reader is cancelled and an endless stream is not drained.

src/features/instance/apis/explorer/ApiExplorer.tsx
dawsontoth: Fixed in c74932cc — persistence now does a fresh read-merge-write per entity (`settings.ts`) instead of writing a stale in-memory map, plus a `storage`-event refresh, so a concurrent sign-out is not clobbered. Kept credential persistence (per the requested behavior) rather than switching to in-memory-only. Added a test for the two-tab/stale-writer sequence.

src/features/instance/apis/explorer/spec.ts
dawsontoth: Fixed in c74932cc — `generateExample` now bounds output with a depth (8) and total-node (1000) budget, returning a compact `null` placeholder past either. Added a test with a 24-level branching shared-reference schema asserting the node count stays bounded.

src/features/instance/apis/explorer/TryItOut.tsx
dawsontoth: Fixed in c74932cc — Send now gates on required query/header/body inputs too (not just path params), and lists them alongside the existing path warning.

src/features/instance/apis/explorer/ApiExplorer.tsx
dawsontoth: Fixed in c74932cc — the wrapper now has a definite height (not `max-height`) so `EndpointList`s `h-full` resolves and the tree scrolls internally. Verified in-browser: bounded ~335px container with `overflow-y:auto` and `scrollHeight > clientHeight`.

dawsontoth:
Great question — yes. The default **Cookie** auth mode uses the browser's instance session cookie (`credentials: "include"`) and stores **no** secret in localStorage; only `{type:"cookie"}` is persisted. Basic/Bearer are opt-in for testing as a specific identity, and they are the only modes that persist a secret (now cleared on sign-out). So a user who just wants cookie auth never hands over stored credentials.

src/features/instance/apis/explorer/settings.ts
```diff
@@ -0,0 +1,33 @@
...
+	return Object.hasOwn(map, entityId) ? map[entityId] ?? {} : {};
+}
+
+/**
+ * Persist one entity's settings with a fresh read-merge-write against localStorage — never a stale
+ * in-memory copy of the whole map. Another tab may have signed an entity out (deleting its key) since
+ * this tab loaded; writing back a stale full map would resurrect that entity's credentials and break
+ * the sign-out guarantee. Merging into a fresh read touches only this entity.
+ */
+export function writeEntitySettings(entityId: string, patch: Partial<ExplorerEntitySettings>): void {
```
kriszyp: A fresh read removes the long-lived stale snapshot, but this read–modify–write is still not atomic across tabs. Tab A can read the map while editing entity A, tab B can then sign entity B out and write the deletion, and tab A can finally write its previously read map and restore B's password/token. Please use independently keyed per-entity storage (so edit and deletion each touch only one atomic key), or coordinate every writer—including sign-out—with the same cross-tab locking mechanism. Please cover this interleaving rather than only the sequence where sign-out finishes before the edit begins.

src/features/instance/apis/explorer/TryItOut.tsx
```diff
@@ -0,0 +1,245 @@
...
+
+	const inputs: RequestInputs = { pathParams, queryParams, headerParams, body };
+	const request = useMemo(
+		() => buildRequest(op, baseURL ?? '', inputs, auth),
+		// The built request depends on the operation, base URL, auth, and every input map/body.
+		// eslint-disable-next-line react-hooks/exhaustive-deps
+		[op, baseURL, auth, pathParams, queryParams, headerParams, body],
+	);
+
+	const hasBody = methodHasBody(op.method) && !!op.operation.requestBody;
```
kriszyp: Body support is derived from `methodHasBody`, not from the operation contract. An OpenAPI DELETE operation with `requestBody.required: true` therefore gets no editor, is not listed in `missingRequired`, and can be sent without its required body. Please derive body handling from `op.operation.requestBody` for methods fetch permits to carry a body, and explicitly handle unsupported GET/HEAD bodies. A DELETE-body component test would guard the complete path.

src/features/instance/apis/explorer/request.ts
```diff
@@ -0,0 +1,261 @@
...
+
+	const headers: Record<string, string> = { Accept: 'application/json' };
+	for (const [name, value] of Object.entries(inputs.headerParams)) {
+		if (value !== '') {
+			headers[name] = value;
+		}
+	}
+	Object.assign(headers, authHeader(auth));
+
+	const hasBody = methodHasBody(op.method) && inputs.body.trim() !== '';
```
kriszyp: Every non-empty body is labeled `application/json`, although `jsonSchemaFromContent` deliberately falls back to non-JSON entries such as `text/plain` or `text/csv`. Those operations will display an editor but send the wrong representation and media type. Please carry the selected content entry's media type through the request inputs/build step—or explicitly restrict try-it-out to JSON instead of silently mis-sending other contracts.

src/features/instance/apis/explorer/request.ts
```diff
@@ -0,0 +1,261 @@
...
+): Promise<{ text: string; truncated: boolean }> {
+	const reader = response.body?.getReader?.();
+	if (!reader) {
+		const text = await response.text();
+		return text.length > maxChars ? { text: text.slice(0, maxChars), truncated: true } : { text, truncated: false };
+	}
+	const decoder = new TextDecoder();
+	let text = '';
+	try {
+		while (text.length < maxChars) {
```
kriszyp: When decoded text reaches `maxChars` exactly, the loop exits without observing whether the stream is done; the reader is then cancelled, but `truncated` is returned as false. A response whose first chunk is exactly the cap and has a remaining tail is consequently cut off without the truncation notice. Please track completion or perform a bounded lookahead before cancelling, and add an exact-cap-first-chunk-plus-tail test.

DavidCockerill (approved):
Approving at this head. The security decision I went looking for is declared in the description with the rejected alternative named, so rather than restate it I checked whether the mitigation actually holds — and it holds more broadly than the description claims.

**On credentials in `localStorage`:**

- `signOutLocally(id)` calls `forgetApiExplorerSettings(id)`, so signing out of an entity drops its stored server and credentials.
- `signOutAllLocally()` iterates every entity through it, and its docblock is explicit that it exists for *"a session already known-dead (e.g. a 401 from CM)"*. So **session expiry reaches the clear**, not just an explicit sign-out click. That was the gap I expected to find — people rarely click sign-out — and it is covered.
- The read-merge-write in `writeEntitySettings` is the subtle one, and the comment names the real hazard: another tab may have signed an entity out since this tab loaded, so writing back a stale full map would **resurrect that entity's credentials** and quietly break the sign-out guarantee. Persisting the in-memory map would have looked equivalent and wasn't.
- `cookie` is the default and adds no `Authorization` header, so the common path stores no secret at all.

**Residual, for the design record rather than as a blocker:** while signed in, an XSS anywhere in Studio can read a stored Basic password or Bearer token, and on an instance that credential is frequently `super_user`. The rejected alternative — persist the auth *type*, keep the secret in memory — costs one re-entry per reload. I would not hold this: the tradeoff is declared, @gemini-code-assist already pushed the sign-out clearing in round 2, and the default mode sidesteps it entirely. Worth landing somewhere more durable than a PR description that gets squashed away, though.

**Worth crediting:**

- **Grouping by path/resource instead of tag.** Harper's specs are untagged, so a tag-based sidebar renders one useless "default" group. Building for the structure the data actually has, and saying why, beats inheriting Swagger's model.
- **The server selector includes the spec's declared servers**, not only Studio's computed REST URL — so try-it-out still works when REST isn't on the port Studio guesses. That is the failure mode a home-grown explorer would most plausibly ship with.
- **Logic extracted from rendering** — `spec.ts`, `request.ts`, `settings.ts` with their own tests, so the parsing and request-building rules are verifiable without mounting a component.
- CORS warning and one-click Enable CORS preserved verbatim rather than reimplemented.

**Coverage, so the approve isn't overclaiming:** I read `settings.ts`, `request.ts`'s auth-header construction, the `authStore` sign-out wiring and the persistence model closely. I did not read closely `SchemaView.tsx`'s recursive model renderer, `TryItOut.tsx`, `OperationDetail.tsx`, `EndpointList.tsx`, the Monaco integration, or the `vite.config.ts` / lockfile changes. If there is a second finding it is most likely in the recursive schema renderer — cyclic `$ref` handling is the classic failure there, and the ancestor-scoped cycle detection mentioned in the description is the part I did not verify.

PR #1639, opened 2026-08-20
fix(apis): independent + resizable API explorer sidebar

dawsontoth:
Follow-up to #1637: polishes the API explorer's endpoint sidebar layout. Two changes:

## 1. Independent sidebar/detail scrolling (the original fix)

On large lists the sidebar double-scrolled — the page grew its own scrollbar alongside the tree's —
and never reached the viewport bottom. The explorer now uses the **independent-scroll layout the
Applications tab already uses**. On `lg`, it's a fixed-height app-shell — `calc(100vh - (--spacing(32)))`,
where `spacing(32)` is the layout's `mt-32` offset (the same token the databases/log sidebars use). The
header takes its natural space and the two-pane row fills the rest, so the sidebar reaches the viewport
bottom exactly, the tree and detail each scroll on their own, and the page itself doesn't scroll — robust
to the CORS banner / a wrapped `info.description` (no per-element offset math). Below `lg` it stacks and
the page scrolls as before, with a `max-h` cap on the sidebar.

## 2. Drag-to-resize sidebar

The sidebar can be dragged wider/narrower, matching the **Applications and Databases tabs**. A thin handle
on its right edge (hover / drag / focus) resizes it; the separator is keyboard-focusable and Arrow keys
nudge it in 10px steps. Width is clamped to `[240px, half-viewport]`, persisted to `localStorage`, and
applies only at `lg+`. It mirrors the databases resize hook (delta-based, since this sidebar sits in-flow).
The width is a global UI preference, kept separate from the per-entity server/auth settings since it carries
no credential.

## For the human reviewer

Judgment calls, most-consequential first:

- **Detail pane now has its own scrollbar (like Apps), and the shell reaches the viewport bottom.** This is
  the requested behavior ("extend to the bottom accurately" + "scroll independently like Apps/Databases").
  Consequence flagged by two reviewers (codex, cursor-grok): the global `NotificationBanner` is
  `fixed bottom-0`, so when an un-acked system notice is showing it overlays the bottom of the panes, and
  the flush-to-bottom shell has no dead space beneath it the way stage's `pb-12` did. I kept the
  flush-to-bottom because it's what was asked for and because the **databases sidebar (the prior art) is
  also `h-[calc(100vh-(--spacing(32)))]` and sits under the same banner identically** — reserving banner
  space is an app-wide concern, not this PR's. Happy to add bottom clearance if you'd rather.
- **Third copy of the resize hook.** `useResizableSidebar` is a near-duplicate of
  `useResizableDatabasesSidebar`. Deliberate — the repo's convention is one copy per feature (applications
  and databases each keep their own). Three reviewers flagged the duplication; filed as **#1640** (P3/Task)
  to extract a shared primitive and fold in the pointer-event/blur and per-mousemove-rerender improvements
  there rather than diverge here.
- **`clampWidth` guards non-finite input.** A corrupted/legacy `localStorage` value (safeParse returns any
  valid JSON) would otherwise render `--api-sidebar-width: NaNpx` and collapse the sidebar to 0; it now
  falls back to the default. Verified against the served bundle: `clampWidth({…}, 1280) === 320`.

## Verification

Live in-browser (`lg`, large list forced), `getComputedStyle` / scroll metrics:

- **Scrolling:** sidebar bottom flush with the viewport (`gapBelowAside: 0`); tree and detail each
  `overflow-y: auto` and scroll independently; `pageOverflowPx: 0`.
- **Resize:** a real `left_click_drag` on the handle took the sidebar 320px → 440px, clamped at the 240px
  minimum past it, committed `440` to `localStorage` on release, and restored `440` after a full reload;
  `aria-valuenow` tracked the width; no console errors.
- **NaN guard / layout hardening:** the served module returns 320 for object/string/NaN input (240 for
  too-small, 440 unchanged for valid); the aside computes `overflow-y: clip` and the detail pane
  `overflow-y: auto` + `overscroll-behavior-y: contain`.

Full gate on Node 24.19.0 — `vitest run` (321 files, 2621 passed), `tsc -b`, `oxlint .`, `dprint check`:
green. Layout itself is browser-verified above (jsdom has no layout engine); the hook's
clamp/delta/persist/keyboard/NaN math and the separator ARIA + CSS-var wiring are unit-tested.

## Review coverage

Authored by Claude (Opus). Two cross-model pre-push rounds via the standard CLI (`--author claude`),
codex as the graded opposite-family leg:

- **Round 1** (`a1b0b8be`, full): **codex + gemini + cursor-grok**, `independent=true`. Surfaced the
  malformed-`localStorage` NaN collapse, missing `overscroll-contain` on the detail pane, and the aside
  `overflow-y-clip` gap — all **fixed** here — plus the duplication (→ #1640) and the banner interaction
  (judgment call above).
- **Round 2** (`d65d3e4d`, full): **codex** (`independent=true`; gemini flaked and the Cursor legs hit the
  branch round-cap). Verdict COMMENTS, no blockers — codex confirmed the three fixes landed
  ("fixed by finite clamping, `overscroll-contain`, and `overflow-y-clip`"). Remaining items are the
  prior-art matches tracked in #1640, the documented banner tradeoff, and a comment nit (trimmed).

The **domain adjudicator failed locally both rounds** (a known zero-byte-log issue on this machine), so
outside findings were hand-triaged rather than auto-adjudicated. `Human-Review-Need: 4` is the
deterministic-policy floor from that degraded coverage (failed gemini/domain legs + `localStorageKeys.ts`
tagged high-risk for a one-line enum add), not an open blocker. HEAD `dd63208b` is `d65d3e4d` plus a
trivial comment-only trim (no re-review needed per the tiering rule).

<sub>Complexity: moderate</sub>

<sub>Human-Review-Need: 4 @ d65d3e4d651a</sub>

### DIFF
--- a/src/features/instance/apis/explorer/ApiExplorer.tsx
+++ b/src/features/instance/apis/explorer/ApiExplorer.tsx
@@ -70,8 +70,11 @@ export function ApiExplorer(
 	}
 
 	return (
-		<div className="flex flex-col gap-5">
-			<header className="flex min-w-0 flex-col">
+		// On lg this is a fixed-height app-shell (its parent sets the height): the header takes its
+		// natural space and the two-pane row fills the rest, so the sidebar reaches the viewport bottom
+		// exactly and the panes scroll internally instead of the page — no page-plus-tree double scroll.
+		<div className="flex flex-col gap-5 lg:min-h-0 lg:flex-1">
+			<header className="flex min-w-0 flex-col lg:shrink-0">
 				<div className="flex items-center gap-2">
 					<h1 className="font-radioGrotesk truncate text-2xl">{spec?.info?.title ?? 'API Explorer'}</h1>
 					{spec?.info?.version && <Badge variant="secondary">v{spec.info.version}</Badge>}
@@ -86,30 +89,24 @@ export function ApiExplorer(
 					</div>
 				)
 				: (
-					<div className="flex flex-col gap-6 lg:flex-row lg:items-start">
-						<aside className="lg:sticky lg:top-4 w-full shrink-0 lg:w-80">
-							{
-								/* Definite height (not max-height) so EndpointList's `h-full` resolves and its inner
-							    tree actually scrolls instead of spilling past the cap on a large spec. */
-							}
-							<div className="h-[28rem] lg:h-[calc(100vh-11rem)]">
-								<EndpointList
-									tree={filteredTree}
-									totalCount={allOperations.length}
-									filteredCount={filteredOperations.length}
-									isFiltering={filter.trim() !== ''}
-									selectedId={view === 'operation' ? selectedOp?.id : undefined}
-									onSelect={selectOperation}
-									filter={filter}
-									onFilterChange={setFilter}
-									authType={auth.type}
-									settingsActive={view === 'settings'}
-									onOpenSettings={() => setView('settings')}
-								/>
-							</div>
+					<div className="flex flex-col gap-6 lg:min-h-0 lg:flex-1 lg:flex-row lg:items-stretch">
+						<aside className="w-full shrink-0 lg:h-full lg:min-h-0 lg:w-80">
+							<EndpointList
+								tree={filteredTree}
+								totalCount={allOperations.length}
+								filteredCount={filteredOperations.length}
+								isFiltering={filter.trim() !== ''}
+								selectedId={view === 'operation' ? selectedOp?.id : undefined}
+								onSelect={selectOperation}
+								filter={filter}
+								onFilterChange={setFilter}
+								authType={auth.type}
+								settingsActive={view === 'settings'}
+								onOpenSettings={() => setView('settings')}
+							/>
 						</aside>
 
-						<main className="min-w-0 flex-1">
+						<main className="min-w-0 flex-1 lg:min-h-0 lg:overflow-y-auto lg:pb-6">
 							{view === 'settings'
 								? (
 									<SettingsPanel
--- a/src/features/instance/apis/explorer/ApiExplorer.tsx
+++ b/src/features/instance/apis/explorer/ApiExplorer.tsx
@@ -11,8 +11,14 @@ import {
 	operationMatchesFilter,
 } from '@/features/instance/apis/explorer/spec';
 import { OpenApiSpec } from '@/features/instance/apis/explorer/types';
+import {
+	maxSidebarWidth,
+	MIN_SIDEBAR_WIDTH,
+	useResizableSidebar,
+} from '@/features/instance/apis/explorer/useResizableSidebar';
 import { useCopyToClipboard } from '@/hooks/useCopyToClipboard';
-import { useEffect, useMemo, useState } from 'react';
+import { cn } from '@/lib/cn';
+import { type CSSProperties, useEffect, useMemo, useState } from 'react';
 
 /**
  * The custom Harper API explorer: a hierarchical, searchable list of the spec's operations alongside
@@ -56,6 +62,10 @@ export function ApiExplorer(
 		?? baseURL;
 	const [copyServer] = useCopyToClipboard(activeServer ?? '');
 
+	// Width drives a CSS variable applied only at lg+; below that the sidebar stacks full-width.
+	const { width: sidebarWidth, isResizing, startResizing, handleKeyDown } = useResizableSidebar();
+	const sidebarWidthVar = { '--api-sidebar-width': `${sidebarWidth}px` } as CSSProperties;
+
 	const filteredOperations = useMemo(
 		() => allOperations.filter(op => operationMatchesFilter(op, filter)),
 		[allOperations, filter],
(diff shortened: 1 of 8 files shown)


src/features/instance/apis/explorer/EndpointList.tsx
```diff
@@ -60,13 +60,17 @@ export function EndpointList({
 	const usingHeaderAuth = authType === 'basic' || authType === 'bearer';
 
 	return (
-		<div className="flex h-full flex-col gap-3">
+		// max-height (not a fixed height) so short lists shrink and don't force a page scrollbar
+		// alongside the tree's own — the offset uses the same `--spacing(32)` layout token as the
+		// databases/log sidebars, plus an allowance for this page's header. `flex-1 min-h-0` on the
+		// tree is what lets it scroll within the cap without a definite parent height.
+		<div className="flex max-h-[28rem] flex-col gap-3 lg:max-h-[calc(100vh-(--spacing(32))-8rem)]">
```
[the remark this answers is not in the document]
dawsontoth: Verified this is valid Tailwind v4, so keeping the current syntax. `--spacing(32)` is the framework's spacing **function** (it compiles to `calc(var(--spacing) * 32)` = 8rem), which the shipped Databases and Log sidebars already use. In the browser the calc resolves correctly — the container computes to exactly `innerHeight − 128px`.

The suggested `(--spacing-32)` is the parenthesis shorthand for `var(--spacing-32)`, but that variable isn't defined by the default theme (only the base `--spacing: 0.25rem` is), so it would resolve to nothing and break the `calc()`.

Also note this exact line has since been refactored into the app-shell height on `APIDocs.tsx`.

src/features/instance/apis/explorer/useResizableSidebar.ts
```diff
@@ -0,0 +1,120 @@
...
+	useEffect(() => {
+		if (!isResizing) {
+			return;
+		}
+		const onMouseMove = (event: MouseEvent) => {
+			const gesture = gestureRef.current;
+			if (!gesture) {
+				return;
+			}
+			setWidth(clampWidth(gesture.startWidth + (event.clientX - gesture.startX), window.innerWidth));
```
kriszyp: `setWidth` is owned by `ApiExplorer` at `src/features/instance/apis/explorer/ApiExplorer.tsx:66`, so every native `mousemove` rerenders the entire endpoint tree and operation detail. For a large spec—or the Try-it-out view containing Monaco—the resize gesture itself becomes a render hotspot even though localStorage writes are deferred. Please keep transient movement off that full render path, such as by isolating the resizing shell and throttling updates with `requestAnimationFrame`, or imperatively updating the CSS variable and committing React/storage state on release.

src/features/instance/apis/explorer/useResizableSidebar.ts
```diff
@@ -0,0 +1,120 @@
...
+		};
+		const onMouseUp = () => {
+			setIsResizing(false);
+			// Persist the final width once, instead of on every mousemove.
+			setPersistedWidth(widthRef.current);
+		};
+		// Suppress text selection across the page while dragging.
+		const previousUserSelect = document.body.style.userSelect;
+		document.body.style.userSelect = 'none';
+		window.addEventListener('mousemove', onMouseMove);
```
kriszyp: This gesture only finishes on `window.mouseup`. If the user drags outside the browser and releases there, no mouseup is guaranteed; `isResizing` stays true, `document.body.style.userSelect` remains `none`, and the width is not committed until some later mouseup. Touch/pen cancellation is also unsupported, and any mouse button can currently arm resizing. Please use Pointer Events with a primary-button guard and pointer capture, then finish on `pointerup`, `pointercancel`, or `lostpointercapture` (with `blur` as a fallback), and cover cancellation in the hook test.

src/features/instance/apis/explorer/ApiExplorer.tsx
```diff
@@ -86,30 +99,55 @@ export function ApiExplorer(
...
-									settingsActive={view === 'settings'}
-									onOpenSettings={() => setView('settings')}
+							<div
+								role="separator"
+								tabIndex={0}
+								aria-orientation="vertical"
+								aria-label="Resize sidebar"
+								aria-valuenow={sidebarWidth}
+								aria-valuemin={MIN_SIDEBAR_WIDTH}
+								aria-valuemax={maxSidebarWidth(window.innerWidth)}
```
kriszyp: This value is read directly from `window.innerWidth` during render, but the resize listener only calls `setWidth`. When the saved width remains in range—for example, resizing from 1200px to 1400px with a 320px sidebar—React can bail out on `setWidth(320)`, leaving `aria-valuemax` at 600 even though keyboard resizing now permits 700. Please make the maximum reactive in the hook (or otherwise update the ARIA value on every viewport resize) and add a no-clamp viewport-resize assertion.

DavidCockerill (approved):
Approving at this head. No findings.

**The NaN guard is the part worth calling out**, because it is a real failure mode with a silent symptom rather than defensive noise. `safeParse` returns any valid JSON, so a corrupted or legacy value under that key could be an object, array or string; the arithmetic would then render `--api-sidebar-width: NaNpx` and collapse the sidebar to zero width with no error anywhere. Falling back to the default is right, and the comment explains the mechanism rather than just asserting a guard.

**And the persistence decision is reasoned against #1637's rather than copied from it.** Width is a global UI preference instead of per-entity, with the reason stated — *"unlike the server/auth settings it carries no credential, so it's fine to share across instances."* Holding that distinction deliberately matters given #1637 stores credentials under a sibling key; it would have been easy to reach for the same per-entity shape out of consistency and quietly widen what is scoped per instance.

Also right: committing to localStorage on release rather than per `mousemove`, keyboard resize on a focusable separator so the control is not mouse-only, and reusing the databases sidebar's gesture model while explaining the one way this differs (in-flow with page padding to its left, so it tracks the delta from drag start rather than raw cursor X, unlike the applications tray anchored at the viewport edge).

One thing to know rather than change: `calc(100vh - (--spacing(32)))` couples the layout to the `mt-32` offset, which the description acknowledges. It is the same token the databases and log sidebars use, so the coupling is at least consistent — just worth remembering it is three places to change rather than one if that offset ever moves.

PR #1641, opened 2026-08-20
feat(apis-explorer): session-scoped credentials and a real log-in flow

dawsontoth:
Builds out the APIs Explorer "Authorize" story so authorization is credible and safe in deployed
environments, where the default cookie flow silently fails.

## What changed

- **Session-scoped credentials.** Per-entity Authorize state moves from `localStorage` to
  `sessionStorage` (cleared on tab close), with a one-time scrub of the legacy plaintext
  `localStorage` secrets on first load. `authStore` sign-out now delegates to a shared
  `forgetEntitySettings`.
- **A real log-in that works cross-site.** Cookie auth isn't sent when the instance is on a
  different site than Studio. The Authorize panel now mints a Bearer **operation token** — one click
  for your current Studio session, or a username/password fallback — sent as an explicit
  `Authorization` header, which crosses origins where cookies don't.
- **Docs + Try it out.** The Authorize panel is now a Documentation / Try it out pair (mirroring
  operations): docs explain Log in / Basic / Bearer / Cookie; Try it out is where you authenticate,
  which flips the lock and unlocks authenticated requests. Default method is Log in.
- **Deep-links.** Auth-required operations link to the Try-it-out log-in view; `requiresAuth` now
  honors OpenAPI optional-auth (`security: [{}]`) and explicit `[]` overrides.

## For the human reviewer

Read `src/features/instance/apis/APIDocs.tsx` and
`src/integrations/api/instance/auth/createInstanceAuthenticationTokens.ts` first — the credential
boundary is the thing to scrutinize.

Decisions a reviewer might reasonably question:

- **decision — credential-mint direct-URL source.** The username/password fallback POSTs
  `create_authentication_tokens` to the **operations client's own `baseURL`** — the address Studio
  already uses to talk to this instance — and only when it passes `isDirectOperationsUrl` (rejects
  the Fabric Connect `/HDBInstance/…`/`/Cluster/…` proxy paths). The check is enforced *inside* the
  mint helper, not just at the call site, so the boundary can't regress. When the only reachable URL
  is the proxy, the password fallback is withheld (fail-closed) and one-click session mint still
  works. Session mint intentionally uses the existing (possibly proxied) client because it authorizes
  as *you* and discloses no one else's credentials.
- **decision — token stored, password not (for Log in).** The log-in flow stores only the minted
  short-lived token in sessionStorage, never the password. Basic auth still stores username+password
  (session-scoped); the docs say so per method. `sessionStorage` is not claimed as an XSS boundary —
  it's a persistence/lifetime choice.
- **decision — code sample shows the live credential.** `buildFetchSnippet` embeds the real
  `Authorization` header in the copyable sample (as the prior Swagger UI did). Kept deliberately: it's
  the user's own credential in their own tab, and redacting would break the copy-paste-run purpose.
  Flagging in case exfil-via-clipboard is in scope for this surface.
- **verified — entity sign-out clears explorer creds.** `signOutOfInstance` → `signOutLocally` →
  `forgetEntitySettings` clears per-entity; full logout (`clearAuthStateLocally`, `logoutOnSuccess`)
  calls `clearSessionStorage()`. So a review note that "normal disconnect paths bypass cleanup" did
  not hold on trace.
- **hardening — the credential POST fails on redirect** (`redirect: 'error'`) so a 3xx can't replay
  it past the direct-URL check, and the log-in form drops the typed password from memory once a token
  is minted.

## Verification

- Full local gate on Node 24.19.0: `tsc -b`, `oxlint`, `dprint check` clean; `vitest run` — all
  2630 tests pass (added coverage for storage/scrub, `isDirectOperationsUrl`, `requiresAuth`, the
  mint helper incl. non-direct refusal + server-error surfacing, and the explorer's login flows,
  method-switch mint race, and deep-link). The suite's process exit is non-zero only from a
  pre-existing, unrelated undici-WebSocket unhandled error in a Chat test (reproduces without this
  diff).
- Dev server smoke: the explorer builds and loads against the stage central manager.
- Cross-origin: the cross-site Bearer flow was verified locally — a minted token is accepted on a
  cross-origin request (the instance's CORS allows the `Authorization` header on preflight), which is
  the behavior the cookie path couldn't provide and the core reason for this change.

## Process

Planning review (step 6) cleared `chosen-approach-sound` after widening the option set. Step-10
cross-model review ran five rounds (codex + gemini + cursor-grok, independent). The Harper-domain
adjudication leg fails on this host (zero-byte log), so outside findings were hand-adjudicated.
Round 1 (BLOCK) → round 2 (CHANGES; blocker + majors fixed) → round 3 found **no blockers or majors** →
round 4 (delta) covers a gemini bot finding — the credential direct-URL gate was case-sensitive, so
`isDirectOperationsUrl` now rejects proxy paths case-insensitively → round 5 is the Final-artifact
check after rebasing onto `stage` to integrate its API-explorer sidebar-resize feature, and found
**no blockers or majors** in the merge. The `Human-Review-Need: 4` floor reflects the high-risk
auth/storage surface plus the failed adjudication leg (degraded coverage), not an open blocker.

Review response (kriszyp): all 8 inline findings are fixed and the threads resolved — per-entity
cleanup centralized in `setUserForIdAndKey` (closing a cross-user credential leak via the
ClusterHome/ClusterCard disconnect paths), an auth-epoch that invalidates in-flight mints and
propagates sign-out across tabs (including an always-on listener for unmounted tabs), credentials
stamped to the server they were authorized against, form drafts resynced on Clear, the legacy scrub
moved to bootstrap, and the "Authorized" indicator relabelled "Credential set".

Follow-up rounds fixed three further majors the reviews surfaced in my own hardening: minted tokens
are now stamped with Studio's computed URL for the entity (not the spec picker's selection, which had
recreated a narrower version of the leak the stamping exists to prevent);
`signOutFromPotentiallyAuthenticatedInstances` — the one sign-out path that flags sign-out directly
rather than via `signOutLocally` — also clears explorer state; and a mirrored cross-tab invalidation
advances this tab's in-memory epoch so an in-flight mint is caught by the epoch check too. Each has a
regression test.

A second review round from kriszyp found two more real gaps, both fixed in `a439aad7`: an event-only
cross-tab signal couldn't survive a reload (sessionStorage outlives one, so a restarted tab never saw
the invalidation and kept a revoked credential) — the sign-out generation is now durable and every
credential is stamped with the generation it was created under, reconciled at bootstrap; and the mint's
abort timer didn't cover body consumption, so a stalled body left the log-in pending forever.

**Coverage gap, stated plainly:** the last two commits (`e832fca6`, `a439aad7`) have **no outside-model
review**. The run for the first produced nothing — Gemini hit an account quota and Codex hung and was
killed with no budget to retry — and I did not re-run for the second. That is an absent review, not a
clean one. What they do have: the full local gate (2661 unit tests, `tsc`, `oxlint`, `dprint`) and a
regression test per fix — including one that caught a bug in my own first attempt at the abort fix
(a blanket `.catch` was swallowing the abort). Everything up to `254c81ee` was reviewed by
codex+gemini. Reviewers may reasonably want to look hardest at those two commits.

Known follow-up (not in this PR): the credential indicator reports that *a* credential is configured,
not that it satisfies the operation's specific security requirement (AND/OR alternatives against
referenced schemes). Happy to fold that in if wanted.

Complexity: moderate — new auth UI surface plus a narrow, well-bounded reach into the operations
client for token minting.

<sub>Review-Coverage: authored=unknown; ran=none; rounds=1 @ a439aad722dd</sub>

<sub>Human-Review-Need: 4 @ a439aad722dd</sub>

### DIFF
--- a/src/features/auth/store/authStore.ts
+++ b/src/features/auth/store/authStore.ts
@@ -2,8 +2,8 @@ import { isLocalStudio, localStudioDevUrl } from '@/config/constants';
 import { getInstanceClient } from '@/config/getInstanceClient';
 import { getCurrentUser } from '@/features/auth/queries/getCurrentUser';
 import {
-	forgetAllEntitySettings,
 	forgetEntitySettings,
+	pruneStaleEntitySettings,
 	scrubLegacySettings,
 } from '@/features/instance/apis/explorer/settings';
 import { SchemaCluster, SchemaHdbInstance } from '@/integrations/api/api.gen';
@@ -76,36 +76,47 @@ class AuthStore {
 	private readonly fabricConnectInFlight = new Map<EntityIds, Promise<LocalUser>>();
 	private readonly operationTokenRefreshInFlight = new Map<EntityIds, Promise<string | null>>();
 
-	// Per-entity count of sign-out events. The API explorer keeps its credential per browser tab
-	// (sessionStorage), so a sign-out in one tab can't reach another's storage directly: this epoch is
-	// bumped on every sign-out and mirrored to localStorage, letting the explorer (a) reject an
-	// in-flight token mint that resolves after a sign-out and (b) clear its own tab's stored credential
-	// when another tab signs the entity out.
-	private readonly explorerAuthEpoch = new Map<EntityIds, number>();
-	private explorerEpochSeq = Date.now();
+	// Sign-out generations for the API explorer, per entity plus a global `'*'` slot. The explorer keeps
+	// its credential per browser tab (sessionStorage), which SURVIVES a reload — so an event-only signal
+	// is not enough: a tab that reloads after another tab signed out would never see the event. The
+	// generation therefore lives in localStorage (durable, shared) and each stored credential records the
+	// generation it was created under, so a stale credential is detected by comparison at read time
+	// regardless of whether this tab was running when the sign-out happened.
 	private readonly explorerAuthEpochKey = 'Studio:ExplorerAuthEpoch';
 
 	constructor() {
 		this.potentiallyAuthenticated = JSON.parse(localStorage.getItem(this.potentiallyAuthenticatedKey) || '{}');
 	}
 
+	private readExplorerGenerations(): Record<string, number> {
+		try {
+			const raw = JSON.parse(localStorage.getItem(this.explorerAuthEpochKey) || '{}') as unknown;
+			return raw && typeof raw === 'object' && !Array.isArray(raw) ? raw as Record<string, number> : {};
+		} catch {
+			return {};
+		}
+	}
+
 	/**
-	 * Current sign-out epoch for an entity — its own count plus the global (`'*'`) count, so a global
-	 * logout advances every entity's epoch even in the same tab (where no `storage` event fires). The
-	 * explorer stamps a mint with it and re-checks before applying.
+	 * Current sign-out generation for an entity — its own count plus the global (`'*'`) count, so a
+	 * global logout advances every entity's generation. The explorer stamps a credential with this and
+	 * compares before using it (and before applying an in-flight mint).
 	 */
 	public getExplorerAuthEpoch(id: EntityIds): number {
-		return (this.explorerAuthEpoch.get(id) ?? 0) + (this.explorerAuthEpoch.get('*') ?? 0);
+		const generations = this.readExplorerGenerations();
+		const own = typeof generations[id] === 'number' ? generations[id] : 0;
+		const all = typeof generations['*'] === 'number' ? generations['*'] : 0;
+		return own + all;
 	}
 
 	private writeExplorerInvalidation(id: EntityIds | '*'): void {
-		// Increment the raw slot (not getExplorerAuthEpoch, which folds in the '*' count).
-		this.explorerAuthEpoch.set(id, (this.explorerAuthEpoch.get(id) ?? 0) + 1);
 		try {
-			// Value must differ each write so other tabs' `storage` listeners fire.
-			localStorage.setItem(this.explorerAuthEpochKey, JSON.stringify({ id, seq: ++this.explorerEpochSeq }));
+			const generations = this.readExplorerGenerations();
+			const current = typeof generations[id] === 'number' ? generations[id] : 0;
+			// The value changes on every write, so other tabs' `storage` listeners still fire.
+			localStorage.setItem(this.explorerAuthEpochKey, JSON.stringify({ ...generations, [id]: current + 1 }));
 		} catch {
-			// Storage disabled — the in-memory epoch still guards the same-tab in-flight-mint race.
+			// Storage disabled: the explorer's per-tab clearing on sign-out still applies within this tab.
 		}
 	}
 
@@ -119,39 +130,13 @@ class AuthStore {
 	}
 
 	/**
-	 * The entity a mirrored invalidation names (`'*'` = all), or null if the event isn't one. Advances
-	 * this tab's in-memory epoch for that entity as a side effect, so an in-flight mint here is
-	 * invalidated by another tab's sign-out too — the epoch stays authoritative in every tab. Each
-	 * installed listener advances it, which is fine: only a *change* in the value is meaningful.
+	 * Subscribe to a change in the explorer sign-out generation (any entity, or a global logout). The
+	 * caller re-compares its own entity's stamped generation rather than trusting the event to name one,
+	 * which is the same check that runs at bootstrap — so a missed event can't leave a stale credential.
 	 */
-	private parseExplorerInvalidation(event: StorageEvent): string | null {
-		if (event.key !== this.explorerAuthEpochKey || !event.newValue) {
-			return null;
-		}
-		let id: string | null = null;
-		try {
-			const parsed = JSON.parse(event.newValue) as { id?: string };
-			id = typeof parsed.id === 'string' ? parsed.id : null;
-		} catch {
-			return null;
-		}
-		if (id !== null) {
-			this.explorerAuthEpoch.set(id, (this.explorerAuthEpoch.get(id) ?? 0) + 1);
-		}
-		return id;
-	}
-
-	/**
-	 * Subscribe to cross-tab sign-out of one entity (or a global `'*'` logout). Fires when another tab
-	 * signs `id` out via the mirrored epoch key; the caller resets its live auth state. The browser only
-	 * delivers `storage` events to other documents, so the signing-out tab is unaffected (it clears
-	 * itself directly). This is component-lifetime; a tab whose explorer is unmounted relies on the
-	 * always-on `installExplorerCrossTabCleanup` listener to scrub its stored credential.
-	 */
-	public onExplorerAuthInvalidated(id: EntityIds, callback: () => void): () => void {
+	public onExplorerAuthInvalidated(_id: EntityIds, callback: () => void): () => void {
 		const handler = (event: StorageEvent) => {
-			const invalidated = this.parseExplorerInvalidation(event);
-			if (invalidated === id || invalidated === '*') {
+			if (event.key === this.explorerAuthEpochKey) {
 				callback();
 			}
 		};
@@ -160,20 +145,21 @@ class AuthStore {
 	}
 
 	/**
-	 * App-level (always-on) listener that clears the per-tab explorer credential for an entity another
-	 * tab signed out — even when this tab's explorer is unmounted, so a tab that navigated away can't
-	 * later restore a revoked token. Also re-scrubs the legacy localStorage secret if a pre-upgrade tab
-	 * rewrote it. Installed once at bootstrap.
+	 * App-level cleanup, installed once at bootstrap. Runs immediately — reconciling credentials stored
+	 * in this tab's sessionStorage (which survives a reload) against the durable generation, so a
+	 * sign-out that happened while this tab was closed or reloading is still honored — and again whenever
+	 * the generation changes, so an unmounted explorer's credential is dropped too.
 	 */
 	public installExplorerCrossTabCleanup(): () => void {
-		const handler = (event: StorageEvent) => {
-			const invalidated = this.parseExplorerInvalidation(event);
-			if (invalidated === '*') {
-				forgetAllEntitySettings();
-			} else if (invalidated) {
-				forgetEntitySettings(invalidated);
-			}
+		const reconcile = () => {
+			pruneStaleEntitySettings(entityId => this.getExplorerAuthEpoch(entityId));
 			scrubLegacySettings();
+		};
+		reconcile();
+		const handler = (event: StorageEvent) => {
+			if (event.key === this.explorerAuthEpochKey) {
+				reconcile();
+			}
 		};
 		window.addEventListener('storage', handler);
 		return () => window.removeEventListener('storage', handler);
--- a/src/features/auth/store/authStore.ts
+++ b/src/features/auth/store/authStore.ts
@@ -72,8 +72,56 @@ class AuthStore {
 	private readonly fabricConnectInFlight = new Map<EntityIds, Promise<LocalUser>>();
 	private readonly operationTokenRefreshInFlight = new Map<EntityIds, Promise<string | null>>();
 
+	// Per-entity count of sign-out events. The API explorer keeps its credential per browser tab
+	// (sessionStorage), so a sign-out in one tab can't reach another's storage directly: this epoch is
+	// bumped on every sign-out and mirrored to localStorage, letting the explorer (a) reject an
+	// in-flight token mint that resolves after a sign-out and (b) clear its own tab's stored credential
+	// when another tab signs the entity out.
+	private readonly explorerAuthEpoch = new Map<EntityIds, number>();
+	private explorerEpochSeq = Date.now();
+	private readonly explorerAuthEpochKey = 'Studio:ExplorerAuthEpoch';
+
 	constructor() {
 		this.potentiallyAuthenticated = JSON.parse(localStorage.getItem(this.potentiallyAuthenticatedKey) || '{}');
+	}
+
+	/** Current per-entity sign-out epoch; the explorer stamps a mint with it and re-checks before applying. */
+	public getExplorerAuthEpoch(id: EntityIds): number {
+		return this.explorerAuthEpoch.get(id) ?? 0;
+	}
+
+	private bumpExplorerAuthEpoch(id: EntityIds): void {
+		this.explorerAuthEpoch.set(id, this.getExplorerAuthEpoch(id) + 1);
+		try {
+			// Value must differ each write so other tabs' `storage` listeners fire.
+			localStorage.setItem(this.explorerAuthEpochKey, JSON.stringify({ id, seq: ++this.explorerEpochSeq }));
+		} catch {
+			// Storage disabled — the in-memory epoch still guards the same-tab in-flight-mint race.
+		}
+	}
+
+	/**
+	 * Subscribe to cross-tab sign-out of one entity. Fires when another tab signs `id` out (via the
+	 * mirrored epoch key); the caller clears its own tab's stored explorer credential in response. The
+	 * browser only delivers `storage` events to other documents, so the tab that signed out is
+	 * unaffected (it clears itself directly).
+	 */
+	public onExplorerAuthInvalidated(id: EntityIds, callback: () => void): () => void {
+		const handler = (event: StorageEvent) => {
+			if (event.key !== this.explorerAuthEpochKey || !event.newValue) {
+				return;
+			}
+			try {
+				const parsed = JSON.parse(event.newValue) as { id?: string };
+				if (parsed.id === id) {
+					callback();
+				}
+			} catch {
+				// Ignore a malformed signal rather than throwing into the storage-event dispatch.
+			}
+		};
+		window.addEventListener('storage', handler);
+		return () => window.removeEventListener('storage', handler);
 	}
 
 	public getAllConnections(): Record<EntityIds, AuthenticatedConnection> {
@@ -153,6 +201,11 @@ class AuthStore {
 			this.flagKeyAsSignedIn(id, key);
 		} else {
 			this.flagKeyAsSignedOut(id);
+			// Every per-entity disconnect funnels through here (ClusterHome/ClusterCard call
+			// setUserForEntity(entity, null)), so clear the API explorer's stored credentials for the
+			// entity here too — otherwise the next user to sign into the same entity would inherit them.
+			forgetEntitySettings(id);
+			this.bumpExplorerAuthEpoch(id);
 		}
 		this.updateConnectionIfChanged(id, false, user);
 	}
@@ -416,6 +469,7 @@ class AuthStore {
 		this.flagKeyAsSignedOut(id);
 		this.updateConnectionIfChanged(id, false, null);
 		forgetEntitySettings(id);
+		this.bumpExplorerAuthEpoch(id);
 	}
 
 	/**
--- a/src/features/auth/store/authStore.ts
+++ b/src/features/auth/store/authStore.ts
@@ -118,17 +118,27 @@ class AuthStore {
 		this.writeExplorerInvalidation('*');
 	}
 
-	/** The entity a mirrored invalidation names (`'*'` = all), or null if the event isn't one. */
+	/**
+	 * The entity a mirrored invalidation names (`'*'` = all), or null if the event isn't one. Advances
+	 * this tab's in-memory epoch for that entity as a side effect, so an in-flight mint here is
+	 * invalidated by another tab's sign-out too — the epoch stays authoritative in every tab. Each
+	 * installed listener advances it, which is fine: only a *change* in the value is meaningful.
+	 */
 	private parseExplorerInvalidation(event: StorageEvent): string | null {
 		if (event.key !== this.explorerAuthEpochKey || !event.newValue) {
 			return null;
 		}
+		let id: string | null = null;
 		try {
 			const parsed = JSON.parse(event.newValue) as { id?: string };
-			return typeof parsed.id === 'string' ? parsed.id : null;
+			id = typeof parsed.id === 'string' ? parsed.id : null;
 		} catch {
 			return null;
 		}
+		if (id !== null) {
+			this.explorerAuthEpoch.set(id, (this.explorerAuthEpoch.get(id) ?? 0) + 1);
+		}
+		return id;
 	}
 
 	/**
@@ -489,6 +499,8 @@ class AuthStore {
 			this.updateConnectionIfChanged(entityId, false, null);
 			this.flagKeyAsSignedOut(entityId);
 			this.fabricConnectAuth.delete(entityId);
+			forgetEntitySettings(entityId);
+			this.bumpExplorerAuthEpoch(entityId);
 			if (entityId === OverallAppSignIn) {
 				continue;
 			}
--- a/src/features/auth/store/authStore.ts
+++ b/src/features/auth/store/authStore.ts
@@ -1,6 +1,7 @@
 import { isLocalStudio, localStudioDevUrl } from '@/config/constants';
 import { getInstanceClient } from '@/config/getInstanceClient';
 import { getCurrentUser } from '@/features/auth/queries/getCurrentUser';
+import { forgetEntitySettings } from '@/features/instance/apis/explorer/settings';
 import { SchemaCluster, SchemaHdbInstance } from '@/integrations/api/api.gen';
 import { Cluster, Instance, LocalUser, User } from '@/integrations/api/api.patch';
 import {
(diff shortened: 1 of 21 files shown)


src/lib/urls/isDirectOperationsUrl.ts
```diff
@@ -0,0 +1,6 @@
+// The Fabric Connect proxy routes operations through these central-manager paths (see
+// getInstanceClient). A URL carrying either segment points at the proxy, not the instance, so it must
+// never receive an instance Bearer JWT or typed credentials — those would reach the central manager.
+export function isDirectOperationsUrl(url: string | null | undefined): url is string {
+	return !!url && !url.includes('/HDBInstance/') && !url.includes('/Cluster/');
+}
```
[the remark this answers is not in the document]
dawsontoth: Good catch — hardened in a3aa3d2c: `isDirectOperationsUrl` now rejects the proxy segments case-insensitively (`/\/(hdbinstance|cluster)\//i`). This gate guards credentials, so failing closed on any casing is worth it.

src/lib/urls/isDirectOperationsUrl.test.ts
```diff
@@ -0,0 +1,20 @@
...
+describe('isDirectOperationsUrl', () => {
+	it('accepts a direct instance/cluster operations URL', () => {
+		expect(isDirectOperationsUrl('https://host.example:9925/')).toBe(true);
+		expect(isDirectOperationsUrl('http://localhost:9925')).toBe(true);
+	});
+
+	it('rejects Fabric Connect proxy paths', () => {
+		expect(isDirectOperationsUrl('https://cm.example/HDBInstance/ins-1/operation')).toBe(false);
+		expect(isDirectOperationsUrl('https://cm.example/Cluster/clu-1/operation')).toBe(false);
+	});
```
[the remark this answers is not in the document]
dawsontoth: Added the lowercase `/hdbinstance/` and `/cluster/` cases in a3aa3d2c.

src/integrations/api/instance/auth/createInstanceAuthenticationTokens.test.ts
```diff
@@ -0,0 +1,70 @@
...
+		vi.stubGlobal('fetch', fetchMock);
+		await expect(
+			mintOperationTokenWithCredentials({
+				operationsUrl: 'https://cm.example/HDBInstance/ins-1/operation',
+				username: 'admin',
+				password: 'pw',
+			}),
+		).rejects.toThrow('non-direct operations URL');
+		expect(fetchMock).not.toHaveBeenCalled();
+	});
```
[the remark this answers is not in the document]
dawsontoth: Added a lowercase proxy-URL rejection case to this test in a3aa3d2c.

src/features/instance/apis/explorer/spec.ts
```diff
@@ -16,6 +17,33 @@ export function resourceOf(path: string): string {
...
+/** A single security alternative that actually demands a credential: a non-empty requirement object. */
+function hasSecurityRequirement(requirement: unknown): boolean {
+	if (!requirement || typeof requirement !== 'object' || Array.isArray(requirement)) {
+		return false;
+	}
+	for (const _ in requirement) {
+		return true;
+	}
+	return false;
+}
```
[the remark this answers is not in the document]
dawsontoth: Keeping the `for...in` here deliberately: an earlier review round flagged `Object.keys(req).length` as allocating an array on the OperationDetail render path, so I switched to this allocation-free check. The input is an OpenAPI security requirement parsed from JSON (a plain object with no enumerable inherited props) and it's already guarded by the `typeof === 'object' && !Array.isArray` check above, so `for...in` only sees own keys. Leaving as-is to avoid reintroducing the per-render allocation.

dawsontoth:
Cross-origin verified: a locally-run test confirms a minted operation token is accepted on a cross-site request (the instance's CORS allows the `Authorization` header on preflight) — the behavior the cookie path couldn't provide and the core reason for this change. That was the last open verification gate, so moving this out of draft.

CI is green and all bot review threads are resolved. Reviewers: please look hardest at the credential boundary in `APIDocs.tsx` + `createInstanceAuthenticationTokens.ts` (direct-URL-only credential mint) and the two documented lifecycle trade-offs of session-scoped storage noted in the description.

_🤖 Generated with [Claude Code](https://claude.com/claude-code)_

dawsontoth:
Rebased onto `stage` to resolve merge conflicts. `stage` had landed the API-explorer sidebar resize + independent scrolling (ddd5819e, 7ada6690); the conflicts were in `ApiExplorer.tsx` (that new resizable-aside layout vs. this PR's auth rewrite) and `localStorageKeys.ts` (its new `ApiExplorerSidebarWidth` key vs. this PR removing `ApiExplorerSettings`). Both features are preserved: the resizable/scrolling sidebar now renders the auth-aware `EndpointList` (method + authorized props), and the enum keeps `ApiExplorerSidebarWidth` without `ApiExplorerSettings`.

Verified after rebase: `tsc -b`, `oxlint`, `dprint` clean and the full unit suite green (2647 passed). Worth a fresh look at the merged `ApiExplorer.tsx` layout.

_🤖 Generated with [Claude Code](https://claude.com/claude-code)_

src/features/instance/apis/APIDocs.tsx
```diff
@@ -31,6 +36,25 @@ export function APIDocs() {
...
+	// base URL — the address Studio already talks to this instance on. When that address is the Fabric
+	// Connect proxy path, it fails the direct check and the credential fallback is withheld so typed
+	// credentials never reach central manager; session mint (below) still works there.
+	const operationsBaseURL = operationsParams.instanceClient.defaults.baseURL;
+	const onSessionMint = useCallback(
+		async () =>
+			(await createInstanceAuthenticationTokens({ instanceClient: operationsParams.instanceClient })).operationToken,
+		[operationsParams.instanceClient],
+	);
+	const onCredentialMint = useMemo(
```
kriszyp: This mint callback is permanently bound to `operationsBaseURL`, while ApiExplorer stores the resulting token only by entity and sends it to whichever independently selected `activeServer` is current. With a multi-server spec, selecting another declared server—or changing servers while minting is pending—attaches this entity’s token to that other origin and may also use the wrong credential for the target. Please bind minted auth to a validated server target and clear/refuse it when the target changes; one safe policy is to restrict one-click minted tokens to the trusted computed Studio server.

src/features/instance/apis/explorer/ApiExplorer.tsx
```diff
@@ -18,43 +22,146 @@ import {
...
+	const [authorizeTab, setAuthorizeTab] = useState<'docs' | 'try'>('docs');
 	const [selectedId, setSelectedId] = useState<string | undefined>(() => allOperations[0]?.id);
 
-	// Server + auth selections persist per entity, so credentials for one instance never apply to
-	// another. Writes go straight to localStorage via a fresh read-merge-write (see settings.ts) so a
-	// concurrent sign-out in another tab isn't clobbered; local state mirrors it for rendering and
-	// refreshes on the cross-tab `storage` event. Ephemeral navigation state (filter, selected
-	// endpoint, which pane is open) is deliberately not persisted.
+	// Opening Authorize from the sidebar lands on Documentation; a deep-link from an auth-required
+	// operation lands on the actionable "Try it out" log-in view.
```
kriszyp: Moving the secret to sessionStorage also removes the previous cross-tab logout propagation: separate top-level tabs have separate sessionStorage areas, and the old `storage` listener is gone. If two tabs hold Basic/Bearer auth, signing out in one leaves the other tab’s credential and in-memory state usable; cookie logout does not neutralize stored Basic credentials, and operation JWTs may remain valid. Please keep secrets tab-local but broadcast entity/all-auth invalidations through `BroadcastChannel` or a non-secret localStorage epoch, then clear both storage and component state in every tab.

src/features/instance/apis/explorer/SettingsPanel.tsx
```diff
@@ -73,83 +97,244 @@ export function SettingsPanel({
...
+
+			{login.status === 'error' && login.error && <p className="text-red text-sm">{login.error}</p>}
+			{!authorized && login.status === 'idle' && (
+				<p className="text-muted-foreground text-xs">Not signed in — authenticated requests will be rejected.</p>
+			)}
 		</div>
 	);
 }
 
+function BasicForm({ auth, onApply }: { auth: ApiAuth; onApply: (username: string, password: string) => void }) {
```
kriszyp: `onClearAuth` empties the parent credential, but these Basic/Bearer forms initialize their local state only on first mount. Clicking Clear therefore leaves the password/token visible in the mounted input, and submitting immediately restores the supposedly cleared credential. Please synchronize local form state when `auth` is cleared or key/remount the form on credential changes; add coverage for both Basic and Bearer Clear flows.

src/features/instance/apis/explorer/ApiExplorer.tsx
```diff
@@ -18,43 +22,146 @@ import {
...
+	useEffect(() => {
+		mountedRef.current = true;
+		return () => {
+			mountedRef.current = false;
+			attemptRef.current++;
+		};
+	}, []);
+	const [loginStatus, setLoginStatus] = useState<'idle' | 'pending' | 'error'>('idle');
+	const [loginError, setLoginError] = useState<string | null>(null);
+
```
kriszyp: The attempt guard is invalidated by UI changes and unmount only, not by external logout/session invalidation. If `clearAuthStateLocally()` clears sessionStorage while this request is pending, a response arriving before router-driven unmount still passes these checks and line 92 writes the old user’s token back after logout. Please include a shared auth/logout generation or authenticated identity in the mint stamp, re-check it before committing, and test logout between mint start and resolution.

src/features/instance/apis/explorer/request.ts
```diff
@@ -11,6 +11,25 @@ export type ApiAuth =
...
+ */
+export type AuthMethod = 'login' | 'basic' | 'bearer' | 'cookie';
+
+/**
+ * Whether the auth carries an explicit credential the request will actually send — a Basic username
+ * or a Bearer token. `cookie` (and an empty Basic/Bearer) is not "authorized" in this sense: it only
+ * relies on an ambient session cookie that may not be sent cross-site. Branch-only, allocation-free:
+ * it runs on the sidebar's lock indicator during ordinary renders.
+ */
+export function isAuthorized(auth: ApiAuth): boolean {
```
kriszyp: This predicate means “an explicit credential is configured,” not that the selected operation is authorized. A Login-minted Bearer token marks a Basic-only operation Authorized; one half of an OpenAPI AND requirement also marks it Authorized; valid cookie/API-key auth is always shown as unauthorized. Please either label this state “Credential configured” or evaluate the operation’s complete security alternatives against their referenced schemes before claiming Authorized.

src/features/auth/store/authStore.ts
```diff
@@ -438,7 +415,7 @@ class AuthStore {
 		this.flagForFabricConnect(id, false);
 		this.flagKeyAsSignedOut(id);
 		this.updateConnectionIfChanged(id, false, null);
-		forgetApiExplorerSettings(id);
+		forgetEntitySettings(id);
```
kriszyp, line 514: Could we centralize every per-entity disconnect/user transition through this cleanup? Explorer credentials are keyed only by entity, but `src/features/cluster/ClusterHome.tsx:84`, `src/features/cluster/ClusterHome.tsx:308`, and `src/features/clusters/components/ClusterCard.tsx:117-120` clear users with `setUserForEntity(..., null)` instead of calling this method. User A can therefore authorize Explorer, disconnect, let user B sign into the same entity, and B’s Explorer will reload A’s stored Basic credentials or Bearer token. Please bind settings to the authenticated identity/auth epoch or make all disconnect and sign-in transitions clear Explorer state, with an A→logout→B regression test.

src/features/instance/apis/APIDocs.tsx
kriszyp, line 50: This mint callback is permanently bound to `operationsBaseURL`, while ApiExplorer stores the resulting token only by entity and sends it to whichever independently selected `activeServer` is current. With a multi-server spec, selecting another declared server—or changing servers while minting is pending—attaches this entity’s token to that other origin and may also use the wrong credential for the target. Please bind minted auth to a validated server target and clear/refuse it when the target changes; one safe policy is to restrict one-click minted tokens to the trusted computed Studio server.

src/features/instance/apis/explorer/ApiExplorer.tsx
```diff
@@ -18,43 +22,146 @@ import {
...
+	// operation lands on the actionable "Try it out" log-in view.
+	const openAuthorize = (targetTab: 'docs' | 'try') => {
+		setAuthorizeTab(targetTab);
+		setView('settings');
+	};
+
+	// Server + auth selections persist per entity in sessionStorage (this browser tab only), so
+	// credentials never cross entities and are cleared when the tab closes. Ephemeral navigation state
+	// (filter, selected endpoint, which pane is open) is deliberately not persisted.
 	const [entitySettings, setEntitySettings] = useState(() => readEntitySettings(entityId));
```
kriszyp, line 64: Moving the secret to sessionStorage also removes the previous cross-tab logout propagation: separate top-level tabs have separate sessionStorage areas, and the old `storage` listener is gone. If two tabs hold Basic/Bearer auth, signing out in one leaves the other tab’s credential and in-memory state usable; cookie logout does not neutralize stored Basic credentials, and operation JWTs may remain valid. Please keep secrets tab-local but broadcast entity/all-auth invalidations through `BroadcastChannel` or a non-secret localStorage epoch, then clear both storage and component state in every tab.

src/features/instance/apis/explorer/SettingsPanel.tsx
kriszyp, line 279: `onClearAuth` empties the parent credential, but these Basic/Bearer forms initialize their local state only on first mount. Clicking Clear therefore leaves the password/token visible in the mounted input, and submitting immediately restores the supposedly cleared credential. Please synchronize local form state when `auth` is cleared or key/remount the form on credential changes; add coverage for both Basic and Bearer Clear flows.

src/features/instance/apis/explorer/ApiExplorer.tsx
```diff
@@ -18,43 +22,146 @@ import {
...
+		mountedRef.current = true;
+		return () => {
+			mountedRef.current = false;
+			attemptRef.current++;
+		};
+	}, []);
+	const [loginStatus, setLoginStatus] = useState<'idle' | 'pending' | 'error'>('idle');
+	const [loginError, setLoginError] = useState<string | null>(null);
+
+	const runMint = async (mint: () => Promise<string>) => {
```
kriszyp, line 105: The attempt guard is invalidated by UI changes and unmount only, not by external logout/session invalidation. If `clearAuthStateLocally()` clears sessionStorage while this request is pending, a response arriving before router-driven unmount still passes these checks and line 92 writes the old user’s token back after logout. Please include a shared auth/logout generation or authenticated identity in the mint stamp, re-check it before committing, and test logout between mint start and resolution.

src/features/instance/apis/explorer/settings.ts
```diff
@@ -1,33 +1,132 @@
...
+
+/**
+ * One-time removal of the pre-sessionStorage `localStorage` map. That map persisted Basic passwords
+ * and Bearer tokens in plaintext across sessions; simply switching this module to sessionStorage would
+ * leave the old value readable indefinitely. We drop it wholesale (no secret is migrated) and always
+ * attempt removal even if reading threw. Bounded limitation: a concurrently-open pre-upgrade tab can
+ * write the key back — this runs on every explorer init, but not continuously.
+ */
+function scrubLegacySettings(): void {
+	if (legacyScrubbed) {
```
kriszyp: The legacy secret cleanup is both lazy and one-shot. It runs only when `readMap()` is called, so upgraded users who never reopen Explorer retain the old localStorage password/token indefinitely. Once it does run, `legacyScrubbed` prevents the “every explorer init” retry described above; a concurrently open pre-upgrade tab can repopulate the key until a full reload. Please run this migration during application bootstrap and retry removal on relevant storage events (or at least remove the one-shot guard), with coverage independent of `readEntitySettings()`.

src/features/instance/apis/explorer/request.ts
kriszyp, line 28: This predicate means “an explicit credential is configured,” not that the selected operation is authorized. A Login-minted Bearer token marks a Basic-only operation Authorized; one half of an OpenAPI AND requirement also marks it Authorized; valid cookie/API-key auth is always shown as unauthorized. Please either label this state “Credential configured” or evaluate the operation’s complete security alternatives against their referenced schemes before claiming Authorized.

src/features/instance/apis/explorer/SettingsPanel.tsx
```diff
@@ -73,83 +97,244 @@ export function SettingsPanel({
...
+	);
+}
+
+function LoginForm({ login, authorized }: { login: LoginController; authorized: boolean }) {
+	const [username, setUsername] = useState('');
+	const [password, setPassword] = useState('');
+	const pending = login.status === 'pending';
+
+	// Don't keep the typed credentials in memory once a token has been minted from them.
+	useEffect(() => {
```
kriszyp, line 216: This cleanup only runs when the boolean `authorized` changes. If a login token is already loaded, submitting different username/password credentials replaces it with another non-empty token, so `authorized` stays true and the submitted password remains in the mounted form. Please clear the credential fields on successful credential mint rather than relying on the generic authorization boolean, and add a regression test that starts already authorized before submitting alternate credentials.

kriszyp (commented):
I think this all seems like a good approach. There some suggestions for further hardening, but I think this generally correct.

src/features/auth/store/authStore.ts
dawsontoth: Fixed in 3268990b: `setUserForIdAndKey` now calls `forgetEntitySettings(id)` on every sign-out, so the ClusterHome/ClusterCard `setUserForEntity(entity, null)` paths clear explorer credentials too. Regression test added (settings.test.ts: 'a sign-out through setUserForIdAndKey drops the stored explorer credential') covering the A → disconnect → B case.

src/features/instance/apis/APIDocs.tsx
dawsontoth: Fixed in 254c81ee, and I took your "bind minted auth to a validated server target" option: the credential is now stamped with the server it was authorized against (`authServer`) and is only sent when the active server still matches — so a token never follows a picker change *or* an implicit `activeServer` recompute. Cleared on logout/method reset.

src/features/instance/apis/explorer/ApiExplorer.tsx
dawsontoth: Fixed across 3268990b + a9ef962c using the localStorage-epoch approach you suggested (not BroadcastChannel, to stay consistent with the store's existing localStorage signalling): `authStore` bumps a per-entity epoch on sign-out and mirrors it to `Studio:ExplorerAuthEpoch`. Two listeners consume it — the mounted explorer resets live state, and an always-on `installExplorerCrossTabCleanup` (wired in main.tsx) scrubs the stored credential even when the explorer is unmounted. `signOutAllLocally` broadcasts a global '*' clear.

src/features/instance/apis/explorer/SettingsPanel.tsx
dawsontoth: Fixed in 3268990b: both forms now resync from the applied `auth` via an effect, so Clear empties the inputs and there's nothing to resubmit. Tests added for the Basic Clear flow and the pasted-Bearer path.

src/features/instance/apis/explorer/ApiExplorer.tsx
dawsontoth: Fixed in 3268990b, extended in 254c81ee: `runMint` captures `authStore.getExplorerAuthEpoch(entityId)` and refuses to apply the token if it changed. `getExplorerAuthEpoch` folds in the global '*' slot so a same-tab `clearAuthStateLocally()` invalidates in-flight mints too (no storage event fires in the originating tab). Test: 'discards a session mint that resolves after the entity was signed out'.

src/features/instance/apis/explorer/settings.ts
dawsontoth: Fixed in 3268990b + a9ef962c: the scrub is now an exported `scrubLegacySettings()` called at app bootstrap in main.tsx (so users who never open the explorer are scrubbed), the one-shot latch is gone, and the always-on cross-tab listener re-scrubs on relevant storage events. It no longer runs on every settings read. Tests are independent of `readEntitySettings`.

src/features/instance/apis/explorer/request.ts
dawsontoth: Relabelled rather than deepened the check (3268990b): the indicator now reads "Credential set" in the panel and on the operation badge, since as you note it only means a credential is configured — not that it satisfies this operation's security requirement. Evaluating full AND/OR alternatives against referenced schemes is a reasonable follow-up; happy to file it if you'd like it in this PR.

src/features/instance/apis/explorer/SettingsPanel.tsx
dawsontoth: Fixed in 3268990b: clearing is now keyed on the mint status transition (pending → idle) rather than the `authorized` boolean, so submitting different credentials while already authorized still clears the fields. Regression test starts already-authorized and then submits alternate credentials.

dawsontoth:
@kriszyp thanks — this was a genuinely valuable review; the first finding was a real cross-user credential leak. All 8 inline findings are fixed and their threads resolved (individual replies on each). Pushed as `3268990b` → `a9ef962c` → `254c81ee`:

- **Cross-user leak (authStore:418).** `setUserForIdAndKey` now clears explorer settings on every sign-out, so the `ClusterHome`/`ClusterCard` `setUserForEntity(entity, null)` disconnect paths are covered. A→disconnect→B regression test added.
- **Token bound to the wrong server (APIDocs:50).** Took your "bind minted auth to a validated server target" option: credentials are stamped with the server they were authorized against and only sent when the active server still matches — covering implicit `activeServer` recomputes, not just picker changes.
- **Cross-tab logout (ApiExplorer:59).** Per-entity auth epoch mirrored through localStorage: the mounted explorer resets live state, and an always-on bootstrap listener scrubs the stored credential even when the explorer is unmounted. `signOutAllLocally` broadcasts a global clear.
- **Mint after logout (ApiExplorer:84).** Mints are stamped with the epoch and refuse to apply if it moved; the global slot folds into the per-entity value so a same-tab full logout also invalidates (no storage event fires in the originating tab).
- **Form drafts / re-auth (SettingsPanel:275, :213).** Forms resync from the applied credential on Clear; the login password clears on the mint status transition, so re-authenticating while already authorized clears it too.
- **Legacy scrub (settings.ts:75).** Now an exported `scrubLegacySettings()` run at app bootstrap, re-runnable, and off the settings read path.
- **"Authorized" (request.ts:28).** Relabelled **"Credential set"** — as you said, it only means a credential is configured. Evaluating full AND/OR alternatives against referenced schemes is noted as a follow-up; say the word and I'll fold it in here.

Also hardened while in here: the proxy-path gate now tests the parsed pathname (rejects bare/cased/concatenated `/HDBInstance123` and query-string forms, while a host merely *named* like the segment stays direct, failing closed on relative URLs), and the direct credential mint has a 30s abort timeout.

Full unit suite green (2655) with regression tests for each fix above; `tsc`/`oxlint`/`dprint` clean.

_🤖 Generated with [Claude Code](https://claude.com/claude-code)_

dawsontoth:
Pushed `e832fca6` with three more fixes that came out of reviewing my own hardening:

- **Minted tokens are stamped with Studio's computed URL for the entity**, not the spec picker's current selection. The previous commit stamped the selection, which recreated a narrower version of the exact leak the stamping exists to prevent — selecting a foreign declared server and then minting would have marked an instance token valid for that origin. Regression test added.
- `signOutFromPotentiallyAuthenticatedInstances` now clears explorer state too — it flags sign-out directly rather than through `signOutLocally`, so it was the last sign-out path skipping the cleanup.
- A mirrored cross-tab invalidation advances this tab's in-memory epoch, so an in-flight mint is caught by the epoch check as well as the attempt guard. Mint abort/network failures also now surface stable messages instead of the browser's opaque abort text.

**One thing to be upfront about:** this last commit has **no outside-model review coverage**. The run I kicked off for it produced nothing — Gemini hit an account quota (resets in ~5h) and Codex hung and was killed with no budget left to retry — so the `Review-Coverage: ran=none` footer is accurate rather than an oversight. Its parent `254c81ee` and everything before were reviewed by codex+gemini. What `e832fca6` does have is the full local gate (2656 unit tests, tsc, oxlint, dprint) plus a regression test per fix.

Given the last few rounds of findings were in my own patches on a security-sensitive surface, I've stopped self-iterating here — @kriszyp @kylebernhardy @DavidCockerill, the credential-to-server stamping model in `ApiExplorer.tsx` and the epoch logic in `authStore.ts` are where a human read is worth the most. CI is green and the branch is mergeable.

_🤖 Generated with [Claude Code](https://claude.com/claude-code)_

src/features/auth/store/authStore.ts
```diff
@@ -95,10 +76,109 @@ class AuthStore {
...
+		return () => window.removeEventListener('storage', handler);
+	}
+
+	/**
+	 * App-level (always-on) listener that clears the per-tab explorer credential for an entity another
+	 * tab signed out — even when this tab's explorer is unmounted, so a tab that navigated away can't
+	 * later restore a revoked token. Also re-scrubs the legacy localStorage secret if a pre-upgrade tab
+	 * rewrote it. Installed once at bootstrap.
+	 */
+	public installExplorerCrossTabCleanup(): () => void {
```
kriszyp, line 153: This listener only observes future `storage` events; startup never reconciles the value already stored in `Studio:ExplorerAuthEpoch`. Because sessionStorage survives a reload in the same tab, tab A can retain an Explorer credential, reload or be discarded while tab B signs the entity out, then start after the invalidation write and miss it entirely. A later user in tab A can consequently restore the stale credential. Please persist a durable per-entity/global generation and store or compare that generation with each credential during bootstrap/read, with a regression test where invalidation happens before the listener is installed.

src/integrations/api/instance/auth/createInstanceAuthenticationTokens.ts
```diff
@@ -29,6 +30,60 @@ export async function createInstanceAuthenticationTokens(
...
+		});
+	} catch (error) {
+		// Replace the browser's opaque abort/network text with something the log-in form can explain.
+		throw new Error(
+			controller.signal.aborted
+				? 'The instance did not respond within 30 seconds.'
+				: `Could not reach the instance to log in. ${error instanceof Error ? error.message : ''}`.trim(),
+		);
+	} finally {
+		clearTimeout(timeout);
```
kriszyp, line 78: The timeout is cleared as soon as `fetch()` returns headers, before `response.json()` consumes the body. An instance that sends headers and then stalls the body can therefore still leave this login pending indefinitely. Please keep the abort timer active through body consumption—put both `fetch` and JSON parsing inside the same `try`/`finally`—and add a test for a response whose body never resolves.

kriszyp (approved):
Nice work, a couple of follow-up suggestions, but I don't consider them blockers.

src/features/auth/store/authStore.ts
dawsontoth: Fixed in `a439aad7` — you're right that an event-only signal can't cover this, since sessionStorage survives a reload. The generation is now durable (localStorage) instead of an in-memory map, and every stored credential records the generation it was created under; a credential whose stamp doesn't match the current generation is not sent, so the check is a read-time comparison rather than a delivered event. `pruneStaleEntitySettings` strips those credentials (keeping non-secret server/method selections) and runs immediately when the app-level cleanup installs at bootstrap, which covers exactly your case — invalidation written before any listener exists. Regression test added for that path ('honors an invalidation written before any listener was installed'), plus stale/matching/unstamped cases. Related fix while in here: the cross-tab listener now re-reads state instead of assuming the event names this entity, so another entity's sign-out no longer clears this one.

src/integrations/api/instance/auth/createInstanceAuthenticationTokens.ts
dawsontoth: Fixed in `a439aad7`: `fetch` and `response.json()` are now inside the same `try`/`finally`, so the abort timer covers body consumption. One subtlety the new test caught in my first attempt — the existing `.catch(() => undefined)` on `json()` swallowed the abort and it surfaced as 'did not return an operation token'; an abort now re-throws so it reports the 30s deadline. Test added with a body that stalls until the signal aborts.

